PR #471, opened 2014-12-30
Set a nice non-empty title on shutdown

incertia:
Got the idea from Vim. Looks better than an untitlted terminal.


boothj5:
Testing on a few terminals the behaviour varies.

On gnome-terminal it works ok, the message is shown until an application rewrites it.
On Terminator, the message is shown, but previously I would see `/bin/bash`
On ITerm, previously I would see `bash`, `git` or whatever was running, with this change the profanity message is always prepended to `(bash)` or `(git)` etc.

Perhaps this could be a user preference, disabled by default?

incertia:
I believe the behaviours are terminal emulator dependent. Both gnome-terminal and xfce4-terminal have options on what to do with the dynamic title (replace, before initial, after initial, ignored). I believe ITerm should have an option to change this behaviour as well. Terminator has no option for initial shell title and it probably displays the path of the current shell command being run (I see `/usr/bin/zsh`).

boothj5:
Since most terminals have options to handle the title, maybe best would be:

Enable the exit message (`titlebar.exit`) by default, but only show it if `titlebar` is enabled, i.e.:

```
if (prefs_get_boolean(PREF_TITLEBAR) && prefs_get_boolean(PREF_TITLEBAR_EXIT)) {
  // show exit message
}
```

And allow users to disable `titlebar.exit` in case it conflicts with their terminal settings and they don't want to change the them.

What do you think?

incertia:
Added the preference setting and added a command to toggle it on and off.

boothj5:
Thanks.

Only comment is that in general related settings use the same command, and the keys have the same prefix with a dot separator, so for example we could have:

```
/titlebar show on|off
/titlebar goodbye on|off
```

With keys:

```
titlebar.show
titlebar.goodbye
```

Perhaps also the old behaviour should be maintained, so if `titlebar.goodbye` is set to `off`, it is cleared.

I'm happy to make this small change with the merge unless you want to.

incertia:
Feel free.

PR #472, opened 2014-12-30
Handle signals when asking for password

incertia:
C-c and C-z no longer interfere with Profanity when asking for password.

### DIFF
--- a/src/command/commands.c
+++ b/src/command/commands.c
@@ -80,6 +80,12 @@ static void _who_roster(gchar **args, struct cmd_help_t help);
 
 extern GHashTable *commands;
 
+// typedef for signal handlers
+typedef void (*sighandler)(int);
+
+void connect_sigint(int sig){}
+void connect_sigtstp(int sig){}
+
 gboolean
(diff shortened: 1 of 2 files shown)


boothj5:
Since ncurses is set to `raw` mode, `ctrl-c` and `ctrl-z` are passed through to profanity rather than generating a signal (apart during the password prompting), so there is no real need to save and restore the handlers.

Also since we want to ignore these signals, we can use `SIG_IGN` as the function rather than creating our own.

Given the above, I think an easier solution is to add two more ignore handlers to the `_init()` function in profanity.c:

https://github.com/boothj5/profanity/blob/master/src/profanity.c#L256

I.e:

```
signal(SIGINT, SIG_IGN);
signal(SIGTSTP, SIG_IGN);
```

incertia:
> Given the above, I think an easier solution is to add two more ignore handlers to the `_init()` function in profanity.c
This sounds reasonable. I only saved and restored them because I didn't know if there were other instances where they would be handled. I am closing this pull request.

Issue #300, opened 2014-02-06
Use Ubuntu Keyring (gnome-keyring) as password storage


dreh23:
It would be convenient to have the user and password stored in the system password storage.

boothj5:
The referenced pull request should now have implemented this feature, the comments contain an example using https://apps.ubuntu.com/cat/applications/libsecret-tools/ with the `gnome-keyring`.

Please reopen if you have any issues.

PR #480, opened 2015-01-07
Add support for evaluated password (keyring support)

peterlvilim:
This pull request is intended to add support for keyrings.  The idea is that you'll now be able to include the account configuration file in your dotfiles which many people keep in a public repository.

I took an approach that is similar to mutt and several other common open source projects.  There is a new field called eval_password.  This field when loaded evaluates a shell command which will return the password.  This means that the user just needs a script that accesses their keyring and avoids introducing dependencies on specific keyrings.  In my case I use: http://www.passwordstore.org/  There are scripts for gnome-keyring and other more popular keyrings as well.

So my new config line looks like:
eval_password = pass pvilim_xmpp@facebook.com
for my facebook account.

I also added code to prevent eval_password and password being set at the same time to cut down on the chance of a user leaking their password or having configuration problems.

Also, how does the testing framework work?  I checked the dev mailing list and docs but wasn't able to find out how to run it.  Once I know that I can add some test coverage.

### DIFF
--- a/src/command/command.c
+++ b/src/command/command.c
@@ -931,6 +931,7 @@ static struct cmd_t command_defs[] =
           "|xa|dnd          : Priority for the specified presence.",
           "resource         : The resource to be used.",
           "password         : Password for the account, note this is currently stored in plaintext if set.",
+          "eval_password    : Shell command evaluated to retrieve password for the account.  Can be used to retrieve password from keyring.",
           "muc              : The default MUC chat service to use.",
           "nick             : The default nickname to use when joining chat rooms.",
           "otr              : Override global OTR policy for this account: manual, opportunistic or always.",
@@ -1260,6 +1261,7 @@ cmd_init(void)
     autocomplete_add(account_set_ac, "dnd");
     autocomplete_add(account_set_ac, "resource");
     autocomplete_add(account_set_ac, "password");
+    autocomplete_add(account_set_ac, "eval_password");
     autocomplete_add(account_set_ac, "muc");
     autocomplete_add(account_set_ac, "nick");
     autocomplete_add(account_set_ac, "otr");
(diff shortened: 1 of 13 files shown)


peterlvilim:
Nevermind on testing comment.  Found make check.  Will push an update tomorrow to fix the test failures.

peterlvilim:
Current tests are now passing as well as my manual testing.  The only thing lacking I believe is unit tests for the additions.  Wanted to get some feedback though before I wrote those.

boothj5:
Good stuff. I've tested with gnome-keyring and https://apps.ubuntu.com/cat/applications/libsecret-tools/ and it works well, so this will close #300 also.

I''ll do a few more tests etc and merge in the next few days.

peterlvilim:
Sounds good.  Thanks for the quick response.  I'll go ahead and add automated tests shortly here.

boothj5:
Merged into master.

I'm making use of this change already myself, very useful thanks,

peterlvilim:
Sounds great.  Funny you posted this.  I was about to push 3 tests I made for it to this branch.  Sorry for delay.  Should I still put them up or do you feel good about the testing?

boothj5:
No problem, I'll pull the tests if you have some.

For reference (this will be added to the website on release), using `secret-tool` with `gnome-keyring`:

Add a password:
```
secret-tool store --label="Jabber password" xmpp bob@server.org
```

Enter password at prompt.

Command used for `eval_password` setting:

```
secret-tool lookup xmpp bob@server.org
```

peterlvilim:
Looks good.  Thanks for merging.  Added unit tests in another pull request.

boothj5:
Again just for reference, until the website is updated for 0.4.6.

On OS X, the `eval_password` can be used to retrieve a password from the Keychain by setting it to:

```
security find-generic-password -s <Keychain Item Name> -w
```

PR #483, opened 2015-01-12
Dynamic input block

Savar:
combining a longer timeout (to save CPU power) with a dynamic mode to be fast on input but efficient without input


Savar:
I didn't get the travis failure.. is this my fault? Locally "make check" is working fine..

boothj5:
Merged, thanks.

The broken test was because the automatic merge on github was against the latest code master, but I'm guessing you are a few commits behind, was a simple fix. I also added the autocompleter for the command.

PR #486, opened 2015-01-13
General fixes related to eval_password

incertia:
* `eval_password` handling code is now in `cmd_connect`, so that the field can be changed without clearing itself.
* `pclose` is called on the `popen`ed `eval_password` command, but this is going to fail if the process never terminates. A solution would be to use `fork` instead of `popen` and manually `kill` the pid.
* `stderr` from the `popen`ed process does not make curses mad anyore.
* Added a sanity check for memory allocation of `account->password`.
* Unrelated, but moved the retrieval/freeing of the default account string to more convenient places.

### DIFF
--- a/src/command/commands.c
+++ b/src/command/commands.c
@@ -84,7 +84,6 @@ gboolean
 cmd_connect(gchar **args, struct cmd_help_t help)
 {
     gboolean result = FALSE;
-    char *def = prefs_get_string(PREF_DEFAULT_ACCOUNT);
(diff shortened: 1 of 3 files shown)


boothj5:
Merged the changes.

This introduced a segfault in `jabber_connect_with_account` when the `eval_password` command returns no output, since `account->password` is no longer set.

Fixed the segfault and added UI messages when the password evaluation fails.

PR #489, opened 2015-01-14
Strip quote chars from name autocomplete

peterlvilim:
This is to address a minor annoyance I had with autocomplete.  When I'm logged into FB all the user names show as people's full names with spaces.  When you autocomplete it adds quote characters around a person's name.  However, if you for instance autocomplete then realize that your /msg "Firstname Lastname" has the right first name but wrong last name and you backspace the lastname then the autocomplete will fail when you hit tab again.  This is because the first quote characterbefore Firstname is being used to match against the autocomplete.

This diff strips out beginning and ending quote characters before doing autocomplete.  I chose to do it at this point in the code because it will only impact the user interaction functions (/msg etc) which is really the only area I know of where it would be helpful.

Also, I haven't tested this in a chat room command.c:1953 because I couldn't find a public one with users in the room (tried a few).

Tests pass.  Let me know if I should add more.  Wasn't sure if there were any good ones that could be added.

### DIFF
--- a/src/command/command.c
+++ b/src/command/command.c
@@ -98,6 +98,7 @@ static char * _role_autocomplete(char *input, int *size);
 static char * _resource_autocomplete(char *input, int *size);
 static char * _titlebar_autocomplete(char *input, int *size);
 static char * _inpblock_autocomplete(char *input, int *size);
+static char * _strip_quotes_from_names(char *input, int *size);
 
 GHashTable *commands = NULL;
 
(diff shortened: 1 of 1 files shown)


boothj5:
Merged, thanks.

Made a few small changes due to other changes in master, and added a few tests.

It might be worth using this on all commands, but I'm currently looking into GNU Readline which has its own autocompletion api, so this fix is good for now.

peterlvilim:
Thanks!

Checked the diff and your changes look good.  Happy to have this feature now.

Issue #493, opened 2015-01-27
Add option for 12-hour timestamp format (AM/PM)


arselinux:
This is a rather simple feature request, and it would be nice for users to be able to select a 12-hour format for timestamps both in the UI and in logs (something along the line of "hh:mm:ss AM/PM". Mentally converting times is easy enough, but it would be a nice addition for users in countries where the 24-hour format is almost never used (like the US).

incertia:
Here is a suggestion. I'll implement it if you guys want. The timestamp format is given by a preference string that is fed to <code>strftime</code>, so controlling the output is fairly easy. e.g. "%H:%M:%S" for your standard 24 hour clock.

boothj5:
Sounds like a good approach.  The only complexity might be in some of the output functions calculating the length of the time string, but probably not too problematic.

incertia:
Done in #496

boothj5:
Merged the pull request from @incertia into master. As mentioned, any format using the following should work:

https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format

E.g. for a non 12 hour clock time with seconds and AM/PM:

```
/time main "%I:%M:%S %p"
/time statusbar "%I:%M:%S %p"
```

```
10:23:04 AM
06:47:10 PM
```

PR #496, opened 2015-01-28
Use custom format string for time preference

incertia:
Exactly as described in title. Any string that `g_date_time_format` takes is usable. This solves #493.


boothj5:
There is also an 'off' setting, so perhaps the command could be:

```
/time set format
/time off
```

Also the width of the time string is used to calculate the word wrapping indent at https://github.com/boothj5/profanity/blob/master/src/ui/window.c#L962

incertia:
You should set the time format to the empty string to turn off formatting. #500 fixes the problem of not being able to send empty strings as arguments to non-freetext commands.

I'll see if I can make something up for calculating the indent width.

incertia:
Sorry for lack of activity. Decided to go with the set/off approach because the empty argument fix was never pulled. Fixed indent problems, applied same logic to statusbar, and probably fixed any indent-similar issues with the statusbar.

boothj5:
Finally got around to merging this request after some testing.  Very useful feature thanks.

PR #503, opened 2015-02-08
Bind ^L to own clear method to retain ui elements

stormc:
When using readline's standard `^L` binding, the screen
is completely cleared, including status and title bar.
By binding `^L` to profanity's clear method, status and
title bar are retained.


boothj5:
Thanks.

PR #508, opened 2015-02-14
Bind ^D to /quit in main window and to /close in other windows

stormc:
In a Shell, I find it convenient to use `^D` to quit the application. Adopted to profanity,
* `^D` quits the application if the main window has focus whereas
* if a chat window has focus, `^D` closes this chat window.


boothj5:
I've been working with GNU Readline in a branch, which could make this a custom key binding, so I'll look at merging this functionality for 0.4.7.

Thanks.

PR #522, opened 2015-03-09
Fix libnotify configure check for FreeBSD

arvedarved:
See boothj5/profanity#428

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -153,22 +153,21 @@ AS_IF([test "x$PLATFORM" = xosx], [LIBS="-lcurl $LIBS"])
 ### Windows uses native OS calls
 ### OSX requires terminal-notifier
 
-AS_IF([test "x$PLATFORM" = xosx],
-        [AS_IF([test "x$enable_notifications" != xno],
+AS_IF([test "x$enable_notifications" != xno],
             [NOTIFIER_PATH="no"
             AC_PATH_PROG(NOTIFIER_PATH, terminal-notifier, no)
             AS_IF([test "x$NOTIFIER_PATH" = xno],
                 [AS_IF([test "x$enable_notifications" = xyes],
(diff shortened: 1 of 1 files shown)


boothj5:
What was the exact issue on FreeBSD with libnotify?

There are two problems I can see with the above patch, firstly you will get a warning mentioning OSX when configuring on other systems. Also, if someone configures with:

```
./configure --enable-notifications
```

The configure script will complete but notifications will not be included. The `--enable-[feature]` settings should exit if the feature cannot be included.

arvedarved:
The problem: there is no libnotify configure check for the PLATFORM="freebsd" case, so the respective variables where not set with --enable-notifications. I think it is not necessary to hide the libnotify configure check inside a PLATFORM check.

boothj5:
The desktop notifications are implemented differently depending on platform.

Cygwin uses native windows desktop notifications.
OSX uses the shell terminal-notifier program.
Linux uses libnotify.

I think the simplest thing would be to add a check for "freebsd" and do the same as for "nix"?

arvedarved:
Yes that would be the simplest thing, but i think it would be better to remove the platform checks and just check if libnotify or terminal-notifier are available.

boothj5:
I've made the change here: f0ffceefdd1b9ddb2cd63c2fa4fb6f15d7c0e635

Let me know if this works for you.

Out of interest, why do you think it would be better to remove the platform checks? Without them we'd be checking for and outputting warnings/errors for dependencies that are not required on those platforms.

arvedarved:
Thanks works for me!.

Well i think it is always better to check for features instead of platforms. Now you need to adjust this check everytime you add support for another platform. Otherwise all platforms supported by libnotify are automatically supported. 

An additional "checking for libnotify...no" does not hurt anyone.

Issue #551, opened 2015-05-29
Window auto tidy on close


PMaynard:
When a window is closed it would be nice if the /wins tidy command is executed automatically.
This should be a fairly easy patch. Could be a configurable option.

boothj5:
Sounds like a good idea, since it should be fairly easy to implement I've added to the 0.4.7 release list.

PR #552, opened 2015-05-29
Added window auto tidy.

PMaynard:
Adds the command /winstidy which is a boolean option enabled by default. 
When a window is closed it will perform a window tidy command, removing gaps from the window list.

Re: Issue #551


boothj5:
The pull request looks good, I've added a couple of small comments inline.

Also most of the UI settings are part of themes, and this would be another useful theme option. All that's required is a line added to https://github.com/boothj5/profanity/blob/master/src/config/theme.c#L425

If you don't have time to do any of the above, just let me know and I'll make the changes during the merge.

Thanks.

boothj5:
Merged, with one minor change, the windows are only tidied on the `/winstidy` command if the argument is `on`.

Thanks.

PR #553, opened 2015-05-29
Added libuuid-devel

pszypowicz:
Package needed on fedora22


boothj5:
Upgraded script to use dnf.

PR #565, opened 2015-06-20
Added handler for libgcrypt's logs

### LINKED ISSUE #562: OTR key generation shows error when terminal resized
Start OTR key generation with `/otr gen`, and then resize window, the following error is displayed:

```
select() error: Interrupted system call
```

On Awesome Window Manager, the key generation still eventually completes and the error disappears.  At the very least, hide or prevent the error message being displayed.

pasis:
This patch fixes #562 by hiding messages to log file.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -226,7 +226,7 @@ fi
 
 AM_COND_IF([BUILD_OTR],
     [AC_CHECK_LIB([gcrypt], [gcry_set_log_handler],
-        [AC_CHECK_HEADERS([gcrypt.h], [],
+        [AC_CHECK_HEADERS([gcrypt.h], [LIBS="-lgcrypt $LIBS"],
             [grypt.h not found. Try to install libgcrypt-devel package])],
         [libgcrypt not found])],
     [])
--- a/configure.ac
+++ b/configure.ac
@@ -224,6 +224,13 @@ if test "x$enable_otr" != xno; then
             [AC_MSG_NOTICE([libotr not found, otr encryption support not enabled])])])
 fi
 
+AM_COND_IF([BUILD_OTR],
+    [AC_CHECK_LIB([gcrypt], [gcry_set_log_handler],
+        [AC_CHECK_HEADERS([gcrypt.h], [],
+            [grypt.h not found. Try to install libgcrypt-devel package])],
+        [libgcrypt not found])],
+    [])
+
 AS_IF([test "x$with_themes" = xno],
     [THEMES_INSTALL="false"],
     [THEMES_INSTALL="true"])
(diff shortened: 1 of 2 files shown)


boothj5:
I'm getting an error on compile:

```
/usr/bin/ld: src/otr/otr.o: undefined reference to symbol 'gcry_set_log_handler@@GCRYPT_1.6'
/lib/x86_64-linux-gnu/libgcrypt.so.20: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
Makefile:1075: recipe for target 'profanity' failed
```

I guess profanity needs to be linked to libgrypt directly on some systems? Tested on Ubuntu 13.04.

pasis:
You're right, this error appears, because -lgcrypt is missed. I will look at it.

pasis:
Added AC_LIB_CHECK, it should append -lgcrypt to LIBS. I'm not sure if this is correct way to set handler, because profanity doesn't use libgcrypt directly and it can make problems with some configurations (static libs, etc).

I can try to write a generic handler that will catch everything from stderr and split it into separated log messages. If it sounds ok I'll take this task :)

boothj5:
Sounds good, would also be handy for the plugins branch, which at the moment also messes up the profanity UI when a plugin writes to stderr.

It'll be a while before the release of 0.4.7, but if you haven't had a chance before then I'll pull the `-lgcrypt` fix instead.

Thanks.

pasis:
I've implemented a prototype of stderr handler to show the idea: https://github.com/pasis/profanity/tree/stderr-log-handler

I described some problem with this approach in the commit message.

boothj5:
Previously without this code, a print to stderr from `log_msg` would appear and break the UI? So I guess whilst an infinite loop is bad, it's a result of bug that would need to be fixed anyway.

In terms of Windows, since profanity can only run in a Cygwin environment I'm not sure how common this would be, but in the worst case the the log message would be appended with no newline?

Difficult decision, but given it means no linking with a library profanity doesn't need directly and that this might also help with plugins that print to stderr, I'm thinking it's worth adding.

What do you think? I'll obviously test thoroughly, and on Cygwin.

pasis:
I will fix problems with new line and infinite loop. The rest is up to you :).

I checked handler with patch, because usual usecase doesn't produce output to stderr:
```
diff --git a/src/log.c b/src/log.c
index 941bea6..5adcddd 100644
--- a/src/log.c
+++ b/src/log.c
@@ -101,7 +101,8 @@ log_info(const char * const msg, ...)
     va_start(arg, msg);
     GString *fmt_msg = g_string_new(NULL);
     g_string_vprintf(fmt_msg, msg, arg);
-    log_msg(PROF_LEVEL_INFO, PROF, fmt_msg->str);
+//    log_msg(PROF_LEVEL_INFO, PROF, fmt_msg->str);
+    fprintf(stderr, "%s\n", fmt_msg->str);
     g_string_free(fmt_msg, TRUE);
     va_end(arg);
 }
```

boothj5:
Close, fixed by #573

PR #573, opened 2015-06-23
Introduced stderr log handler

### LINKED ISSUE #562: OTR key generation shows error when terminal resized

pasis:
This patch fixes #562. So, ignore #565.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -34,6 +34,7 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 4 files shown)


boothj5:
I'll test this out and merge, thanks.

boothj5:
Tested with window resizing, also tested in the plugins branch and it also fixes issues with plugins writing to stderr on load and execution etc.

Saved me some work, thanks.

PR #580, opened 2015-06-24
Don't mix allocator types in preferences.c

pasis:


boothj5:
Merged, I'll keep an eye out as I think there may be more of these.

PR #581, opened 2015-06-24
PGP: Suppressed error message on shutdown

pasis:


boothj5:
I found the same issue and fixed it around the same time.

Issue #567, opened 2015-06-22
Compilation error: gpgme.h not found


pasis:
gpg.c has `#include <gpgme.h>`, but my system installs this header to /usr/include/gpgme/gpgme.h. I found out that gpgme package provides gpgme-config and it returns proper paths:
```
$gpgme-config --cflags
-I/usr/include/gpgme
$gpgme-config --libs
-L/usr/lib64 -lgpgme -lassuan -lgpg-error
```

This error can be fixed by using gpgme-config in configure.ac, similar to https://github.com/GNOME/seahorse/blob/master/configure.ac#L116

boothj5:
Should be fixed now. Testing locally shows the libs from `gpgme-config` are used, but the `--cflags` output is empty so I can't test this part.  Let me know if it works for you.

pasis:
It works. But, I'd suggest to use AM_CPPFLAGS instead of CFLAGS, because other checks use it. Thanks.

boothj5:
Made the change.

After reading the autotools docs I guess its generally not a good idea to manipulate `CFLAGS` as it's for user specified args.

PMaynard:
I believe there is also an issue regarding Largefile support (LFS). I tried compiling on Debain 32bit 7.8 and got the following error on make:

```
In file included from src/pgp/gpg.c:40:0:
/usr/include/gpgme.h:80:2: error: #error GPGME was compiled with _FILE_OFFSET_BITS = 64, please see the section "Largefile support (LFS)" in the GPGME manual.
make[1]: *** [src/pgp/gpg.o] Error 1
```

As far as I understand there is a simple fix in configure.ac but I know nothing about it. Found these links might be helpful [LFS Docs](https://www.gnupg.org/documentation/manuals/gpgme/Largefile-Support-_0028LFS_0029.html) and in GPG [mail list](http://lists.gnupg.org/pipermail/gnupg-devel/2011-June/026116.html)

boothj5:
I've taken a look at the docs and added the `AC_SYS_LARGEFILE` macro to configure.ac.

Let me know if it makes a difference.

PMaynard:
Still get the same error: 

    In file included from src/pgp/gpg.c:43:0:
    /usr/include/gpgme.h:80:2: error: #error GPGME was compiled with _FILE_OFFSET_BITS = 64, please see the section "Largefile support (LFS)" in the GPGME manual.
    make[1]: *** [src/pgp/gpg.o] Error 1

[...]

PR #582, opened 2015-06-24
Added include config.h to pgp/gpg.c

pasis:
This patch should fix problem that is described in #567, but I can't check it.


PMaynard:
I can confirm this patch fixes my issue in #567

PR #586, opened 2015-07-01
Add no-copy and no-store hints to OTR messages

SamWhited:
This adds the `no-copy` and `no-store` message processing hints to OTR messages as per [XEP-0334](https://xmpp.org/extensions/xep-0334.html).

While the `no-copy` hint effectively does the same thing as the Carbons `private` hint, some clients choose to respect one but not the other.

The `no-store` hint ensures that OTR messages (which are perfect forward secret and can't be read later) are not stored in archives such as the MAM archive.


boothj5:
Merged, thanks.

PR #637, opened 2015-10-06
Have the installer on cygwin check for an existing installation of apt-cyg

halfur:
The install-all.sh script doesn't check for an existing installation of the [apt-cyg](https://github.com/transcode-open/apt-cyg) script before downloading and installing it into /usr/local/bin/. As the readme of that script recommends installing it into /bin/ this can lead to redundant installation of multiple, potentially different versions of the script.

This patch adds a check for an existing installation of apt-cyg and extracts the installation method into a function.

### DIFF
--- a/install-all.sh
+++ b/install-all.sh
@@ -63,12 +63,7 @@ cygwin_prepare()
     echo Profanity installer... installing dependencies
     echo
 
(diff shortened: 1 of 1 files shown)


boothj5:
Thanks, I'll pull this when I get a chance to test on my Windows VM.

boothj5:
Tested and merged, thanks.

PR #652, opened 2015-10-22
dereferencing pointers

jubalh:
In two cleanup/free function the pointer got set to NULL so one can later check whether it got freed or not. However this change didn't have any effect outside of the function.
This dereferences it hopefully correctly.

Please review.

### DIFF
--- a/src/config/preferences.c
+++ b/src/config/preferences.c
@@ -203,7 +203,6 @@ prefs_free_string(char *pref)
     if (pref) {
         g_free(pref);
     }
-    pref = NULL;
 }
 
 
--- a/src/muc.c
+++ b/src/muc.c
@@ -976,6 +976,5 @@ _occupant_free(Occupant *occupant)
         free(occupant->jid);
         free(occupant->status);
         free(occupant);
-        occupant = NULL;
     }
 }
--- a/src/ui/buffer.c
+++ b/src/ui/buffer.c
@@ -76,7 +76,6 @@ buffer_free(ProfBuff buffer)
 {
     g_slist_free_full(buffer->entries, (GDestroyNotify)_free_entry);
     free(buffer);
-    buffer = NULL;
 }
 
 void


jubalh:
Hmm right, it's different with structs. How was that..

pasis:
Just my thoughts about the patch: I agree that setting to NULL is a mistake in these functions. But I'm not agree with changing of the prototypes. In most cases strings can't be used after `prefs_free_string()` and changing the argument to (char **) gives no advantage. But from other side this makes interface not canonical and not predictable.

So my suggestion is to remove `pref = NULL` and leave prototype as is. If you really need an interface which sets argument to NULL you should create new one with a proper name (usually, it is the same name, but with `free0` instead of `free`).

Let's see what James thinks.

boothj5:
I will get around to your pull requests soon, I haven't yet had time this week!

In regards to the pointer references:

1) I also agree with you on removing the NULL assignments within free() functions. Whilst it doesn't cause any problems, it is of course completely useless.

2) You've already discussed and come to a conclusion in the chat room about the function prototypes but just to add my thoughts...

I basically follow the style that seems most common in stdlib and other libraries I've used.  i.e. firstly a function is supplied to allocate memory for a data type, returning a pointer to that memory. and secondly a function is supplied to free the memory associated with a pointer of that data type.  

It does of course require some discipline on behalf of the programmer to NULL the reference once free'd, but from what I've seen, this is very common practice for C developers.  It's not necessarily any more "correct" than your idea, just more common.

boothj5:
One more thought about memory allocation and pointer references.

In terms of coding style, most of the time a function (or a block of code) should usually follow the pattern of:

1) Call functions to allocate memory for data types
2) Use those data types
3) Call functions to free memory allocated in 1)

Using this style, since the pointers themselves only exist in the scope of the function/block, the NULL assignment is not even necessary after 3).

I'm sure there are many cases where profanity doesn't currently follow this pattern, but I think it would be good to aim for as the code evolves, because keeping the scope of memory allocations small would reduce the chances of errors (segfaults etc).

jubalh:
Then I will change this PR to remove the ptr = NULL in the functions

jubalh:
Changed as suggested.

PR #657, opened 2015-10-28
Fix statusbar time in complex theme

BlueMonday:
The complex theme had the time in the status bar set to `$H:$M:%S` which resulted in the time being displayed as `$H:$M:30`.


boothj5:
Merged, thanks.

Issue #585, opened 2015-07-01
Export roster list to txt/csv file


longk15t:
@boothj5 could we do this on 0.4.7 ? I'm very appreciate about it.

incertia:
what sorts of fields are you looking for in the file? it shouldn't be too hard to export a list of emails with some extra data.

longk15t:
just emails, no need other information. I really need this feature.

boothj5:
Added in #680 from @incertia

PR #680, opened 2015-11-24
Implement feature request in #585 (export command)

incertia:
Fairly self explanatory. Opens up a file descriptor to the desired file (fails and prints the corresponding message) and writes the bare jid and name as CSV entries.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@
 - Last Activity (xep-0012)
 - Run command sequence with /script
 - Account startscript property
+- Add /export command
 
 0.4.7
 =====
--- a/src/command/command.c
+++ b/src/command/command.c
@@ -1777,7 +1777,7 @@ static struct cmd_t command_defs[] =
         CMD_ARGS(
             { "<filepath>", "Path to the output file." })
         CMD_EXAMPLES(
-            "/export /path/to/output.cxv",
+            "/export /path/to/output.csv",
             "/export ~/contacts.csv")
     },
 };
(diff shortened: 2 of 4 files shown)


boothj5:
Thanks, a few minor comments:

1) If not connected the command shows `No contacts in roster.`  Perhaps it could show `You are not currently connected.`, see `/who` for an example.

2) The home directory shortcut is not working for me, e.g. `/export ~/rosterbak.csv` shows `error: cannot open ~/rosterbak.csv: No such file or directory`. I'm not sure `~` is guaranteed to work with every implementation of `open()`, so a safer option may be to replace it with `getenv("HOME")` if it is the first character of the argument.

3) Is the comment `/* temporary, we SHOULD pass everything to an escape function (to escape out quotes) and then use fputs */` still needed? Ideally I'd rather not include a temporary solution, if the current solution works and you see no potential issues, we could remove the comment.

4) Small note on code conventions, just to fit with the rest of the code. Spaces are included between brackets, e.g. `if (something) {` rather than `if(something){`. Also static functions start with a `_` and the open brace is on a newline.  I can make these styling changes on merge if you want.

incertia:
I can make these changes.

boothj5:
One other small thing, since the min and max args are specified with the command definition (1, 1 in this case) the check for `args[0]` is not needed (I'm aware this is not obvious when implementing the command function).

Returning `FALSE` from a command will quit Profanity.  So you could either remove the `args[0]` check altogether, or just call `cons_bad_cmd_usage(command)` and return `TRUE` which will be redundant, but safer than returning `FALSE`.

incertia:
Everything should be OK now except for a missing space on L810. You can add that in when merging.

boothj5:
Nice one thanks.  Merged.

Issue #743, opened 2016-02-23
Compilation Fails with Implicit Declaration Error


femnad:
Hello,

Compling profanity on the current HEAD (ae46e64) fails with the following error:
```

src/ui/mucwin.c: In function ‘_mucwin_print_mention’:
src/ui/mucwin.c:383:5: error: implicit declaration of function ‘free’ [-Werror=implicit-function-declaration]
     free(mention_section);
     ^
src/ui/mucwin.c:383:5: error: incompatible implicit declaration of built-in function ‘free’ [-Werror]
```
The feature introduced via 2f82f50 uses `free` for freeing a variable. Since I was able to replicate this on multiple environments I will shortly create a pull request which fixes the error by replacing `free` with `g_free` which is already included. I can't see any reason why `free` would be preferred in this case, however, in order to play it safe and document the fix I am creating this issue.

boothj5:
Fixed by including `stdlib.h`.

Although `g_free` is probably not a problem, I try to use it only when memory has been allocated by the same library.

PR #744, opened 2016-02-23
Use g_free for Freeing the Message Section

femnad:
Using `g_free` in the `_mucwin_print_mention` function fixes the implicit declaration error.

Related issue: #743


boothj5:
Closed as per #743

PR #758, opened 2016-03-12
Tray Icon for Profanity

Dav1d23:
Please let me know if you would like me to add something else, or to change/check something.

Thanks!


boothj5:
Just to let you know I haven't forgotten this request, I've been away, will look into merging this over the next few days.

Thanks.

boothj5:
Did a quick merge to try this out.

First thing, we'll need to make the `libgtk2.0-dev` dependency optional, so the compile will still work on systems that don't have it.

There are other examples of this in `configure.ac` (`libnotify` for example) where a `#define` is created `HAVE_LIBNOTIFY` if it's found, or a message is shown during configure `AC_MSG_NOTICE` but the process continues.  All code that uses it has to be wrapped by a check:

```
#ifdef HAVE_LIBNOTIFY
...
dependency specific code
...
#endif
```

Let me know if you need any help.

I also think the Copyright in the new files needs to be the same as the others, since I own the Copyright for the Profanity source, and splitting the ownership could get complicated.  I don't want to take any credit away from you though, so if you want to add an additional comment that you wrote it, including your details that's fine by me.

Dav1d23:
Ok I'll try it out and I'll ket you know!
Copyright-side, sorry, I just tried it on a local stub before applying it to profanity and I forgot to change... I'll do it as well!
Thx.

Dav1d23:
Done.

boothj5:
Tested the conditional compilation and works fine. Also tested the functionality on XFCE and Awesome Window Manager, and works on both. It's a pretty cool feature :)

I'll just spent a little time looking through the code, and add any comments here.

Thanks.

src/profanity.c
```diff
@@ -94,17 +98,15 @@ char *saved_status;
...
-static gboolean gtk_enabled = FALSE;
 
 void
-prof_run(char *log_level, char *account_name, gboolean use_gtk)
+prof_run(char *log_level, char *account_name)
 {
-    gtk_enabled = use_gtk;
     _init(log_level);
     plugins_on_start();
-    if (gtk_enabled) {
```
Dav1d23: should have left this one, IMHO

boothj5: I guess you mean that we should check that gtk initialised ok and set this flag?

Sounds good if you want to add it back.

src/tray.c
```diff
@@ -0,0 +1,162 @@
...
+                }
+                icon_msg_filename = g_string_new(icons_dir->str);
+                g_string_append(icon_msg_filename, "/proIconMsg.png");
+            }
+            g_string_free(name, true);
+            name = g_string_new(g_dir_read_name(dir));
+        }
+        g_string_free(name, true);
+    } else {
+        fprintf (stderr, "Unable to open dir: %s\n", err->message);
```
boothj5: `stderr` is piped to the profanity logs, but if you include `log.h` you can use `log_error()` which uses the profanity log handling code instead. It takes a string format and varargs and doesn't need a newline.

boothj5:
@Dav1d23 I'm guessing you've been busy, do you still want to make the minor changes above?  If you don't have time, I'm happy to make any changes and merge.

Dav1d23:
Hello!
Yup, quite busy these 2 weeks :) 
I've fixed the 2 minor things on the way - and an assert on shutdown :D
I think it should be fine right now!

boothj5:
Thanks for the work, now merged to master.

I'm going to add a few extra feature requests (/tray enable, /tray disable commands etc), if you want to look into any of them, just let me know so we're not working on the same thing.

Dav1d23:
I will be a bit busy next weeks as well, so i guess (if you have time) it
is better if you could make these changes ;)

Thanks!

(If you don't, I'll be happy to do it when I'll have some time :) )

PR #768, opened 2016-03-23
libotr upgrade on Fedora

0xPoly:
So libotr released a patch (libotr 4.1.1) for a interger overflow vulnerability. [0] I was surprised to find profanity was still using a vulnerable version of libotr, turns out the reason libotr version 3.x is forced in the Fedora install. 

This patch fixes that. Tested on Fedora 23 x86_64. I checked the dependencies for other distros, they do not appear to need to be changed. Not sure about cygwin.

[0] https://lists.cypherpunks.ca/pipermail/otr-users/2016-March/002581.html

### DIFF
--- a/install-all.sh
+++ b/install-all.sh
@@ -34,7 +34,7 @@ fedora_prepare()
     echo Profanity installer... installing dependencies
     echo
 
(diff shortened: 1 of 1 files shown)


boothj5:
Merged, thanks.

PR #788, opened 2016-04-12
Implement XEP-0363: HTTP File Upload

asdil12:
https://xmpp.org/extensions/xep-0363.html

Autocompletion works now.

sendfile    
--------

### Synopsis
```
/sendfile <file>
```

### Description
Send a file using XEP-0363 HTTP file transfer.

### Arguments
`<file>` : Path to the file.

### Examples
```
/sendfile /etc/passwd
/sendfile ~/images/nice.jpg
```


boothj5:
Thanks for this, might take me a little white to test and look through the code, but I'll definitely aim to have it in the next release.

I'll post any comments here.

asdil12:
I'm currently trying to get the testsuite running through.
For some reason it doesn't fail on my machine but on travis.

src/xmpp/iq.c
```diff
@@ -308,6 +309,21 @@ iq_disco_info_request(gchar *jid)
 }
 
 void
+iq_disco_info_request_onconnect(gchar *jid)
+{
+    xmpp_ctx_t * const ctx = connection_get_ctx();
+    char *id = create_unique_id("disco_info_onconnect");
```
boothj5: Some of the functional tests are failing because they stub responses based on stanza IDs. Using this function has incremented the number used in the unique IDs, so subsequent verifications in the tests fail as the IDs don't match up.

Since this stanza is only sent once on connect we could just give the ID as `disco_info_onconnect` instead of using the unique ID function.

However, since this is actually a problem with the tests (stabber is still in early development and needs more ways to prime/verify stanzas) and not the code, we could also fix all the IDs used in the tests.

The first option is quicker and won't cause any problems in this particular case... thoughts?

asdil12: I'm calling this function multiple times: Once for each server item (component).
So I need the unique ids.

boothj5: In that case I'll fix the tests.

boothj5: Fixed in a branch https://github.com/boothj5/profanity/tree/xep_0363_http_file_upload_pr

If you merge this to your branch I think it will update this pull request.

asdil12: OK, I'll do that.
Be aware that it will break your tests when you add components to your stub server.

boothj5: Indeed, there's loads of work to do on Stabber, it's really just a start as I couldn't find anything out there to run tests against.

asdil12:
Just tried it with ejabberd and it worked fine.

boothj5:
Tested on Ubuntu and things seem to be working fine, error messages when expected about missing feature, file sizes etc.

I had to make some changes to get things compiling on OSX.

* The forward reference `typedef struct prof_win_t ProfWin;` in `http_upload.h` isn't supported by default on OSX, as its a C11 addition.  So I moved some things around in the headers.  I'm gonna take a look at tidying the header structure, but thought it better to get this pull request in first.
* Added an extra arg to the CFLAGS: `-Qunused-arguments` to ignore the unused `-lpthread` argument on OSX
* Included `-lgio-2.0` manually, on other systems it seems to be included in the `glib-2.0` flags, but not OSX.

If you want to merge the following branch to this pull request it has all the changes (and anything else thats been added): https://github.com/boothj5/profanity/tree/xep_0363_http_file_upload_osx

I'm currently trying to get it compiling on Cygwin as there's a couple of issues there too.  Will update this pull request once I've got it working.

boothj5:
Cygwin is now compiling too, if you merge the branch https://github.com/boothj5/profanity/tree/xep_0363_http_file_upload_osx I'll take one final look over the code.

Thanks

asdil12:
I cherry-picked your new commits into my branch.

boothj5:
One of the changes was missed in the cherry picks since I added it during the merge.  The following commit has it if you want to cherry pick it: d0952895e62dd48660b1f8cc45e43eaa451937d1

src/command/command.c
```diff
@@ -795,6 +800,21 @@ static struct cmd_t command_defs[] =
...
+            CMD_TAG_CHAT,
+            CMD_TAG_GROUPCHAT)
+        CMD_SYN(
+            "/sendfile <file>")
+        CMD_DESC(
+            "Send a file using XEP-0363 HTTP file transfer.")
+        CMD_ARGS(
+            { "<file>", "Path to the file." })
+        CMD_EXAMPLES(
+            "/sendfile /etc/passwd")
```
boothj5: Thinking we should change the example, as someone will probably do and then blame Profanity...

src/profanity.c
```diff
@@ -323,6 +323,10 @@ _init(char *log_level)
     signal(SIGINT, SIG_IGN);
     signal(SIGTSTP, SIG_IGN);
     signal(SIGWINCH, ui_sigwinch_handler);
+    if (pthread_mutex_init(&lock, NULL) != 0) {
+        log_error("Mutex init failed");
```
boothj5: Perhaps profanity should exit if the lock can't be initialised?

src/tools/http_upload.c
```diff
@@ -0,0 +1,339 @@
+/*
+ * http_upload.c
+ *
+ * Copyright (C) 2016 Dominik Heidler <dominik@heidler.eu>
```
boothj5: Would you be ok to change the copyright headers to be the same as other files (i.e. under my name)?

src/tools/http_upload.c
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    unsigned int ulperc = 0;
+    if (ultotal != 0) {
+        ulperc = (100 * ulnow) / ultotal;
+    }
+
+    char *msg;
+    if (asprintf(&msg, "Uploading '%s': %d%%", upload->filename, ulperc) == -1) {
+        msg = NULL;
```
boothj5: I think passing NULL to the `win_update_entry_message` could cause a segfault as it does a `strdup(msg)`, not sure what the right thing to do is, maybe have a fallback message.

src/profanity.c
```diff
@@ -405,6 +409,9 @@ _shutdown(void)
     if (saved_status) {
         free(saved_status);
     }
+    //FIXME: kill all threads
```
boothj5: Do you think we should try to clean up any threads at this point?

src/tools/http_upload.c
```diff
@@ -0,0 +1,339 @@
...
+    CURL *curl;
+    CURLcode res;
+
+    upload->cancel = 0;
+    upload->bytes_sent = 0;
+
+    pthread_mutex_lock(&lock);
+    char* msg;
+    if (asprintf(&msg, "Uploading '%s': 0%%", upload->filename) == -1) {
+        msg = NULL;
```
boothj5: Same a previous comment, this `msg` is passed to `strdup()` further down.

src/tools/http_upload.c
```diff
@@ -0,0 +1,339 @@
...
+    free(output.buffer);
+
+    pthread_mutex_lock(&lock);
+    prefs_free_string(cert_path);
+
+    if (err) {
+        char *msg;
+        if (upload->cancel) {
+            if (asprintf(&msg, "Uploading '%s' failed: Upload was canceled", upload->filename) == -1) {
+                msg = NULL;
```
boothj5: Same as previous comment, there's a few following this one too.

src/profanity.c
asdil12: It is not very easy without skipping some memory free - so I would simply let the main thread keep the lock, let all threads running and exit (which will kill all threads and free all memory)

boothj5: Ok, sounds good.

asdil12:
I updated my patch according to your commants and cherry-picked the missing OSX patch.

boothj5:
Merged to master. Thanks again.

PR #792, opened 2016-04-16
Make tray icon configurable using /tray cmd

asdil12:
Allow to enable/disable the tray icon.


boothj5:
Merged into master with a couple of minor changes: moved the preference to the notifications group and made the command available even if support is not built in, as this is how other library dependent commands are currently handled.

Thanks.

PR #864, opened 2016-09-30
Fix http file upload on 32bit systems

asdil12:
Switched from `off_t` to `size_t`

### DIFF
--- a/src/tools/http_upload.c
+++ b/src/tools/http_upload.c
@@ -323,7 +323,7 @@ file_mime_type(const char* const file_name)
     return out_mime_type;
 }
 
-off_t file_size(const char* const filename)
+size_t file_size(const char* const filename)
 {
     struct stat st;
     stat(filename, &st);
--- a/src/tools/http_upload.h
+++ b/src/tools/http_upload.h
@@ -46,7 +46,7 @@
 
 typedef struct http_upload_t {
     char *filename;
-    off_t filesize;
+    size_t filesize;
     curl_off_t bytes_sent;
     char *mime_type;
     char *get_url;
@@ -61,7 +61,7 @@ GSList *upload_processes;
 void* http_file_put(void *userdata);
 
 char* file_mime_type(const char* const file_name);
-off_t file_size(const char* const file_name);
+size_t file_size(const char* const file_name);
 int is_regular_file(const char *filename);
 
 #endif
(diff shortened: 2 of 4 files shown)


boothj5:
I've added an include for `config.h` to `http_upload.c` which should enable 64 bit file offsets on any systems (including 32 bit) that support it.  This is dealt with using the `AC_SYS_LARGEFILE` macro in `configure.ac` which will `#define _FILE_OFFSET_BITS 64` if needed.

I think implicitly casting `off_t` to `size_t` could cause issues, since on systems where `_FILE_OFFSET_BITS = 64`, `off_t` can hold 8 bytes, whereas `size_t` may only hold up to 4 bytes.

 What was the actual issue you were trying to fix with this pull request?

asdil12:
I encountered the following problem:

When I compiled profanity 0.5.0 on an *armv7h* device and tried to send a file:

```
ls -l /home/dominik/profanity/profanity-0.5.0.tar.gz
-rw-r--r-- 1 dominik dominik 687169 30. Sep 16:56 /home/dominik/profanity/profanity-0.5.0.tar.gz
```

```
/sendfile /home/dominik/profanity/profanity-0.5.0.tar.gz
```

```xml
<iq id="prof_http_upload_request_26" to="im.f3l.de" type="get">
    <request xmlns="urn:xmpp:http:upload">
        <filename>profanity-0.5.0.tar.gz</filename>
        <size>646312206325677121</size>
        <content-type>application/x-compressed-tar</content-type>
    </request>
</iq>
```

This looks like some integer casting problem (that doesn't seem to happen on *x86_64*) and the pull request solves it. It also doesn't break file upload on *x86_64* linux.

The question is: Will you ever send files larger than 4GiB via XMPP-HTTP file transfer?

asdil12:
I just tried the new master with your change and it seems to help:

```xml
<iq id="prof_http_upload_request_21" to="im.f3l.de" type="get">
    <request xmlns="urn:xmpp:http:upload">
        <filename>profanity-0.5.0.tar.gz</filename>
        <size>687169</size>
        <content-type>application/x-compressed-tar</content-type>
    </request>
</iq>
```

asdil12:
So this means, that the error was already occuring when the `file_size(…)` function casted `st.st_size` to `off_t` when returning.
I initially thought, that the problem was the cast to `intmax_t` when formatting the stanza.

boothj5:
Just to make sure, does it work for you now on both 32bit and 64bit?

If so, I'll raise an issue (and close it) just for the record for the 0.5.1 patch release.

Thanks.

asdil12:
I did some tests:

- It works on *x86_64* for small files and files > 4GiB
- It works on *armv7h* for small files and files > 4GiB

boothj5:
Great thanks.  Attached and closed the issue.

Issue #855, opened 2016-09-19
Option to configure /clear behaviour


boothj5:
Currently the `/clear` command clears the main window, and anything previously displayed cannot be seen.

Add option for clear to clear the screen but also allow the user to see the history using Page Up.

PR #874, opened 2016-10-21
Change /clear behaviour. Closes issue #855.

spiridoncha:
I just change /clear behavior, but it's not configurable. Should I implement some option for /clear, which change behavior or something like that?

### DIFF
--- a/src/config/preferences.c
+++ b/src/config/preferences.c
@@ -1511,6 +1511,7 @@ _get_group(preference_t pref)
 {
     switch (pref)
     {
+        case PREF_CLEAR_PERSIST_HISTORY:
         case PREF_SPLASH:
         case PREF_BEEP:
         case PREF_THEME:
@@ -1626,6 +1627,8 @@ _get_key(preference_t pref)
 {
     switch (pref)
     {
+        case PREF_CLEAR_PERSIST_HISTORY:
+            return "clear.persist_history";
         case PREF_SPLASH:
             return "splash";
         case PREF_BEEP:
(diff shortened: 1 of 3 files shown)


boothj5:
@spiridoncha apologies for the delay in responding, I took a bit of a break after the last release.

I think making it configurable would be a good idea.  If you'd like to discuss let me know here.

Thanks,

James,

spiridoncha:
@boothj5 how should I make this behavior configurable?
I see some cases:

1. Create option for `profrc` in section `[ui]` with name `clear.persist_history` and default value is `true` or something like that.
2. Create alternative `/clear` command with new behavior and other name.
3. Create argument for current version of `/clear` with syntax `/clear full` or similar.

boothj5:
I think option 1 would be best, but maybe defaulting to `false`.

Usually when I allow changing existing behaviour with a configuration option I'll default to the original behaviour (clearing history in this case), just so current users get the same behaviour as before.

boothj5:
A couple of comments:

1) Most config options can be set with commands, in this case perhaps:
`/clear set persist_history on` - sets `clear.persist_history=true`
`/clear set persist_history off` - sets `clear.persist_history=false`
If you have any other ideas, let me know.

2) When the window buffer is full (1000 lines) the `/clear` command with `clear.persist_history=true` doesn't clear the screen. I came across the same issue when I briefly looked into clearing the screen, it might require some changes to how the screen is buffered.

jubalh:
@spiridoncha still interested in working on this?

jubalh:
I'll merge this and work on it in separate commits.

jubalh:
@spiridoncha thanks for your contribution!

Issue #906, opened 2017-02-09
Screen not redrawn after PGP secret key entry


ghost:
After specifying the PGP key ID with pgp.keyid within the accounts file on startup profanity asks for the secret key passphrase. It then doesn't redraw the screen, leaving it in an odd looking state. Really the screen should be redrawn.

Using the /clear command helps, but still leaves some items on the roster looking strange.

maxime-alves:
I found a way to fix this temporarily : 
/theme load mytheme
;D

Also it's good to say that this happens with the pinentry-curses program, because it shows itself in-screen, and doesn't pop-up another window. This is really problematic when working in a console-only session.

PR #908, opened 2017-02-10
Redraw the screen after entry of the PGP secret key passphrase

bashrc:
This might not be the most elegant way to do it, but this causes the screen to be redrawn after entering the PGP secret key passphrase on startup if pgp.keyid is set within the accounts file. It prevents the screen from appearing to be completely broken.

The problem seems to be that entering the secret key passphrase initiates a callback function which displays the input box, and when that callback is finished and the input box goes away then nothing happens to redraw the screen as it previously was.

### DIFF
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -118,6 +118,16 @@ sv_ev_roster_received(void)
             cons_show_error("Invalid PGP key ID specified: %s, %s", account->pgp_keyid, err_str);
         }
         free(err_str);
+
+        // Redraw the screen after entry of the PGP secret key
+        ProfWin *win = wins_get_current();
+        char *theme = prefs_get_string(PREF_THEME);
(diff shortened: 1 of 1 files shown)


boothj5:
Linking to #906 

I haven't forgotten this pull request, just need to find a little time to test etc.

Thanks.

willeponken:
Hello, I'm experiencing this issue, any reason for this not being merged? :)

jubalh:
Hi @willeponken. Sorry I intended to take a look at this (and the rest of the remaining PRs) but didn't get to it yet.
Mainly because I don't use PGP in profanity. Did you build with this patch enabled?

jubalh:
Fixed in https://github.com/boothj5/profanity/commit/6034b833be41909982b0d2bdac0a1d8499ad8e76

Issue #879, opened 2016-11-18
Case insensitive autocomplete


Ppjet6:
Would it be possible to allow completion to ignore case-sensitivity? For example I start with `f`, `foo` and `Foobar` are in the chatroom, I want to be able to complete to both. I guess this can apply to more than nicknames.

boothj5:
Added this functionality in master.

PR #916, opened 2017-03-17
Ignore accents when doing tab-completion for names

klement:
Use libicu, if available, to transliterate candidate strings
to remove accents before doing tab-completion.

This allows e.g.

/msg S<tab>

to match e.g. Slovak "Štefan" without having to deal with local
keyboard layout...

Signed-off-by: Klement Sekera <klement.sekera@gmail.com>


boothj5:
I found a solution without adding any new libraries and a bit less code, it also covers case insensitive (#879) autocompletion.

It might not cover all the transliterations available with this pull request, but ignores accents.

The changes are in master if you want to test it.

klement:
Looks good and works for my use case. Thanks.

PR #926, opened 2017-04-12
server_events: fix potential NULL pointer deref

msantos:


PMaynard:
Applied patch, ran clang static analyses and can confirm this patch addresses the two potential NULL pointer defefs.

jubalh:
Thanks for the patch @msantos

PR #941, opened 2017-06-16
Disable autoping if server doesn't support it

### LINKED ISSUE #859: Only perform autoping when server supports ping requests
Currently if autoping is enabled, profanity will send ping requests.  If the server returns a "cancel" error, the autopings are disabled.

If the server does not support ping requests, and does not return a "cancel" error, profanity will eventually timeout and disconnect (if `/autoping timeout` is enabled).

Before sending ping requests, check that the server supports it, and don't disconnect when it doesn't.  Log an INFO message that autoping is set, but the current server doesn't support it.

asdil12:
fixes #859

This patch has not been tested.

### DIFF
--- a/src/xmpp/iq.c
+++ b/src/xmpp/iq.c
@@ -967,6 +967,15 @@ _autoping_timed_send(xmpp_conn_t *const conn, void *const userdata)
         return 1;
     }
 
+    if (connection_jid_for_feature(STANZA_NS_PING) == NULL) {
+        log_warning("Server doesn't advertise %s feature, disabling autoping.", STANZA_NS_PING);
+        prefs_set_autoping(0);
+        cons_show_error("Server ping not supported, autoping disabled.");
+        xmpp_conn_t *conn = connection_get_conn();
+        xmpp_timed_handler_delete(conn, _autoping_timed_send);
+		return 1;
+    }
+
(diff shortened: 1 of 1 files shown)


boothj5:
Tested and merged, thanks.

Used the `connection_supports()` function which does the same thing, but reads a little better.

asdil12:
But it doesn't test for connention status, does it?

boothj5:
The `_autoping_timed_send()` function checks for connection status before the feature check is made.

asdil12:
Yes, but shouldn't the `connection_supports()` function check it - like the other function?

boothj5:
`connection_jid_for_feature()` checks the hash table is not NULL, which although is usually the case when there is no connection, isn't quite the same as checking the connection status, just the code being defensive.

This NULL check could be added to `connection_supports()` too, which could also lead to `connection_supports()` just calling `connection_jid_for_feature()` and returning FALSE on NULL.  Feel free to do this refactor.

You could argue that the `connection_supports()` function then becomes redundant, but returning a boolean from a function saves the hassle of looking into the function to figure out whether or not the result should be free'd. It's there for convenience.

The general pattern throughout the code is to check the connection status before calling connection related functions.

PR #942, opened 2017-06-19
Improve connection_supports

asdil12:


jubalh:
After talking with @asdil12 we decided to use the macro change but discarded to check for connection_support only on the main component.

Chery picked as 86a9ef7b5c89ab7830903b2c19a08e8174ecd0ca

PR #944, opened 2017-06-24
Legacy SSL support

### LINKED ISSUE #536: Error connecting to pvp.net (legacy SSL)
Problems have been reported trying to connect to the pvp.net chat network.  The settings are used as below, connecting to the US server using the IP address as the altdomain.

Other clients reportedly connect successfully using the same settings.

The error in the log is as follows:

```
29/03/2015 21:18:44: xmpp: DBG: Connecting via altdomain.
29/03/2015 21:18:44: xmpp: DBG: sock_connect to 192.64.174.69:5223 returned 4
29/03/2015 21:18:44: xmpp: DBG: attempting to connect to 192.64.174.69
29/03/2015 21:18:44: xmpp: DBG: connection successful
29/03/2015 21:18:44: conn: DBG: SENT: <?xml version="1.0"?><stream:stream to="pvp.net" xml:lang="en" version="1.0" xmlns="jabber:client" xmlns:stream="http://etherx.jabber.org/streams">
29/03/2015 21:18:44: xmpp: DBG: Socket closed by remote host.
29/03/2015 21:18:44: xmpp: DBG: Closing socket.
29/03/2015 21:18:44: prof: DBG: Connection handler: XMPP_CONN_DISCONNECT
29/03/2015 21:18:44: prof: DBG: Connection handler: Login failed
```

pvp

pasis:
Add `legacy` tls policy to /account and /connect commands. When this
policy is specified the connection is connected with
XMPP_CONN_LEGACY_SSL flag. Notice, legacy SSL and STARTTLS are not
compatible and user has to know when exactly the new policy should be
used.

To enable it, run one of the next commands:
```
  /connect <jid> tls legacy [server <host>]
  /account <name> set tls legacy
```

Notice, there is no SRV record for legacy SSL. Therefore, you may need
`server` property to connect successfully. Refer to configuration
provided by your server.

This patch closes #536.

### DIFF
--- a/src/command/cmd_ac.c
+++ b/src/command/cmd_ac.c
@@ -568,6 +568,7 @@ cmd_ac_init(void)
     tls_property_ac = autocomplete_new();
     autocomplete_add(tls_property_ac, "force");
     autocomplete_add(tls_property_ac, "allow");
+    autocomplete_add(tls_property_ac, "legacy");
     autocomplete_add(tls_property_ac, "disable");
 
     join_property_ac = autocomplete_new();
--- a/src/command/cmd_defs.c
+++ b/src/command/cmd_defs.c
@@ -158,7 +158,7 @@ static struct cmd_t command_defs[] =
             CMD_TAG_CONNECTION)
         CMD_SYN(
             "/connect [<account>]",
-            "/connect <account> [server <server>] [port <port>] [tls force|allow|disable]")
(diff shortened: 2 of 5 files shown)


mrsdizzie:
This works great in my (limited) testing -- I can now connect using the legacy TLS option.

PR #992, opened 2018-04-20
Format /me messages even if own nick is mentioned

jubalh:
Adapting mucwin_message() to format /me messages like they should be
even if own nick appears in them.

Regards: https://github.com/boothj5/profanity/issues/950

### DIFF
--- a/src/ui/mucwin.c
+++ b/src/ui/mucwin.c
@@ -483,7 +483,7 @@ mucwin_message(ProfMucWin *mucwin, const char *const nick, const char *const mes
         ch = mucwin->message_char[0];
     }
 
-    if (g_strcmp0(nick, mynick) != 0) {
+    if ((g_strcmp0(nick, mynick) != 0) && (strncmp(message, "/me ", 4) != 0)) {
         if (g_slist_length(mentions) > 0) {
             win_print_them(window, THEME_ROOMMENTION, ch, nick);
             _mucwin_print_mention(window, message, mynick, mentions);


jubalh:
> Testing locally I found that whilst the message is shown correctly to the sender, the receiver sees the message as if they sent it.

Not sure what you mean here. I just tested again and it looks fine to me:
profan
Logged in in profanity as `m`. Sending from another client as `michael`.

boothj5:
If I run master I see the following:
prof_master

If I run in the branch for this pull request (https://github.com/jubalh/profanity/tree/memention), I see the following:
prof_branch

In both cases the following commands are run:

1. Profanity (as boothj5): `/me says hello`
2. Profanity (as boothj5): `/me says chanleemoon`
3. Pidgin (as chanleemoon): `/me says hello`
4. Pidgin (as chanleemoon): `/me says hello boothj5`

The change from:
```
if (g_strcmp0(nick, mynick) != 0) {
```
to:
```
if ((g_strcmp0(nick, mynick) != 0) && (strncmp(message, "/me ", 4) != 0)) {
```
means that if I receive a message from someone that starts with "/me " then the statement evaluates to false and the else is executed:
```
win_println_me_message(window, ch, mynick, "%s", message);
```
which will print the message as if I sent it (`mynick`).

Perhaps what is slightly confusing is the function name `win_print_me_message` as it sounds like its for printing a message with the `/me ` prefix but what it really means is it's printing a message that I sent, rather than a message someone else sent (which is what `win_print_them_message` is used for).

The handling of the `/me ` prefix is done lower down the call stack...
https://github.com/boothj5/profanity/blob/42dc13275d9868bb69bc909b9640d0e208e89d3f/src/ui/window.c#L1447

I'm not sure why you're seeing different behaviour though, are all your changes pushed to the branch?

If you have the time to look into it further I'll do my best to help, I know that looking through someone else's large code base is not easy!

jubalh:
> I'm not sure why you're seeing different behaviour though, are all your changes pushed to the branch?

I tested again and had still the same results. Then I realized: oh my! I was using two codebases, the one I tested with was the thing I backported to 0.5.1 and not the actual git master. And I didn't backport it correctly because in 0.5.1 actually only `win_print()` is used, no `me` and `them`.
Thats why it worked there.

The reason why I didn't compile on master was that it was easier for me to build the rpm package and apply/backport the patches (I thought..).
Master didn't build because of:

```
./configure: line 12573: syntax error near unexpected token `,'
./configure: line 12573: `ACX_PTHREAD(, AC_MSG_ERROR([pthread is required]))'
```
(autoconf 2.69, automake 1.15.1)

Will look into this now and then continue with this.

My guess is that I will have to check for the `/me` in all the nested if statements and if it is found use the `win_print()` function.

> If you have the time to look into it further I'll do my best to help, I know that looking through someone else's large code base is not easy!

Thank you. Your explanations where very helpful here (especially the `me` vs `them`).

jubalh:
> Will look into this now and then continue with this.
Got it, `autoconf-archive` was missing.

PR #994, opened 2018-04-21
Fail plugin unloading if plugin does not exist

philipflohr:
the plugins_unload function always returned true.
Therefore, plugin unloads for nonexisting plugins succeeded.


jubalh:
Please remove the whitespaces after `return FALSE;` and if possible squash your commits into one.

I'm thinking of merging this so it is in before https://github.com/boothj5/profanity/pull/999, do you agree?

jubalh:
@philipflohr what about this issue? #999 got merged some time ago.

PR #995, opened 2018-04-21
let plugins_unload_all fail if one plugin unload operation fails

philipflohr:
previously TRUE was returned once one unload operation succeeded

### DIFF
--- a/src/plugins/plugins.c
+++ b/src/plugins/plugins.c
@@ -247,7 +247,7 @@ plugins_load(const char *const name)
 gboolean
 plugins_unload_all(void)
 {
-    gboolean result = FALSE;
+    gboolean result = TRUE;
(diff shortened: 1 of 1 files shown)


jubalh:
I'm thinking of merging this so it is in before https://github.com/boothj5/profanity/pull/999, do you agree?

jubalh:
@philipflohr what about this issue? https://github.com/boothj5/profanity/pull/999 got merged some time ago.

jubalh:
@philipflohr what about this issue? https://github.com/boothj5/profanity/pull/999 got merged some time ago.

philipflohr:
Sorry, have been away for a while...
I still think this is a good idea, are there any more changes needed before you can merge this?

jubalh:
done :)

PR #999, opened 2018-05-07
Extended plugin handling

philipflohr:
This PR adds
1. The possibility to uninstall a plugin (unload and delete)
2. The possibility to update a plugin (previously if a plugin was installed there was no possibility to update it)


src/command/cmd_funcs.c
```diff
@@ -6662,6 +6663,81 @@ cmd_plugins_install(ProfWin *window, const char *const command, gchar **args)
...
+        if (!g_str_has_suffix(path, ".py") && !g_str_has_suffix(path, ".so")) {
+            cons_show("Plugins must have one of the following extensions: '.py' '.so'");
+            free(path);
+            return TRUE;
+        }
+
+        GString* error_message = g_string_new(NULL);
+        gchar *plugin_name = g_path_get_basename(path);
+        gboolean result = plugins_unload(plugin_name);
+        result |= plugins_uninstall(plugin_name);
```
pasis: You should use logical operations instead of bitwise here. Also, you need to use `and` instead of `or`. Replace it with `&&`.

philipflohr:
I'm adding your changes if this pr has a chance of getting merged;)
@boothj5 Is there any?

jubalh:
Just my personal opinion here, not trying to answer for boothj5.

I think the goal you are trying to achieve makes sense. So I don't see a reason why it shouldn't be implemented. @pasis has also rights to commit to the repo, so I assume he is reviewing your code to get it in.

boothj5:
Agreed, these features make sense.

If everyone is happy with the code, feel free to merge the pull request.

Thanks @philipflohr

src/command/cmd_funcs.c
```diff
@@ -6662,6 +6663,81 @@ cmd_plugins_install(ProfWin *window, const char *const command, gchar **args)
...
+        if (!g_str_has_suffix(path, ".py") && !g_str_has_suffix(path, ".so")) {
+            cons_show("Plugins must have one of the following extensions: '.py' '.so'");
+            free(path);
+            return TRUE;
+        }
+
+        GString* error_message = g_string_new(NULL);
+        gchar *plugin_name = g_path_get_basename(path);
+        gboolean result = plugins_unload(plugin_name);
+        result && plugins_uninstall(plugin_name);
```
pasis: Ok, now it doesn't compile :)

You want to achieve this logic here:
```
1. unload
2. if unload is successful then uninstall
3. if unload and uninstall are successful then install
```

So the above chain is executed until the 1st fail. One way to implement this:
```
result = unload();
result = result && uninstall();
result = result && install();
```
In the above example uninstall() and install() are called only if `result` is TRUE.

src/command/cmd_funcs.c
```diff
@@ -6662,6 +6663,81 @@ cmd_plugins_install(ProfWin *window, const char *const command, gchar **args)
...
+        } else {
+            cons_show("Failed to install plugin: %s. %s", plugin_name, error_message->str);
+        }
+        g_free(plugin_name);
+
+        free(path);
+        return TRUE;
+    }
+
+    if (is_dir(path)) {
```
pasis: Looks like `path` is not freed here and at the end of the function.

src/command/cmd_funcs.c
```diff
@@ -6662,6 +6663,81 @@ cmd_plugins_install(ProfWin *window, const char *const command, gchar **args)
...
+    }
+
+    if (is_regular_file(path)) {
+        if (!g_str_has_suffix(path, ".py") && !g_str_has_suffix(path, ".so")) {
+            cons_show("Plugins must have one of the following extensions: '.py' '.so'");
+            free(path);
+            return TRUE;
+        }
+
+        GString* error_message = g_string_new(NULL);
```
pasis: `error_message` is not freed anywhere in this function.

src/common.c
```diff
@@ -107,28 +108,22 @@ mkdir_recursive(const char *dir)
...
-
-    FILE *target = fopen(targetpath, "wb");
-    if (target == NULL) {
-        fclose(source);
-        return FALSE;
-    }
-
-    while((ch = fgetc(source)) != EOF) {
-        fputc(ch, target);
+    GFile *source = g_file_new_for_path(sourcepath);
```
pasis: Due to [documentation](https://developer.gnome.org/gio/stable/GFile.html#g-file-new-for-path), `source` and `dest` must be freed with `g_object_unref()`.

src/common.c
```diff
@@ -107,28 +108,22 @@ mkdir_recursive(const char *dir)
...
+    GFile *dest = g_file_new_for_path(targetpath);
+    GError *error = NULL;
+    gboolean success = false;
+    
+    if (overwrite_existing)
```
pasis: Suggestion. This `if` can be replaced with:
```
GFileCopyFlags flags = overwrite_existing ? G_FILE_COPY_OVERWRITE : G_FILE_COPY_NONE;
success = g_file_copy (source, dest, flags, NULL, NULL, NULL, &error);
```

src/plugins/plugins.c
```diff
@@ -149,10 +150,12 @@ plugins_install_all(const char *const path)
     get_file_paths_recursive(path, &contents);
 
     GSList *curr = contents;
+    GString *error_message = NULL;
     while (curr) {
+        error_message = g_string_new(NULL);
```
pasis: `error_message` must be freed.

src/plugins/plugins.c
```diff
@@ -167,26 +170,41 @@ plugins_install_all(const char *const path)
 }
 
 gboolean
-plugins_install(const char *const plugin_name, const char *const filename)
+plugins_uninstall(const char *const plugin_name)
+{
+    plugins_unload(plugin_name);
+    char *plugins_dir = files_get_data_path(DIR_PLUGINS);   
+    GString *target_path = g_string_new(plugins_dir);
```
pasis: `target_path` must be freed.

src/plugins/plugins.c
```diff
@@ -167,26 +170,41 @@ plugins_install_all(const char *const path)
...
+    free(plugins_dir);
+    g_string_append(target_path, "/");
+    g_string_append(target_path, plugin_name);
+    GFile *file = g_file_new_for_path(target_path->str);
```
pasis: `file` must be freed with `g_object_unref()` after `g_file_delete()`.

src/command/cmd_funcs.c
```diff
@@ -6616,12 +6616,13 @@ cmd_plugins_install(ProfWin *window, const char *const command, gchar **args)
             return TRUE;
         }
 
+        GString* error_message = g_string_new(NULL);
```
pasis: `error_message` must be freed.

pasis:
@philipflohr, there are few memory leaks in your patch. Fix them please.

pasis:
Now it looks good. I will check why it fails to build later. Maybe some library is missed.

pasis:
Profanity won't build with this patch, because configure.ac is not complete. This is why Travis job fails. Try this to reproduce:
```
./configure --disable-notifications --disable-icons --disable-otr --disable-pgp --disable-plugins --disable-c-plugins --disable-python-plugins --without-xscreensaver
make
```
Since `g_object_unref()` is in src/common.c now, we need to check for libgio-2.0 properly in configure.ac. Basically, instead of adding `-lgio-2.0` manually, we need to do so via pkg-config.

philipflohr:
@pasis I have no idea how to do this correctly. Could you do this?

pasis:
@philipflohr, sure. But I will have time for fixing and testing in few days. Sorry for delay.

pasis:
I've fixed libgio-2.0 checks and rebased patchset. @philipflohr, I've squashed your fixes into single commit, hope, you're ok with it. @jubalh, merge #1012 if you're fine with changes and current PR can be closed after that.

P.S. sorry for delay, was too busy with RL stuff...

jubalh:
Merged https://github.com/boothj5/profanity/pull/1012

PR #1003, opened 2018-06-26
Add missing library for expect lib check

paulfariello:
On debian one can't link directly on expect without linking on tcl.


jubalh:
@paulfariello I suspect this is not true. The Debian package seems to be build quite normal and doesn't use any such [patch](https://salsa.debian.org/xmpp-team/profanity/tree/debian/master/debian/patches).

Will close this. Feel free to reopen if you want further discussion :)

PR #1010, opened 2018-08-14
Use uuid in create_unique_id instead of counter

jubalh:
Message IDs should be unique so they can be used by XEPs like delivery receipts, chat markers, message correction.

So far it used a counter so restarting profanity will cause the counter
to be 0 again.

Let's rather use an UUID since we have such a function in the
xmpp/xmpp.h already.

Closes https://github.com/boothj5/profanity/issues/998

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -56,14 +56,13 @@
 #include "log.h"
 #include "common.h"
 #include "tools/p_sha1.h"
+#include "xmpp/xmpp.h"
 
 struct curl_data_t
 {
     char *buffer;
     size_t size;
 };
-
-static unsigned long unique_id = 0;
 
 static size_t _data_callback(void *ptr, size_t size, size_t nmemb, void *data);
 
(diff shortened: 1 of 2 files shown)


jubalh:
> Since now the ID generation has a dependency on the src/xmpp folder, it might be good to move the code to this folder too. Generally common.c is used by all major modules (xmpp, ui, encryption, commands, plugins etc) and doesn't have a dependency on them. It's not a big problem, or something that will break, but just trying to keep the dependency tree tidy.

I see. It makes sense. `create_unique_id()` is only used in the xmpp subfolder. The rest of the functions in common.c is used everywhere. Since the ID probably is never needed except in the XMPP subfolder maybe I should move that function to a file in xmpp. I didn't find a suitable one, do you have an suggestion? I could also create `src/xmpp/id.{c, h}`.

boothj5:
Perhaps move the `create_unique_id` function to the `xmpp/connection.c` rename as `connection_create_stanza_id` and expose it in the `xmpp/connection.h` header? Open to suggestions if you have any other ideas.

PR #1011, opened 2018-09-05
Get rid of p_sha1 dependency

jubalh:
Use libstrophe for sha1.
Regarding https://github.com/boothj5/profanity/issues/882


jubalh:
@boothj5 This makes common.c include strophe.h.
Since `p_sha1_hash()` is actually only used in once in the whole code, in `stanza_create_caps_sha1_from_query()` in `src/xmpp/stanza.c`, I thought about making it a non exported function there. But I think then we can not test it in the unit tests.

boothj5:
@jubalh I think moving it out of `common.c` sounds like a good idea.

And yes, anything that uses libstrophe/libmesode directly cannot be unit tested, but I think this is ok, once we have functional tests up and running again we can try to get it tested there.

jubalh:
Travis seems to have a problem with `xmpp_sha1_digest()`. Not sure why it should be in libstrophe. Maybe this test instance uses and old version?

jubalh:
@pasis Yes! Thanks for the comment. I woke up this morning and I figured that this must be the reason :)

Seems like syncing libmesode with libstrophe should be done first.

jubalh:
latest release of libstrophe: 0.9.2
latest release of libmesode: 0.9.1

xmpp_sha1_digest() was added to libstrophe in 0.9.2

I think we need to sync up the projects, but probably requiring >= 0.9.2 would also be good? Or at least mentioning this in the ChangeLog or somewhere so packagers are aware.

@boothj5 I remember asking about libmesode/libstrophe version difference right at the beginning when I asked about joining the project. Maybe now it's time that we take care of this.

@boothj5, @pasis could you guys describe me the process you used to sync the projects? I suppose you did not copy over the changes in one big chunk? It would be good to copy them over and release new libmesode soon. Then we can continue with profanity :)

Help is appreciated! If you don't find the time I can try to do it with your explanation :) Thanks!

jubalh:
@boothj5 told me he will take care of the libmesode sync and release :-)

Will update the configure script to check for the right version.

boothj5:
@jubalh I've done a merge locally for libstrophe -> libmesode and also tested your branch against it which works fine.

I haven't yet pushed the changes to libmesode since the way I merged lost some of the history so I just want to check if there's a better way.  I should hopefully get this done during the week, or worst case next weekend.

jubalh:
@boothj5 great, thank you! Much appreciated.

jubalh:
@boothj5 Thank you!

PR #1017, opened 2018-10-29
Add prof_win_close

twardziejszy:
Added possibility to close a window with key binding. _(So you're less likely to send someone e.g. „c/los” message)_


jubalh:
Thanks @twardziejszy and sorry for the long waiting time.

PR #1020, opened 2018-10-31
tests: remove cflag -w and fix warnings

pasis:
The flag hides errors. But we want unit tests to be correct in order
to reveal errors in the main code.

The patch removes tests_unittests_unittests_CFLAGS and makes autotools
use AM_CFLAGS instead. Therefore, unit tests are built with flags
derived from configure.ac.

@jubalh Please, take a look.


jubalh:
This looks good. Thank you!

PR #1022, opened 2018-11-10
Splash logo improvements

rjc:


jubalh:
Thanks for the suggestion.

This seems to be a quite subjective thing.

Personally I think the 'a' is better in the original version. But I agree on the 'o' maybe.. actually not sure.
So I showed it to a couple of persons and they all think the original person is more readable.

So I personally would decline this PR. But will wait what @pasis or @boothj5 will say.

jubalh:
I understand.
I'll add screenshots so others can add their opinion more easily:

Original:
profanity-original

Proposed Change #1:
profanity-change1

boothj5:
Usually when something subjective like this comes up, I'd suggest adding it as a customisation option.

Default: What we currently have
Allow specifying a splash logo file to customise it.

Would be a little more work though...

rjc:
Ouch! I have clearly made a typo - `a` was **not** meant to look like that! I've taken care while editing the original with but clearly not enough.

What I propose is to leave the *outside* of the `o` as it is and only change the *inside* to brackets, and use the right bracket `)` in `p`, as it is already done in `a` - albeit a mirror image, i.e. left bracket `(`. 

To sum it up, I propose to change three vertical bars `|` to brackets - one in `p` and two in `o`, and narrowing the `r`.

What say you?

jubalh:
I think it looks good. Like I said only the 'a' was an issue for me. Could you do your change and push and/or submit a picture so we can see how it looks?

rjc:
This is what I had in mind.

proposed logo

jubalh:
I like it.

rjc:
Before I do so, I wanted to post all three options. Please pick one you think looks best:

- `o` *flat* and `a` *round*
flat_o_round_a

- both `o` and `a` *flat*
o_and_a_both_flat

- both `o` and `a` *round*
o_and_a_both_round

I'm not sure about the first one any more - it simply makes letters, which are naturally both rounded (`o` and `a`), inconsistently *flat* and *rounded*, respectively, on their left edges which aren't adjacent to any other letter. At the same time, the right edge of `o` looks much better *flat* next to the `f`, the same goes with right edge of `p` - so I wouldn't change right edges of `p` and `o` to rounded ones.

I like both consistency **and** aesthetics so, personally, I'd lean towards the 3rd option - both `o` and `a` rounded on edges which aren't adjacent to any other letters.

What say you?

rjc:
I also had my reservations - it depends entirely on the font. If brackets are *too* rounded, i.e. their radius is short, then it doesn't look too good, especially coupled with the rounded `r` and `f` above them, respectively.

Flat edges look pretty much the same with *any* font. 2nd option, then?

rjc:
OK, I've adjusted the logo accordingly, as well as fixed the issue with `a` that I had introduced accidentally in the first commit.

jubalh:
Thanks @rjc

PR #1029, opened 2019-01-21
Fix optionnal sessionid in xep 0050

paulfariello:
sessionid appears to be optional in command messages.
Handle it correctly.


jubalh:
Thanks!

PR #1036, opened 2019-02-26
Don't stop plugin installation if given path points to a directory

### LINKED ISSUE #1035: Unable to install plugins from directory
Pointing /plugins install <path> to a directory results in Not a file: <path> instead of installing all plugins from the directory

philipflohr:
The command "/plugins install <path_to_folder> should install all
plugins in the given folder. This commit removes the check if the path
points to a file (which is repeated in the section installing a plugin
from file) and makes the existing code installing multiple plugins accessible

Fixes #1035

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -6612,12 +6612,6 @@ cmd_plugins_install(ProfWin *window, const char *const command, gchar **args)
         path = strdup(path);
     }
 
-    if (access(path, R_OK) != 0) {
-        cons_show("File not found: %s", path);
-        free(path);
-        return TRUE;
-    }
-
     if (is_regular_file(path)) {
         if (!g_str_has_suffix(path, ".py") && !g_str_has_suffix(path, ".so")) {
             cons_show("Plugins must have one of the following extensions: '.py' '.so'");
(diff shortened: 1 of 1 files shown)


philipflohr:
As #1035 was no bug, this PR does not really fix something. 
Still it does a little bit of refactoring and removes an IMO unnecessary test.

jubalh:
Will try to look at it soon.

jubalh:
@pasis what's your opinion on this one?

pasis:
This patch won't work as is. `is_regular_file()` doesn't handle errors and this is the issue. When you use `is_regular_file()` or `is_dir()` for non existent path, you will read uninitialized memory and the result will be unpredictable. You can fix that with handling errors:

```C
int
is_regular_file(const char *path)
{
    struct stat st;
    int rc;
    rc = stat(path, &st);
    return rc == 0 && S_ISREG(st.st_mode);
}
```

Plus I would prefer to see different errors for non existent file/dir and when `path` is a special file.

jubalh:
@philipflohr would you like to change this PR according to @pasis comment?

philipflohr:
I'd love to but I don't have spare time at the moment.
If anyone else is interested in doing this I'm totally fine, otherwise you'll  need to wait some months for the changes.

jubalh:
Did https://github.com/profanity-im/profanity/commit/c135f989ec779f1f2d6438c2220156baaf616065 in preperation to merge this.

jubalh:
@philipflohr do you want to rebase?

Issue #1044, opened 2019-03-22
Use after free in ui_leave_room


dbaer:
I keep getting a segfault on close, and the cause appears to be that in `ui_leave_room()`, `roomjid` is a lingering reference to a field of the (now freed) `ProfMucWin`. Fortunately it's an easy fix:

```diff --git a/src/ui/core.c b/src/ui/core.c
--- a/src/ui/core.c
+++ b/src/ui/core.c
@@ -834,12 +834,12 @@ void
 ui_leave_room(const char *const roomjid)
 {
     ProfWin *window = (ProfWin*)wins_get_muc(roomjid);
+    GList *privwins = wins_get_private_chats(roomjid);
     if (window) {
         int num = wins_get_num(window);
         ui_close_win(num);
     }
 
-    GList *privwins = wins_get_private_chats(roomjid);
     GList *curr = privwins;
     while (curr) {
         ProfPrivateWin *privwin = curr->data;
```

pasis:
Good catch! The patch looks good. Please, open pull request.

jubalh:
Fixed by https://github.com/boothj5/profanity/pull/1045

PR #1045, opened 2019-03-22
Fix use after free bug

dbaer:
This fixes issue #1044 - I also found the same pattern in a few other places in src/ui/core.c.


jubalh:
Thanks @dbaer !

PR #1054, opened 2019-04-12
Don't touch OMEMO cryto keys on disconnect if OMEMO wasn't loaded

paulfariello:
If OMEMO wasn't loaded, removing key would create empty identity.txt file. Then at load time it would segfault profanity.
    
Another commit should fix load of malformed identity.txt.


jubalh:
Will wait for second commit before merge :)

jubalh:
For some reason this builds with old travis config and thus fails.

PR #1056, opened 2019-04-12
Build OMEMO support in travis

paulfariello:


paulfariello:
gcrypt >= 1.7.0 is required.

PR #1059, opened 2019-04-13
Add omemo_start_sessions to unittest stub

jubalh:


jubalh:
This fixes the unittest build.
However `prof_whole_occurrences_tests` fails. Locally it works. I wonder whether this is some encoding/locale problem.

jubalh:
Got it :-) Will push tomorrow.

jubalh:
That was it. Had to set a locale.
Yay! Travis unittests are back up again! The first time including OMEMO!

PR #1061, opened 2019-04-14
Handle message without id in MUC

paulfariello:
Some clients seems to send message without id.

Experienced issue with messages coming from Monal.


jubalh:
Thanks!

Issue #1052, opened 2019-04-12
OMEMO chat flagged as unencrypted when started from another client


pmaziere:
Debian testing, Profanity master HEAD, XMPP server prosody

I enabled carbon copy on all my XMPP clients
When I start an OMEMO encrypted  chat from my Conversations client, profanity opens a window for this chat , but it is flagged as "unencrypted" instead of "OMEMO" despite each message is prefixed with the OMEMO char.

[profanity.log](https://github.com/boothj5/profanity/files/3072277/profanity.log)

That's not a functional issue, but it is still confusing
Thanks for looking at it

jubalh:
@pmaziere thanks for reporting.

Just for the record, at the time of this writing HEAD points at 61df0c8e8513a1aa9912e37019a63778ec3ed06c
(HEAD is like a variable, you need to mention the hash/revision :-))

kaffeekanne:
@pmaziere i suppose you did not start an OMEMO session with the chat partner from profanity? Maybe that issue can be solved alongside with omemo policy.

jubalh:
@pmaziere did you use `/omemo start`?

kaffeekanne:
That also happens, when you receive an omemo message from someone else. Had that recently with @paulfariello. Initiating a message with `/msg` looks up in `trust.txt` and starts/resumes the omemo session. Externally started conversations do not at the moment.

pmaziere:
@jubalh no I did not, but I trusted the fingerprints before opening the chat
@kaffeekanne that's what I observed: chat window started from another client is not flagged as OMEMO, but messages are still prefixed with the OMEMO char, which is inconsistent. I don't know if the OMEMO indicator should be trusted from the window flag or from the message prefix

paulfariello:
This issue can be fixed alongside #1040

paulfariello:
@pmaziere the OMEMO window flag means that you are going to send OMEMO encrypted messages. The OMEMO char means that message either received or sent were OMEMO encrypted.

jubalh:
@pmaziere please give master a try! :)

pmaziere:
I set policy to always, quit , then restart profanity

case 1:
- starting myself a chat from other client => OMEMO tagged window in profanity

quit and restart

case 2:
- contact starts a chat => unencrypted tagged window in profanity
         reply from other client => window is still tagged unencrypted but message appears prefixed with OMEMO char
[...]

PR #1062, opened 2019-04-16
Add OMEMO policy

### LINKED ISSUE #1040: Use OMEMO by default
Once #1039 is merged (@paulfariello great work, many thanks for your efforts!), I think it would be great if profanity would default to OMEMO messaging.

This includes to automatically run `/omemo gen` if necessary and run `/omemo start buddy@example.org` automatically as soon as a conversation is started.

If OMEMO messaging is not available on the own server or the remote server profanity should indicate that the session is not using OMEMO (like as an example Conversations does).

Additionaly I'd suggest to implement the possibility to manually stop the omemo encryption in order to communicate with clients without OMEMO support.

paulfariello:
There is 3 policy:

- manual: OMEMO session are only started manually
- automatic: OMEMO session are only started if they have been started
  manually before
- always: OMEMO session are always started unless they have been ended
  manually before

This should fix #1040 and #1052

### DIFF
--- a/src/command/cmd_ac.c
+++ b/src/command/cmd_ac.c
@@ -164,6 +164,7 @@ static Autocomplete otr_log_ac;
 static Autocomplete otr_policy_ac;
 static Autocomplete omemo_ac;
 static Autocomplete omemo_log_ac;
+static Autocomplete omemo_policy_ac;
 static Autocomplete connect_property_ac;
 static Autocomplete tls_property_ac;
 static Autocomplete alias_ac;
@@ -591,11 +592,17 @@ cmd_ac_init(void)
     autocomplete_add(omemo_ac, "untrust");
     autocomplete_add(omemo_ac, "fingerprint");
     autocomplete_add(omemo_ac, "clear_device_list");
+    autocomplete_add(omemo_ac, "policy");
 
     omemo_log_ac = autocomplete_new();
     autocomplete_add(omemo_log_ac, "on");
     autocomplete_add(omemo_log_ac, "off");
     autocomplete_add(omemo_log_ac, "redact");
+
+    omemo_policy_ac = autocomplete_new();
+    autocomplete_add(omemo_policy_ac, "manual");
+    autocomplete_add(omemo_policy_ac, "automatic");
+    autocomplete_add(omemo_policy_ac, "always");
 
     connect_property_ac = autocomplete_new();
     autocomplete_add(connect_property_ac, "server");
@@ -1080,6 +1087,7 @@ cmd_ac_reset(ProfWin *window)
     autocomplete_reset(otr_policy_ac);
     autocomplete_reset(omemo_ac);
     autocomplete_reset(omemo_log_ac);
+    autocomplete_reset(omemo_policy_ac);
     autocomplete_reset(connect_property_ac);
     autocomplete_reset(tls_property_ac);
     autocomplete_reset(alias_ac);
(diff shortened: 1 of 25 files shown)


jubalh:
I like the term 'automatic. In OTR the same thing was called 'opportunistic'. So we could either let them differ since they are slightly different. We could use 'automatic' in both OMEMO and OTR or use 'opportunsitic' in OMEMO to be the same as OTR. Not being a native english speaker I'm not sure which term is more appropriate.

jubalh:
@paulfariello please dont forget to fix the unit tests.

jubalh:
One more :)

```
tests/unittests/test_cmd_otr.c: In function ‘cmd_otr_gen_generates_key_for_connected_account’:
tests/unittests/test_cmd_otr.c:181:28: error: too few arguments to function ‘account_new’
     ProfAccount *account = account_new(account_name, "me@jabber.org", NULL, NULL,
                            ^~~~~~~~~~~
In file included from ./src/config/accounts.h:41,
                 from ./src/otr/otr.h:41,
                 from tests/unittests/test_cmd_otr.c:13:
./src/config/account.h:71:14: note: declared here
 ProfAccount* account_new(const gchar *const name, const gchar *const jid,
```

jubalh:
Maybe "This should fix #1040 and #1052" should be added to commit message too.

jubalh:
From http://profanity.im/otr.html:

>  *manual* - The default. Unencrypted messaging is allowed, OTR sessions must be started manually.
*opportunistic* - Upon sending an unencrypted message with the /msg command, profanity will attempt to start an OTR session with the contact. The contact's policy must be set to opportunistic or always for the session to start.
*always* - An error message will be displayed if you try to start a conversation without starting an OTR session and the message will not be sent. 

So opportunistic is actually trying if it is possible. Our automatic is more like 'remembering' that we started before and thus enabling automatically again. So I think 'automatic' is good.

optmzr:
I pulled in this PR to test and I'm having a weird issue, when I've set `/omemo always` and the OTR policy is set to manual, every time I join a MUC an OTR session is started (I expect an OMEMO session). I've never had OTR enabled for this room before.

paulfariello:
Thanks @optmzr for reporting the issue.

The OTR thing is an issue. I'm not building OTR so haven't seen this behavior yet.
The @Unsubscribed has been encountered by @kaffeekanne too.

The log about old counter is a known behavior with MAM on MUC.

paulfariello:
Fixed the `@Unsubscribed` issue with MUC.

There is still an issue with policy `always` and autojoined MUC.
We need to have check for MUC anonymous state. Otherwise `always` policy and autojoined MUC will result in inconsistent behavior.

Not checking for MUC anonymous state is a unreferenced issue right now. @jubalh can we reference it?

jubalh:
> Not checking for MUC anonymous state is a unreferenced issue right now. @jubalh can we reference it?

@paulfariello yes sure!

paulfariello:
@optmzr I can't reproduce the behavior you had with OTR starting when joining MUC.
Can you test it with latest commits?

paulfariello:
@jubalh I agree concerning the `automatic` definition. We should keep it like that since behavior is a bit different from OTR.

optmzr:
Yep, I'll check.

Den ons 17 apr. 2019 12:05Paul Fariello <notifications@github.com> skrev:

optmzr:
@paulfariello It works now with the latest commit!

PR #1063, opened 2019-04-16
Handle presence received before roster

### LINKED ISSUE #1050: Contacts are marked offline despite being online
Everything was fine with 0.6.0 from debian testing

I removed the debian package, compiled and installed 61df0c8e8513a1aa9912e37019a63778ec3ed06c with OMEMO support to test it.

Now, after connection, contacts are marked offline despite being online as shown by another xmpp client (conversations). 
Some of them are set online when they start chatting, while others stay offline despite being engaged in a chat.
All chats are OMEMO enabled.

I could not find a way to log debugging information, so if you need more, I'll be glad to help given that you provide me with a mean to do so :)

Thanks for this long awaited OMEMO enabled profanity !

paulfariello:
Presence of contact not found in roster are filtered out.
But sometimes roster is received after a first few presences.

We choose to store presences until we receive roster and then process
this presences.

This should fix #1050 

Unit test are currently failing. Once fixed PR will be mergeable.

### DIFF
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -115,6 +115,7 @@ sv_ev_roster_received(void)
         ui_show_roster();
     }
 
+    roster_process_pending_presence();
     char *account_name = session_get_account_name();
 
 #ifdef HAVE_LIBGPGME
--- a/src/xmpp/roster_list.c
+++ b/src/xmpp/roster_list.c
@@ -67,7 +67,15 @@ typedef struct prof_roster_t {
     GHashTable *group_count;
 } ProfRoster;
 
+typedef struct pending_presence {
+    char *barejid;
+    Resource *resource;
+    GDateTime *last_activity;
+} ProfPendingPresence;
+
 static ProfRoster *roster = NULL;
+static gboolean roster_received = FALSE;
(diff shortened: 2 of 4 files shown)


jubalh:
Seems all good. Thanks!

jubalh:
I wonder why this issue hasnt been raised before. Also the bugreporter confirms that in 0.6.0 everything was fine. So this seems to has to do with changes after 0.6.0, probably OMEMO?
I don't quite see the connection though.

kaffeekanne:
I noteiced that behaviour sometimes too, but did not dig into it as other things were beeing worked on and i had no memory if it was behaving correct in current/old releases. should i test further before merge?

paulfariello:
AFAICT it's not linked. Nevertheless, one side effect of OMEMO is that it is doing more stuff at startup. So it may delay the roster request a bit and thus let more time to the server to send first presences.

jubalh:
@kaffeekanne @paulfariello can you test this PR if you have some time?

kaffeekanne:
`\o/` looks fine. All my previous offline  contacts are online now. Thats for the short test in my bubble at least.

PR #1072, opened 2019-04-18
Add /occupants indent option

jubalh:
Regards https://github.com/boothj5/profanity/issues/690

@kaffekanne asked for this.

### DIFF
--- a/src/config/theme.c
+++ b/src/config/theme.c
@@ -452,6 +452,11 @@ _load_preferences(void)
         prefs_set_occupants_size(occupants_size);
     }
 
+    if (g_key_file_has_key(theme, "ui", "occupants.indent", NULL)) {
+        gint occupants_indent = g_key_file_get_integer(theme, "ui", "occupants.indent", NULL);
+        prefs_set_occupants_indent(occupants_indent);
+    }
+
     if (g_key_file_has_key(theme, "ui", "roster.size", NULL)) {
         gint roster_size = g_key_file_get_integer(theme, "ui", "roster.size", NULL);
         prefs_set_roster_size(roster_size);
--- a/src/ui/occupantswin.c
+++ b/src/ui/occupantswin.c
@@ -46,26 +46,34 @@ _occuptantswin_occupant(ProfLayoutSplit *layout, Occupant *occupant, gboolean sh
     theme_item_t presence_colour = theme_main_presence_attrs(presence_str);
     wattron(layout->subwin, theme_attrs(presence_colour));
 
-    GString *msg = g_string_new(" ");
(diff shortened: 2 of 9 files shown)


jubalh:
@kaffeekanne please give it a try!

kaffeekanne:
awesome, thanks! Indenting works, but autocompletion `/occupants in<tab>` does not. `/occupants show jid` is indented 4 spaces now, i think that was ment to be 2 spaces off the nick before it was configurable. I dont need JIDs in that list, but since we have that feature, it should look correct.

Some equivalent to `/roster header char none|<char>` would complete that new behaviour would be highly appreciated. I set headers to another colour to distinguish them from nicks.

jubalh:
> awesome, thanks! Indenting works, but autocompletion /occupants in<tab> does not.

Done now :)

> /occupants show jid is indented 4 spaces now, i think that was ment to be 2 spaces off the nick before it was configurable.

I adjusted it to be always one space more that occupants indent. I think that's enough.

kaffeekanne:
So if the group char could be turned off this PR would be ready for merging.

src/ui/occupantswin.c
```diff
@@ -46,18 +46,37 @@ _occuptantswin_occupant(ProfLayoutSplit *layout, Occupant *occupant, gboolean sh
     theme_item_t presence_colour = theme_main_presence_attrs(presence_str);
     wattron(layout->subwin, theme_attrs(presence_colour));
 
-    GString *msg = g_string_new("   ");
+    GString *spaces = g_string_new(" ");
+
+    int indent = prefs_get_occupants_indent();
+    int current_indent = 0;
```
paulfariello: What is the purpose of `current_indent`?

src/config/theme.c
```diff
@@ -452,6 +452,11 @@ _load_preferences(void)
         prefs_set_occupants_size(occupants_size);
     }
 
+    if (g_key_file_has_key(theme, "ui", "occupants.indent", NULL)) {
+        gint occupants_indent = g_key_file_get_integer(theme, "ui", "occupants.indent", NULL);
+        prefs_set_occupants_indent(occupants_indent);
```
paulfariello: Doesn't it overwrites the user preference?

src/ui/occupantswin.c
jubalh: Right, going to remove that. I thought about passing this to win_sub_print() at first.

src/config/theme.c
jubalh: Yeah it does. But so far it always does overwrite the userprefs with the theme one if you load a theme.

PR #1073, opened 2019-04-23
Fix indentation time confing

kaffeekanne:
Small fix of indention visible when usig `/prefs`. Also first PR! \o/


jubalh:
Nice!

Now you should update the commit message `git commit --amend`. And use `Fix indention for time confing` instead of `Update console.c`. Then do `git push -f yourremotename`.

jubalh:
Now it looks like "John Doe" committed :-)
Also you fixed one typo in 'config' (before 'comfig') and added a new one instead 'confing'.

kaffeekanne:
next time i will post an issue again 😫

jubalh:
Better. Now we have `indention` instead of `indentation` though :-)

PR #1074, opened 2019-04-23
Occupants header char

jubalh:
Regards https://github.com/boothj5/profanity/issues/690

### DIFF
--- a/src/command/cmd_ac.c
+++ b/src/command/cmd_ac.c
@@ -183,6 +183,8 @@ static Autocomplete form_field_multi_ac;
 static Autocomplete occupants_ac;
 static Autocomplete occupants_default_ac;
 static Autocomplete occupants_show_ac;
+static Autocomplete occupants_header_ac;
+static Autocomplete occupants_char_ac;
 static Autocomplete time_ac;
 static Autocomplete time_format_ac;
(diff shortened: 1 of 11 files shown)


kaffeekanne:
Works for me, no idea, what travis complains about. Theming not tested, as i have not created any custom theme before.

#1072 and #1074 need to be mentioned in documentation, maybe as an checklist issue forget about it?

jubalh:
> #1072 and #1074 need to be mentioned in documentation, maybe as an checklist issue forget about it?

What do you mean by this? Mentioned where exactly?

kaffeekanne:
Are http://profanity.im/themes.html and http://profanity.im/reference.html#occupants autogenerated or hand made in https://github.com/boothj5/www_profanity_im ?

jubalh:
Like mentioned before https://github.com/boothj5/www_profanity_im contains everything you need. I'm unsure what you mean with autogenerated. It's not markdown transformed to html if that's what you mean. We write the html by hand. Otherwise the script/markdown would be in there and not plain html pages ;)

jubalh:
I created the commit to change the docu now. But I guess in the future I'll do it only once the tracker issue is done. So I only have to edit the file once :-)

PR #1080, opened 2019-04-24
OMEMO: Check if ID is present before using it

jubalh:
Fixes https://github.com/boothj5/profanity/issues/1079


jubalh:
@paulfariello I think this should be enough. Can you reiview? :)

Issue #690, opened 2015-12-07
Add more customisation options to occupants panel


boothj5:
Similar to the recent additions to the `/roster` panel (https://github.com/boothj5/profanity/issues/665), add further customisations to the chat room `/occupants` panel.

Ideas:
1) Enable/disable word wrapping
2) Option to show presence and status message
3) Options to group/order by presence

jubalh:
Added some more features to the occupants panel. I think they make the most sense. Will close this issue for now and tackle more improvements in new issues/PRs if someone requests it or starts to work on it.

PR #1084, opened 2019-05-02
Add occupants char

jubalh:
`/occupants char` now sets a character that is displayed before the nick
in MUCs. Similar like `/roster char` is displaying a char for the roster.

Regards #690


jubalh:
Oh seems we lost travis since the move to `profanity-im` organization. @boothj5 could you add it back again? :-)

kaffeekanne:
Have not tested theming, otherwise works as intended. I understood theming needs testing beyond thos and the last PRs anyway.

PR #1091, opened 2019-05-08
Fix unit tests on 32-bit architectures

frnky:
The first commit should fix the build issues on 32-bit architectures, which debacle mentioned in the MUC.
The macro mock_ptr_type does an intermediate cast to uintptr_t which silences the compiler warnings.

The second commit is not necessary to fix the unit tests on 32-bit architectures but improve the coding style by using a similar macro to mock non-pointer types.


jubalh:
Thanks!

PR #1098, opened 2019-05-21
Force UI update before generating OMEMO crypto materials

optmzr:
The message wouldn't show up before the generation of OMEMO crypto materials started.


jubalh:
Thanks!

PR #1106, opened 2019-06-03
Don't clear saved account data in session_disconnect()

pasis:
If connection loss occurs, it calls session_disconnect() eventually.
This function clears saved account data which is required for
reconnection. Therefore, when reconnect timer expires, we get errors:

```
02/06/2019 04:53:42: stderr: ERR: (profanity:17115): GLib-CRITICAL **: 04:53:42.305: g_key_file_has_group: assertion 'group_name != NULL' failed
02/06/2019 04:53:43: prof: ERR: Unable to reconnect, account no longer exists: (null)
```

To solve it, don't clear the saved data in session_disconnect(). It will
be cleared properly on connection loss if reconnect timer is not
configured. But won't be cleared with /disconnect command.
So, after /disconnect the data will live in memory until the next
/connect.

Also, remove some copy-paste in connection loss path.

### DIFF
--- a/src/xmpp/session.c
+++ b/src/xmpp/session.c
@@ -113,14 +113,11 @@ session_connect_with_account(const ProfAccount *const account)
 
     log_info("Connecting using account: %s", account->name);
 
+    _session_free_saved_account();
+    _session_free_saved_details();
+
     // save account name and password for reconnect
-    if (saved_account.name) {
-        free(saved_account.name);
-    }
     saved_account.name = strdup(account->name);
-    if (saved_account.passwd) {
-        free(saved_account.passwd);
-    }
     saved_account.passwd = strdup(account->password);
 
     char *jid = NULL;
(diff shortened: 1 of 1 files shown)


src/xmpp/session.c
```diff
@@ -196,20 +196,6 @@ session_connect_with_details(const char *const jid, const char *const passwd, co
 void
 session_autoping_fail(void)
 {
-    if (connection_get_status() == JABBER_CONNECTED) {
-        log_info("Closing connection");
-
-        char *account_name = session_get_account_name();
-        const char *fulljid = connection_get_fulljid();
-        plugins_on_disconnect(account_name, fulljid);
```
jubalh: I think we should still call this to make sure plugins can react on this?

pasis: It is still called, I just removed redundant copy-paste:

session_autoping_fail() -> session_lost_connection() -> sv_ev_lost_connection() -> ev_disconnect_cleanup() -> session_disconnect().

And in the last function:
```C
    // if connected, send end stream and wait for response
    if (connection_get_status() == JABBER_CONNECTED) {
        log_info("Closing connection");

        char *account_name = session_get_account_name();
        const char *fulljid = connection_get_fulljid();
        plugins_on_disconnect(account_name, fulljid);
```

jubalh: Oh my bad! Thanks!

jubalh:
awesome @pasis! Thanks a lot!

Issue #1100, opened 2019-05-22
Use travis matrix to test more systems


jubalh:
We should test more systems (which are up to date enough to have our dependencies).

- [ ] Arch Linux
- [ ] Debian Testing
- [x] openSUSE Tumbleed
- [ ] Mac OS X

We use Docker so we dont limit ourselves to the distributions that travis supports. I'm not sure whether that will create problems for us when we try to test OS X.

We will need separate docker files to init each system, update the travis.yml to use a build matrix and call the common travis-build.sh.

paulfariello:
OS X is supported by travis.

jubalh:
But I'm not sure if you can mix using docker with what they natively support.

optmzr:
@jubalh Could I give this a try?

jubalh:
@optmzr Sure!

AFAIK if you create a PR with the changed travis files it should then use those files for that PR so we can test it.
And maybe this helps: https://media.ccc.de/v/1354-continuous-integration-using-travis-and-docker/

Snapper uses docker and this [travis.yml](https://github.com/openSUSE/snapper/blob/master/.travis.yml) file shows how they do it.
In their case they have various dockerfiles, like [Dockerfile.Debian](https://github.com/openSUSE/snapper/blob/master/Dockerfile.debian) etc.
So basically we need the same to install the different OSes and then call `./travis-build.sh` in each of those.

PR #1108, opened 2019-06-04
Use travis matrix to test more systems (#1100)

optmzr:
## Add Travis CI tests for Arch, Debian and OSX/macOS

+ Arch and Debian are run in Docker containers, as openSUSE Tumbleweed.

+ OSX/macOS doesn't use any containers.
   * Homebrew is used to fetch all the dependencies.
   * The dependencies are declared in the Brewfile.travis file.

+ The travis-build.sh script has been modified to check for the current
  OS and the different configure flags has been moved into an array
  that'll be looped through instead.
  The xscreensaver (for libXScrnSaver) flags has been removed for macOS
  as it only makes sense for systems running X11, which macOS doesn't
  (usually) do.

+ Some minor shellcheck fixes, too.

Fixes: #1100


jubalh:
:-)

jubalh:
I hand you the travis osx hero badge now!

optmzr:
@jubalh Only took 36 commits :sunglasses:

optmzr:
Yes! I'll fix :)

optmzr:
@jubalh Fixed.

I also noted that Travis CI has support for Windows too:
https://blog.travis-ci.com/2018-10-11-windows-early-release
https://travis-ci.community/c/environments/windows

Here's one project doing it:
https://travis-ci.org/yargs/yargs/jobs/439923030

Maybe something for the future? :)

jubalh:
Cool! Yeah maybe :-)

I don't have any Windows machine but I planned to bug a friend to test the current Window documentation.
I'm not sure how Windows+Cygwin play together but it seems that this was the preferred way to install profanity.


> https://blog.travis-ci.com/2018-10-11-windows-early-release

So they use chocolatey, which seems to be nice. It doesn't have libxml packaged yet (which we need for libstrophe/libmesode). Probably more stuff missing.

I thought maybe telling windows users to install the Linux subsystem for Windows (and install openSUSE) so they only need to do a `zypper in profanity` might be an alternative for people who don't want to build themselves.

jubalh:
@optmzr thanks a lot for the work on this! :)

PR #1110, opened 2019-06-05
Feature/704 ui behaviour reconnect

jubalh:


kaffeekanne:
have not tested this PR, but keep in mind we could lose messages that are sent between disconnection and reconnection. fetching the history and comparing with alreaddy written chatlogs is probably a lot and error prone work, but what do you think about it? would be poormans MAM ;)

jubalh:
Yep you are right about this :)

I save the timestamp of the last message we received in MUC (general for all MUCs not for each one), and when we rejoin and we get the MUC history I check if the messages have newer timestamps. In which case they get displayed.
In my tests it worked :)

paulfariello:
Could we use message id to filter which history messages we have to display ?

jubalh:
> Could we use message id to filter which history messages we have to display ?

Then we would have to save all IDs since they are not numerated. Sounds a bit much right?
But based on your input I saw that I could implement this at another postion (src/xmpp/message.c) where we already test for the timestamp. Thanks!

paulfariello:
You are right and we can't use id if they are not in received MUC history. But they will be required when we'll have MAM and OMEMO support.

jubalh:
I think I'll merge this tomorrow.

Issue #1117, opened 2019-06-06
omemo fingerprint autocompletion leads to crash


jubalh:
```
#0  0x00007f5b0f89bd8b in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f5b0f885549 in __GI_abort () at abort.c:79
#2  0x00007f5b0f885421 in __assert_fail_base
    (fmt=0x7f5b0f9ebf48 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x4a86e6 "roster != NULL", file=0x4a86c0 "src/xmpp/roster_list.c", line=498, function=<optimized out>)
    at assert.c:92
#3  0x00007f5b0f894262 in __GI___assert_fail
    (assertion=0x4a86e6 "roster != NULL", file=0x4a86c0 "src/xmpp/roster_list.c", line=498, function=0x4a8910 <__PRETTY_FUNCTION__.12661> "roster_contact_autocomplete") at assert.c:101
#4  0x000000000042e309 in roster_contact_autocomplete
    (search_str=0x7fffe2854c70 "", previous=0) at src/xmpp/roster_list.c:498
#5  0x000000000048833c in autocomplete_param_with_func
    (input=0x23dfc90 "/omemo fingerprint ", command=0x4c71f3 "/omemo fingerprint", func=0x42e2d5 <roster_contact_autocomplete>, previous=0) at src/tools/autocomplete.c:273
#6  0x0000000000482d97 in _omemo_autocomplete
    (window=0x24ed0b0, input=0x23dfc90 "/omemo fingerprint ", previous=0)
    at src/command/cmd_ac.c:2188
#7  0x0000000000481142 in _cmd_ac_complete_params
    (window=0x24ed0b0, input=0x23dfc90 "/omemo fingerprint ", previous=0)
    at src/command/cmd_ac.c:1539
#8  0x000000000047f4ae in cmd_ac_complete
    (window=0x24ed0b0, input=0x23dfc90 "/omemo fingerprint ", previous=0)
    at src/command/cmd_ac.c:991
#9  0x0000000000453a1b in _inp_rl_tab_handler (count=1, key=9) at src/ui/inputwin.c:534
#10 0x00007f5b0faf4276 in _rl_dispatch_subseq
    (key=9, map=<optimized out>, got_subseq=<optimized out>) at ../readline.c:862
[...]

PR #1118, opened 2019-06-06
Only complete certain omemo commands if connected

jubalh:
Some of the omemo commands depend on the roster being present.
Several of those functions call `assert()` if that's not the case.

Modify omemo autocompletion in such a way that only things that don't
need the roster will be completed. Only works on first level.

When typing `/omemo` it will still complete to `/omemo start` but not
offer suggestions for that, but print a message that you need to be
connected.

Fix https://github.com/profanity-im/profanity/issues/1117


jubalh:
What I forgot to mention is that this problem occurs if one run `/omemo something` *before* connecting.

PR #1124, opened 2019-06-07
Remove OMEMO session when untrusting a fingerprint

paulfariello:
We have to store known_devices in a long term fashion otherwise we might
not be able to delete session if it is remove from device list server
side.

Also ensure that OMEMO is removed from trusted storage.


jubalh:
Thanks!

Issue #1122, opened 2019-06-07
Decrypt OMEMO messages before initial trust


kaffeekanne:
Right now OMEMO encrypted messages get decrypted only if we trust the sending key fingerprint. I see no harm in decrypting even if we don't trust ouer opponent yet, bcause we only gain information but don't risk anything as we still don't send messages encrypted for that opponents device.

This way somebody can trust our key, send us encrypted messages, that we can read and answer encrypted, if we want to. 

This is not blind trust before verification, BTBV!

paulfariello:
We should at least clearly show that messages are untrusted. Something like red text. Otherwise impersonation is not prevented.

jubalh:
Its possible that the user defined red to be the normal text color for his theme. But I don't know of a better way.
Maybe: Print text red and notice in titlebar too where we currently have [OMEMO] -> [OMEMO untrusted].

kaffeekanne:
i think it can not be [OMEMO untrusted], but [untrusted] alone. [OMEMO] means, we have started an OMEMO session, but has nothing to do with trusting keys. also you can trust one key, but not the other (sending) one.

My idea append/prepend the actual message with [untrusted] because of the mentioned problems. 

Second idea, i like in this context: colorize the encryption/omemo char. green received from trusted device id, red receved from untrusted device id. If we actively stored untrusted device id fingerprints, we cour even use yellow for "not decided" fingerprints. thats how gajim vizualizes fingerprints of new device ids, before sending or not sending encrypted to them. that would chime in nicely with gajims behaviour.

optmzr:
As @kaffeekanne, appending the messages with `[Untrusted]` should be "perfectly" annoying so that the user will make sure to trust the user.

PR #1126, opened 2019-06-07
Decrypt all incoming omemo msg

paulfariello:


paulfariello:
Addresses #1122

jubalh:
For the record:
Currently messages are mostly passed as char*. We planned to have a ProfMessage struct to add more meta information that we sometimes need (like stanza Id, read state). This will make it easier to later implement last message correction or changing messages in the buffer. Also printing lines for last read or messages in different colours if they are unread.
This was planned for 0.8.0.

But since with this PR we already ran into something it was decided to create a first version of them now already. So we can store the enc_type of a message along with the message instead of passing as seperate parameter.

src/xmpp/message.h
```diff
@@ -35,9 +35,34 @@
...
+   /* The raw body from xmpp message, either plaintext or OTR encrypted text */
+   char *body;
+   /* The encrypted message as for PGP */
+   char *encrypted;
+   /* The message that will be printed on screen and logs */
+   char *plain;
+   GDateTime *timestamp;
+   prof_enc_t enc;
+   gboolean trusted;
+} prof_message_t;
```
jubalh: Actually I would do it like this:
typedef struct prof_message_t {
...
} ProfMessage;

I think all other structs also follow that convention. For example see prof_chat_win_t -> ProfChatWin.

src/xmpp/message.c
```diff
@@ -999,38 +1040,39 @@ _handle_chat(xmpp_stanza_t *const stanza)
...
 
     // standard chat message, use jid without resource
-    xmpp_ctx_t *ctx = connection_get_ctx();
-    GDateTime *timestamp = stanza_get_delay(stanza);
-    if (!message && body) {
-        message = xmpp_stanza_get_text(body);
+    prof_message_t *message = message_init();
+    message->jid = jid;
+
+    message->timestamp = stanza_get_delay(stanza);
```
jubalh: Will this always have a value or is it possible that sometimes we dont get one and need to get the receiving time ourselves? Of course this then could differ from when it was sent..

Issue #632, opened 2015-10-01
Option of setting the Time format for all chat activity types


MalsR:
Could we have an option to set a default time format for all the different activity types as opposed to setting it manually please? /time all set "%H:%M:%S"

e.g. 
/time statusbar set "%H:%M:%S"
/time chat set "%H:%M:%S"

clarkem:
Is this just a matter of changing` PREF_TIME_STATUSBAR` in `_get_default_string()` ?

boothj5:
I know the request mentions "default", but if I understand correctly I think the request is not to change the defaults for each setting, but to be able to change all settings in one command for all window types.

So instead of:
```
/time console set <format>
/time chat set <format>
/time muc set <format>
/time mucconfig set <format>
/time private set <format>
/time xml set <format>
```

As a short cut the user can do:
```
/time all set <format>
```

Excluding `statusbar` and `lastactivity` settings since they are not for main windows, but used slightly differently.

PR #1129, opened 2019-06-17
Add option to set all window related time formats

jubalh:
Implements feature requested at: https://github.com/profanity-im/profanity/issues/632

So instead of:

/time console set <format>
/time chat set <format>
/time muc set <format>
/time mucconfig set <format>
/time private set <format>
/time xml set <format>

As a short cut the user can do:

/time all set <format>

Excluding statusbar and lastactivity settings since they are not for main windows, but used slightly differently.


jubalh:
Arch, Debian, OSX give green light. TW repos are slow today. Lets not wait for it.

PR #1130, opened 2019-06-19
Fix memory leaks

jubalh:
Draft PR for now. More fixes will follow.

### DIFF
--- a/src/omemo/omemo.c
+++ b/src/omemo/omemo.c
@@ -305,6 +305,7 @@ omemo_on_disconnect(void)
     g_key_file_free(omemo_ctx.trust_keyfile);
     g_string_free(omemo_ctx.sessions_filename, TRUE);
     g_key_file_free(omemo_ctx.sessions_keyfile);
+    _g_hash_table_free(omemo_ctx.session_store);
     g_string_free(omemo_ctx.known_devices_filename, TRUE);
     g_key_file_free(omemo_ctx.known_devices_keyfile);
 }
--- a/src/omemo/omemo.c
+++ b/src/omemo/omemo.c
@@ -387,6 +387,7 @@ omemo_start_sessions(void)
             const char *jid = p_contact_barejid(contact);
             omemo_start_session(jid);
         }
+        g_slist_free(contacts);
     }
 }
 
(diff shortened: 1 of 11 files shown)


jubalh:
valgrind seems to produce some false positives always when using glib: https://stackoverflow.com/questions/4254610/valgrind-reports-memory-possibly-lost-when-using-glib-data-types

But even running with: `G_DEBUG=gc-friendly G_SLICE=always-malloc valgrind --leak-check=full --log-file=out ./profanity` I get some messages about:

```
==14661== 7 bytes in 1 blocks are definitely lost in loss record 494 of 2,730                                                                                 
==14661==    at 0x483777F: malloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)                                                                  
==14661==    by 0x5B839F9: strdup (strdup.c:42)                                                                                                               
==14661==    by 0x479DF0: theme_init (theme.c:114)                                                                                                            
==14661==    by 0x42B6D9: _init (profanity.c:190)                                                                                                             
==14661==    by 0x42B6D9: prof_run (profanity.c:98)                                                                                                           
==14661==    by 0x4282A0: main (main.c:172)
```

To my knowledge the relevant code is correct.
Calling g_hash_table_new_full() followed by inserting g_hash_table_insert() and at the end g_hash_table_destroy().

I run glib2-devel in version 2.60.3 and there appears a relevant bugreport about glib itself having a leak:
https://gitlab.gnome.org/GNOME/glib/issues/1749 which is planned to be fixed in 2.61.4.

jubalh:
I was on 2.60.2. 2.60.3 already helps a lot.
There is also a suppression file coming with it. So calling valgrind like this makes sense:

```
G_DEBUG=gc-friendly G_SLICE=always-malloc valgrind --tool=memcheck --leak-check=full --leak-resolution=high --num-callers=30 --show-leak-kinds=definite --log-file=valout --suppressions=/usr/share/glib-2.0/valgrind/glib.supp ./profanity
```

jubalh:
Valgrind still shows some readline related memory leaks. I'm not yet sure if they are false positives or real ones.

And there are a couple of more OMEMO leaks but I feel like someone with more knowledge can fix this better than I can. @paulfariello would you mind taking a look? With the valgrind line in the comment above you probably are able to quickly see whats wrong and how to fix it.

PR #1132, opened 2019-06-21
Fix NULL pointer when handling non encrypted message in MUC

paulfariello:


jubalh:
Thanks for the quick response!

Issue #1083, opened 2019-05-02
Roster assertion failure kills profanity with SIGABRT


ccxcz:
Built profanity from commit 722cb5fdc3ced8b76b58992326c8761227240202
```
Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007f65eb48fff0 in __GI_abort () at abort.c:79
#2  0x00007f65eb4864ea in __assert_fail_base (fmt=0x7f65eb5ebf18 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x55e4f9083994 "roster != NULL",     
    file=file@entry=0x55e4f908397d "src/xmpp/roster_list.c", line=line@entry=158, function=function@entry=0x55e4f9083c20 <__PRETTY_FUNCTION__.12916> "roster_get_contact")   
    at assert.c:92
#3  0x00007f65eb48656a in __GI___assert_fail (assertion=assertion@entry=0x55e4f9083994 "roster != NULL", file=file@entry=0x55e4f908397d "src/xmpp/roster_list.c",            
    line=line@entry=158, function=function@entry=0x55e4f9083c20 <__PRETTY_FUNCTION__.12916> "roster_get_contact") at assert.c:101                                            
#4  0x000055e4f8fdd4b5 in roster_get_contact (barejid=<optimized out>) at src/xmpp/roster_list.c:158                                                                         
#5  roster_get_contact (barejid=<optimized out>) at src/xmpp/roster_list.c:156
#6  0x000055e4f8ff9860 in _display_name (tab=<optimized out>, tab=<optimized out>) at src/ui/statusbar.c:523                                                                 
#7  0x000055e4f8ff9ff3 in _tabs_width () at src/ui/statusbar.c:481
#8  status_bar_draw () at src/ui/statusbar.c:265
#9  0x000055e4f8ff658d in ui_update () at src/ui/core.c:140
#10 0x000055e4f8fdaa94 in prof_run (log_level=<optimized out>, account_name=<optimized out>) at src/profanity.c:129                                                          
[...]

PR #1145, opened 2019-07-04
[FIX] triggering assert after disconnecting from ethernet

1g0rb0hm:
On macOS, everytime a disconnect from the docking station happens and the ethernet connection disappears, **profanity** crashed with the below stack trace.

I am not too familiar with the code. However, it appears that the **assert** is a bit too aggressive in this particular case and we should not **abort** but rather mark the presence update as pending.

```
Application Specific Information:
Assertion failed: (roster != NULL), function roster_get_contact, file src/xmpp/roster_list.c, line 135.

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff6cd352c6 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff6cdf0bf1 pthread_kill + 284
2   libsystem_c.dylib             	0x00007fff6cc9f6a6 abort + 127
3   libsystem_c.dylib             	0x00007fff6cc6820d __assert_rtn + 324
4   profanity                     	0x0000000105c396a6 roster_get_contact + 100
5   profanity                     	0x0000000105c5264d _display_name + 70
6   profanity                     	0x0000000105c51e1e status_bar_draw + 644
7   profanity                     	0x0000000105c4efa6 ui_update + 364
8   profanity                     	0x0000000105c372d3 prof_run + 524
9   profanity                     	0x0000000105c8b79b main + 392
10  libdyld.dylib                 	0x00007fff6cbfa3d5 start + 1
```

### DIFF
--- a/src/xmpp/roster_list.c
+++ b/src/xmpp/roster_list.c
@@ -120,12 +120,10 @@ roster_destroy(void)
 gboolean
 roster_update_presence(const char *const barejid, Resource *resource, GDateTime *last_activity)
 {
-    assert(roster != NULL);
-
     assert(barejid != NULL);
     assert(resource != NULL);
 
-    if (!roster_received) {
+    if (roster == NULL || !roster_received) {
         ProfPendingPresence *presence = malloc(sizeof(ProfPendingPresence));
         presence->barejid = strdup(barejid);
         presence->resource = resource;


jubalh:
Which version of profanity are you running?
I think this is a duplicate of https://github.com/profanity-im/profanity/issues/1083 fixed in https://github.com/profanity-im/profanity/pull/1109

jubalh:
Closing due to inactivity and because I strongly suspect we addressed this already.

PR #1154, opened 2019-07-14
 Always check for directory changes with sendfile auto completion

### LINKED ISSUE #1099: Auto completion for /sendfile doesn't show new files
Steps to reproduce:
 1. Create a directory with a single file, `test1`.
 2. Start Profanity and write `/sendfile <dir>` and press tab to cycle through the available files.
 3. While Profanity is running, create a new file `test2`.
 4. Re-try step 2, the new file doesn't appear.

Workaround: Do autocompletion for a different directory and then with the original directory, and it'll show the new file.

I guess this is due to this line here:
https://github.com/profanity-im/profanity/blob/883dbe1911766e31e4483eabe713edb89e1eec4f/src/command/cmd_ac.c#L1340

Where the directory won't be re-opened if the `last_directory` matches with the current.

optmzr:
Instead of only checking for files when 'last_directory' has changed, do
it every time.

Add autocomplete_update function that updates the items while retaining
last_found and search_str.

Fixes #1099

### DIFF
--- a/src/command/cmd_ac.c
+++ b/src/command/cmd_ac.c
@@ -1290,11 +1290,15 @@ cmd_ac_uninit(void)
     autocomplete_free(statusbar_show_ac);
 }
 
+static void
+_filepath_item_free(char **ptr) {
+    char *item = *ptr;
+    free(item);
+}
+
 char*
 cmd_ac_complete_filepath(const char *const input, char *const startstr, gboolean previous)
 {
-    static char* last_directory = NULL;
-
     unsigned int output_off = 0;
 
     char *result = NULL;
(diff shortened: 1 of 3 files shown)


src/command/cmd_ac.c
```diff
@@ -1330,6 +1330,7 @@ cmd_ac_complete_filepath(const char *const input, char *const startstr, gboolean
             return NULL;
         }
     }
+    // XXX: Why free and then use?
```
jubalh: inpcp is just a char pointer. So we free the last strdup()-ed space and now use inpcp to point to a new location.

src/command/cmd_ac.c
```diff
@@ -1339,56 +1337,56 @@ cmd_ac_complete_filepath(const char *const input, char *const startstr, gboolean
...
+                if (asprintf(&acstring, "%s/%s", directory, dir->d_name) == -1) {
+                    free(foofile);
+                    return NULL;
                 }
-                autocomplete_add(filepath_ac, acstring);
-                free(acstring);
             }
-            closedir(d);
+
+            char *acstring_cpy = strdup(acstring);
```
jubalh: acstring_cpy is never freed. Maybe the array has a free func?

optmzr: I believe `g_array_free(files, TRUE);` further down should free these?

jubalh: valgrind still shows it as allocated

src/command/cmd_ac.c
```diff
@@ -1290,11 +1290,15 @@ cmd_ac_uninit(void)
     autocomplete_free(statusbar_show_ac);
 }
 
+void
+filepath_item_free(char **ptr) {
```
jubalh: I think this can be a static function and prepended with `_`.

src/command/cmd_ac.c
jubalh: Much better now!

jubalh:
Thanks for working on this!

Issue #1159, opened 2019-07-20
Crash when adding/removing a contact to/from a roster group


mdosch:
When adding or removing a contact to/from a roster group via `/group add group contact` or `/group remove group contact` profanity crashes.

```
Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff66ab535 in __GI_abort () at abort.c:79
#2  0x00007ffff6702508 in __libc_message (action=action@entry=do_abort, 
    fmt=fmt@entry=0x7ffff680d28d "%s\n") at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007ffff6708c1a in malloc_printerr (
    str=str@entry=0x7ffff680eff8 "double free or corruption (out)")
    at malloc.c:5341
#4  0x00007ffff670a730 in _int_free (av=0x7ffff6844c40 <main_arena>, 
    p=0x555555d63090, have_lock=<optimized out>) at malloc.c:4306
#5  0x00005555555a098d in _free_group_data (data=0x55555a167f90)
    at src/xmpp/roster.c:341
#6  0x000055555559033d in _iq_id_handler_free (handler=0x555555d60410)
    at src/xmpp/iq.c:259
#7  0x00007ffff71c6c5e in ?? ()
   from target:/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#8  0x00005555555901d3 in _iq_handler (conn=0x555555b2b8e0, stanza=0x55555a14fce0, 
    userdata=0x555555ceb970) at src/xmpp/iq.c:217
#9  0x00007ffff6879ca5 in handler_fire_stanza (conn=conn@entry=0x555555b2b8e0, 
    stanza=stanza@entry=0x55555a14fce0) at src/handler.c:124
#10 0x00007ffff687695a in _handle_stream_stanza (stanza=0x55555a14fce0, 
    userdata=0x555555b2b8e0) at src/conn.c:1179
#11 0x00007ffff688270f in _end_element (userdata=0x555555cddf60, 
    name=<optimized out>) at src/parser_expat.c:190
[...]

PR #1160, opened 2019-07-22
Fix segfault in group add/remove command

jubalh:
Also regarding destructor.
Fix https://github.com/profanity-im/profanity/issues/1159


jubalh:
@mdosch please test :)

mdosch:
I confirm the fix works so I closed #1159.

PR #1161, opened 2019-07-23
Fix memleaks in OMEMO

jubalh:


jubalh:
@paulfariello can you review?

PR #1174, opened 2019-08-20
Don't render (all) delayed messages as MUC history

weiss:
Double-check that a `<delay/>` tag on a groupchat message was actually added by the MUC service (rather than the sending client) before assuming it was received from the MUC history.

### DIFF
--- a/src/xmpp/stanza.c
+++ b/src/xmpp/stanza.c
@@ -1174,6 +1174,12 @@ stanza_create_caps_sha1_from_query(xmpp_stanza_t *const query)
 GDateTime*
 stanza_get_delay(xmpp_stanza_t *const stanza)
 {
+    return stanza_get_delay_from(stanza, NULL);
+}
+
+GDateTime*
+stanza_get_delay_from(xmpp_stanza_t *const stanza, gchar **from)
+{
     GTimeVal utc_stamp;
     // first check for XEP-0203 delayed delivery
     xmpp_stanza_t *delay = xmpp_stanza_get_child_by_name(stanza, STANZA_NAME_DELAY);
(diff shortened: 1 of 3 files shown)


src/xmpp/stanza.c
```diff
@@ -1201,11 +1210,17 @@ stanza_get_delay(xmpp_stanza_t *const stanza)
...
                 g_date_time_unref(utc_datetime);
+                if (from) {
+                    *from = g_strdup(xmpp_stanza_get_attribute(x, STANZA_ATTR_FROM));
+                }
                 return local_datetime;
             }
         }
     }
 
+    if (from) {
```
jubalh: Why do we need this?

weiss: As I mentioned in the room, this patch is just a quick hack to test/show the required behavior change, it's not necessarily meant to be merged as-is.  The patch needs two values returned from `stanza_get_delay()` (the timestamp and the `from`), and from a quick glance at the code it wasn't obvious to me how you typically do that.  So I just went for returning `from` via an out parameter, which the caller can set to `NULL` if he's not interested in it.  `if (from)` checks whether it's `NULL`.

jubalh: Yes, but why would we need to set it to `NULL`  if it's set?

jubalh: Ok I see it now. We could also do `gchar *from = NULL` so the `g_free(from);` is okay, but your way is saver.

weiss: Ah.  This is the case where we didn't find a `from` tag because we didn't find a delay attribute at all.  Could be omitted if we're sure no caller will ever look at `from` if the function returned `NULL` (i.e. right now it could safely be omitted).

jubalh:
Thanks for the patch @weiss !

PR #1177, opened 2019-08-23
Add 256 colors support

aaptel:
Themes can now use color names from the xterm color name list [1].

1: https://jonasjacek.github.io/colors/


jubalh:
Awesome! Thanks a lot!

PR #1187, opened 2019-09-11
Add color for main.text.history

mdosch:
Use the color schemes default for history.


jubalh:
I don't think that actually makes a difference does it?

PR #1195, opened 2019-09-20
Add coloring of statusbar.current tab in view

quite:
This theme color applies to the tab title text of the statusbar tab that
is currently shown.

The (somewhat confusingly named) `statusbar.active` theme color now
applies to all other tabs (before, it applied to all tabs).

Coloring of a tab that is highlighted/has new messages is done as before
using the `statusbar.new` theme color.

The default color is set to `green`.


quite:
I did not add this item to all the current themes...

src/config/theme.c
```diff
@@ -99,6 +99,7 @@ theme_init(const char *const theme_name)
     g_hash_table_insert(defaults, strdup("statusbar.text"),          strdup("white"));
     g_hash_table_insert(defaults, strdup("statusbar.brackets"),      strdup("cyan"));
     g_hash_table_insert(defaults, strdup("statusbar.active"),        strdup("cyan"));
+    g_hash_table_insert(defaults, strdup("statusbar.current"),       strdup("green"));
```
jubalh: please use cyan here, so that it behaves like before for users.

src/ui/statusbar.c
```diff
@@ -337,6 +337,9 @@ _status_bar_draw_extended_tabs(int pos)
         pos = _status_bar_draw_bracket(is_current, pos, "[");
 
         int status_attrs = theme_attrs(THEME_STATUS_ACTIVE);
```
jubalh: Maybe do:
int status_attrs;
and then use if/else.

jubalh:
First of all: thanks for creating an issue and then also the pull request with the change!
Generally I looks very good. I also like how you documented everything and how you wrote the commit message. Thanks for this!

src/ui/statusbar.c
quite: I guess then I should not change *from* that variant in the other place, as I did?

quite: I'll change any way you prefer. I just did it the way it's more readable for me, starting generic, going towards specific.

jubalh: It would just be nice not to assign a value twice if we can go around it.

quite:
Done! Don't forget profanity-im/profanity-im.github.io#35

jubalh:
Thanks for your contribution!

PR #1197, opened 2019-09-24
Fix typo

mdosch:
`Fix crash when jid has no not part` should probably be `Fix crash when jid has no node part`.


jubalh:
Thanks!

PR #1198, opened 2019-09-25
Update theme solarized dark

mdosch:
The theme `solarized dark` is updated to play well with
the newly added `statusbar.active` from #1195


jubalh:
Thanks!

Issue #1201, opened 2019-10-06
User messages double logged in MUC logs


Shungy:
* Enable MUC logging
* Write something
* Check the log file
* What you write is double-logged

v. 0.7.1

jubalh:
We log both outgoing MUC messages and incoming MUC messages.
Logging only the latter should suffice.

jubalh:
With the current fix we also filter out messages sent from ourselves by other clients.

jubalh:
To solve this nicely we will need https://github.com/profanity-im/profanity/issues/1207 XEP-0359: Unique and Stable Stanza IDs.
Then we can filter incoming messages based on origin-id.

jubalh:
@Shungy should be fixed now!

Shungy:
Great! Thank you.

Issue #1207, opened 2019-10-16
Support XEP-0359: Unique and Stable Stanza IDs


jubalh:
[XEP-0359](https://xmpp.org/extensions/xep-0359.html)

This will help us identify which messages in a MUC were sent by us (https://github.com/profanity-im/profanity/issues/1201), and also will be useful later when we implement XEP-0313 MAM (https://github.com/profanity-im/profanity/issues/660) and XEP-0308 Last Message Correction (https://github.com/profanity-im/profanity/issues/805) too.

jubalh:
We don't implement </stanza-id> yet, because we don't use it until we have MAM.

jubalh:
https://github.com/profanity-im/profanity/issues/1207 merged

PR #1209, opened 2019-10-18
XEP-0359

jubalh:
This implements https://github.com/profanity-im/profanity/issues/1207.
And is used to fix https://github.com/profanity-im/profanity/issues/1201

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -489,3 +489,23 @@ get_file_paths_recursive(const char *path, GSList **contents)
         entry = g_dir_read_name(directory);
     }
 }
+
+char*
+get_random_string(int length)
+{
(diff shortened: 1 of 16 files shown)


jubalh:
It seems with the changes done here we get a crash in: `src/xmpp/message.c` `message_handlers_init()` when looking up handlers: `ProfMessageHandler *handler = g_hash_table_lookup(pubsub_event_handlers, curr->data);`.

Steps to reproduce:
* open Profanity and connect
* `/autoping set 10`
* `/autoping timeout 10`
* stop WiFi/connection
* wait for `Lost connection`
* restart wifi
* `/connect`

PR #1213, opened 2019-10-21
Change default text colors to default

### LINKED ISSUE #535: Text color on white background is unreadable
The text color on white backgrounds is a light gray and doesn't match the command line default, i.e. apparently the profanity user interface doesn't adapt to the terminal theme like, say, irssi.

I didn't find any theme which overrides the text color in the distributed themes.

IMHO the default profanity theme should not define colors but should use the terminal-defined ones, as other CLI apps do. It would be an usability boost in the default setup for the many terminals which use a black on white color scheme by default (like gnome-terminal and konsole do)

The version I use is 0.4.6.

mdosch:
Using "default" as color for the main window and input 
area text instead of "white" makes it better usable
with terminals using black/dark text on white/light
background while still being usable with white/light
text on black/dark background.

May fix #535

### DIFF
--- a/src/config/theme.c
+++ b/src/config/theme.c
@@ -72,17 +72,17 @@ theme_init(const char *const theme_name)
 
     defaults = g_hash_table_new_full(g_str_hash, g_str_equal, free, free);
 
-    g_hash_table_insert(defaults, strdup("main.text"),               strdup("white"));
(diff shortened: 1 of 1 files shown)


mdosch:
# Before

2019-10-21-202300_scrot

# After

## Black on white

2019-10-21-204226_scrot

## White on black

2019-10-21-204259_scrot

jubalh:
Super! Thank you!

PR #1217, opened 2019-10-29
omemo: Check stanza names when iterating nodes

aszlig:
Some clients (eg. PSI) are sending the stanzas delimited by whitespace text nodes, which will fail while looping through the `<prekeys/>` children and also print weird errors when iterating through the `<list/>` of devices.

When debugging this, I was looking at the XML of Gajim and PSI and first was somehow confused why Profanity printed `OMEMO: received device without ID` while the XML looked identical (minus the actual IDs and the JIDs of course).

However, Gajim was sending the XML without whitespace nodes in between and PSI did not, so for example the following (with the relevant whitespace nodes marked with `X`):

```xml
  <message type="headline" to="..." from="...">
    <event xmlns="http://jabber.org/protocol/pubsub#event">
      <items type="headline" node="eu.siacs.conversations.axolotl.devicelist">
        <item id="...">
          <list xmlns="eu.siacs.conversations.axolotl">
          X <device id="..."/>
          X <device id="..."/> X
          </list>
        </item>
      </items>
    </event>
    <delay xmlns="urn:xmpp:delay" stamp="..." from="..."/>
  </message>
```

... would result in three times the `OMEMO: received device without ID` error, because we actually have three XML nodes here.

Now since the `<list/>` children above aren't really a problem and only annoying, text nodes in the `<prekeys/>` stanza actually cause `omemo_start_device_session_handle_bundle` to return failure.

I've fixed this by explicitly matching the stanza names we are interested in, skipping everything else.

Reported-by: @devhell


jubalh:
Thanks for your contribution! Looks good. Waiting for a quick review from @paulfariello.

Issue #200, opened 2013-07-13
irssi down arrow behaviour


alandrees:
In irssi, the down arrow stores the current line in history and blanks the input line.  Due to the similarity of interfaces, I find myself looking for the same functionality.

Ppjet6:
I still have this problem on Arch with aur/profanity-git (4789.44979ac-1)

jubalh:
In case we switch for https://github.com/profanity-im/profanity/issues/228 away from using readline for history we also need to be aware of this here.

kaffeekanne:
I like the idea of that feature, although one could use the terminal multiplexer/emulators copy paste function, to achieve this. But how would that down-arrowed message/command be brought back to the input line? Down arrow again if input line is empty, append to text on input line, over write current input line?

PR #1219, opened 2019-10-31
Store current input line in history

jubalh:
Regards https://github.com/profanity-im/profanity/issues/200

This doesn't work yet. And I have no idea why.
Weird behaviour:

- start profanity
- type 'ASDF'
- ctrl+arrow down
-> text vanishes (like intended)
- arrow up
-> nothing happens (intended is that the last history item [ASDF]
appears)

- type 'ABC'
- press enter
- arrow up
-> ABC appears
- enter

- type 'UUU'
- ctrl+arrow down
- type 'ZZZ'
- enter
- arrow up
- ZZZ appears
- arrow up
- UUU appears

So in the latter case we added to history and deleted from the input
line and then immediately entered new text and pressed enter, to add
this to the history too.
When we do this the not sent text succesfully was stored in history.

### DIFF
--- a/src/ui/inputwin.c
+++ b/src/ui/inputwin.c
@@ -127,6 +127,7 @@ static int _inp_rl_win_pagedown_handler(int count, int key);
 static int _inp_rl_subwin_pageup_handler(int count, int key);
 static int _inp_rl_subwin_pagedown_handler(int count, int key);
 static int _inp_rl_startup_hook(void);
+static int _inp_rl_down_arrow_handler(int count, int key);
 
 void
 create_input_window(void)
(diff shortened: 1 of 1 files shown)


jubalh:
I have no idea why the history behaves like it does. Hopefully someone else can take a look!

jubalh:
@boothj5 maybe you have more experience with this. Do you see why this doesn't work as expected?

jubalh:
Merged in 330ef3bcf3f78b0e3844c042b480c8484e22f777 and ba0f92522247eab264bf2b316adfe3754d26a818

PR #1222, opened 2019-11-08
Add (experimental) solarized-light theme

mdosch:
This theme should (when finished) look well
with the solarized-light color scheme.
It is still very rough and needs a lot of polishing
but I will try to improve it when I find time.
Otherwise it might be start for someone
using solarized-light to improve it.


jubalh:
Thanks!

Issue #1220, opened 2019-11-01
Profanity is sometimes highlighting a mention wrong.


mdosch:
Profanity is configured to highlight my Nickname for mentions but sometimes it highlights something wrong.

![2019-11-01-141849_scrot](https://user-images.githubusercontent.com/5103003/68027801-96271500-fcb3-11e9-86d0-b5aa3288432c.png)

## Expected Behavior
In this example it should highlight "Martin"

## Current Behavior
It highlights other stuff and also mixes in the username "debacle" maybe caused by the [<200e>](https://en.wikipedia.org/wiki/Left-to-right_mark).

## Context
```
01/11/2019 14:17:25: xmpp: DBG: RECV: <message id="e7de6887-7feb-4925-ba8a-39ae4a0c3379" type="groupchat" lang="en" to="REDACTED" from="chat@dino.im/debacle"><body>I hope, this quoting style never makes it to XMPP.
<200e>&gt; [13:00:19] <200e>Martin<200e>: Yeah, at least we get paid for that. :D
<200e>&gt; &gt; [12:58:30] <200e>Ge0rG<200e>: Martin: I know. My personal and my business behaviors in that regard are opposite as well
&gt; <200e>&gt; &gt; [12:55:25] <200e>Martin<200e>: Ge0rG: In my company they want TOFU as the expect to be able to scroll through the full history if needed...</body><origin-id id="e7de6887-7feb-4925-ba8a-39ae4a0c3379" xmlns="urn:xmpp:sid:0"/><active xmlns="http://jabber.org/protocol/chatstates"/></message>
```

## Environment
Version 0.7.1dev.master.f71de61b
Debian Bullseye (testing)

```
gnome-terminal --version                                                                      :(
# GNOME Terminal 3.34.2 using VTE 0.58.2 +BIDI +GNUTLS
```

jubalh:
Another example:

Gajim output:
```
‎[04:06:29 PM] ‎jubalh‎: DebXWoody, not yet. Does it work?
‎[04:07:17 PM] ‎debacle‎: ‎[15:04:27] ‎Martin‎: debacle: Let's throw some stuff at me, let's see if it happens again.
```
![issue](https://user-images.githubusercontent.com/1658215/68034564-29674700-fcc2-11e9-9e5d-41e845201269.png)

mdosch:
It seems these <200e> are coming from Gajim. This excerpt is from the Gajim MUC:

```
[...]

PR #1226, opened 2019-11-12
Filter RTL unicode characters out

jubalh:
Gajim sends \u200E and \u200F for RTL.
It is planned that Gajim stops doing this and uses some GTK feature to
get the same result.

However users expressed the whish that we filter out such characters in
incoming messages before displaying them to make Profanity more robust.

I'm still not sure whether I like the solution because it means a lot of
allocating/deallocating upon every new message.

Fix https://github.com/profanity-im/profanity/issues/1220

### DIFF
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -68,6 +68,8 @@
 
 #include "ui/ui.h"
 
+static void _clean_incoming_message(ProfMessage *message);
+
 void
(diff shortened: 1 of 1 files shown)


jubalh:
@mdosch can you test whether this works?

jubalh:
Merged with https://github.com/profanity-im/profanity/commit/ccb4d919b182963eab247bde27a4d39d08cb572f

Issue #1247, opened 2020-01-01
OMEMO 1on1 chats with psi+ fail due to prekey="true/1"


svensp:
## Expected Behavior
I can encrypt 1on1 chats with psi+ users with omemo

## Current Behavior
OMEMO Messages are discarded, silently in case of the UI.

Log says
```
905 01/01/2020 17:45:39: prof: DBG: OMEMO: decrypting message with existing session
906 01/01/2020 17:45:39: prof: ERR: OMEMO: cannot deserialize message
907 01/01/2020 17:45:39: prof: ERR: OMEMO: cannot decrypt message key
```
despite the keys in the message being tagged with `prekey="1"`

## Possible Solution
Checking for `"1"` as well as `"true"` makes the on encryption work both ways but only after the psi+ client sent a message.
```c
323         key->prekey =                                                           
324            (g_strcmp0(xmpp_stanza_get_attribute(key_stanza, "prekey"), "true") == 0)
325            || (g_strcmp0(xmpp_stanza_get_attribute(key_stanza, "prekey"), "1") == 0);
```
A quick google on the type `xs:boolean` looks to me like this should be an expected value here and it is not psi+ breaking the spec

## Steps to Reproduce (for bugs)
1. Turn on debug logging in Profanity 0.7.1
2. Initiate OMEMO between PSI+ 1.4.983 and Profanity(0.7.1)

## Context
I'm trying to use OMEMO encryption between my profanity 0.7.1 on Debian bullseye and a friend who uses Psi+ under Windows

## Environment
Debian bullseye package version 0.7.1-1
```
Profanity, version 0.7.1
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.13.1-unknown)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (2.7.17)
GTK icons: Enabled
```

jubalh:
@svensp I highly appreciate that you open a bug and *also* a pull request to address this bug! Very nice! Thanks!

[...]

PR #1248, opened 2020-01-01
OMEMO message keys allow prekey="1" alongside prekey="true"

svensp:
Short version: the prekey type `xs:boolean` allows `1` as true value alongside `true` but it is not currently accepted when checking whether an omemo message is encrypted with a prekey or not.

Long version: check out issue #1247 for the full motivation behind this.

### DIFF
--- a/src/xmpp/omemo.c
+++ b/src/xmpp/omemo.c
@@ -359,7 +359,9 @@ omemo_receive_message(xmpp_stanza_t *const stanza, gboolean *trusted)
 
         key->data = g_base64_decode(key_text, &key->length);
         free(key_text);
-        key->prekey = g_strcmp0(xmpp_stanza_get_attribute(key_stanza, "prekey"), "true") == 0;
+        key->prekey = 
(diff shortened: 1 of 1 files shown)


jubalh:
LGTM. @paulfariello what do you say?

jubalh:
I only have one request: please ammend the commit message to be more precise.
Use header that sais a bit more, prefix it with `omemo:`.
Describe what you described here in this issue in the actual body of the commit message and also write either: `Regards https://github.com/profanity-im/profanity/issues/1247` or `Fixes https://github.com/profanity-im/profanity/issues/1247`.

svensp:
New commit with more info.
I opted for Regards because encryption only starts working after psi+ sent a message. So profanity users can't initiate an omemo encrypted conversation. I'm not sure if this is actually the same issue but I'd like to keep looking for it in #1247 for now

jubalh:
Looks much better now. Thanks!

paulfariello:
Thanks for fixing this issue. It seems alright.

PR #1252, opened 2020-01-11
Fix crash with malformed OMEMO keys

paulfariello:


src/xmpp/omemo.c
```diff
@@ -371,15 +394,26 @@ omemo_receive_message(xmpp_stanza_t *const stanza, gboolean *trusted)
...
-    g_list_free_full(keys, (GDestroyNotify)omemo_key_free);
-    g_free(iv_raw);
-    g_free(payload_raw);
-    g_free(iv_text);
-    g_free(payload_text);
+out:
+    if (keys) {
+        g_list_free_full(keys, (GDestroyNotify)omemo_key_free);
+    }
+    if (iv_raw) {
```
jubalh: here and on the following lines: free(NULL) is a noop. Isn't that true for g_free too?

paulfariello: Right, it's a noop too.

PR #1260, opened 2020-01-31
Fix OMEMO fingerprint autocompletion

paulfariello:
Only autocomplete fingerprint related to selected user.

Add a context to `autocomplete_with_func` so we can pass the selected jid.


jubalh:
other possibility would have been to have autocomplete_param_with_func_and_context() and let autocomplete_param_with_func() call this one with context parameter to set to NULL.
and only in omemo case actually call autocomplete_param_with_func_and_context().
Less API changes :)
But it's okay if you don't want to change this back.

PR #1263, opened 2020-02-05
Add default pref for PREF_COLOR_NICK

paulfariello:
Fix a segfault if corresponding preference is not saved in config file.


jubalh:
What did you do to get a segfault?

jubalh:
Just:
```
/occupants color on
```
segfaults.

Thanks paul!

PR #1270, opened 2020-02-14
Disallow sendfile in e2ee chat sessions

moppman:
Currently, the /sendfile command sends files via http_upload, but these files are not encrypted. 
However, users that have e2ee enabled in a chat session reasonably expect file encryption. 
That is why we should disallow /sendfile in e2ee sessions until file encryption is implemented.

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -4796,17 +4796,42 @@ cmd_sendfile(ProfWin *window, const char *const command, gchar **args)
         free(filename);
         return TRUE;
     }
-    
-    ProfChatWin *chatwin = (ProfChatWin*)window;
-    assert(chatwin->memcheck == PROFCHATWIN_MEMCHECK);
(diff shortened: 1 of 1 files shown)


src/command/cmd_funcs.c
```diff
@@ -4796,6 +4796,16 @@ cmd_sendfile(ProfWin *window, const char *const command, gchar **args)
         free(filename);
         return TRUE;
     }
+    
+    ProfChatWin *chatwin = (ProfChatWin*)window;
+    assert(chatwin->memcheck == PROFCHATWIN_MEMCHECK);
+    
+    if (chatwin->pgp_send || chatwin->is_omemo || chatwin->is_otr) {
```
jubalh: this doesn't have to be chatwin. Could also be mucwin or privwin.

jubalh:
Thanks for the contribution @moppman !

PR #1278, opened 2020-02-26
Fix few memory leaks

pasis:


src/main.c
```diff
@@ -175,7 +175,17 @@ main(int argc, char **argv)
         return 0;
     }
 
-    prof_run(log, account_name, config_file, log_file);
+    prof_run(log ? log : "INFO", account_name, config_file, log_file);
+
+    /* Free resources allocated by GOptionContext */
+    if (log)
```
jubalh: `g_free(NULL)` is a noop. So we don't need the `if` AFAIK.

PR #1282, opened 2020-03-19
SQLite

jubalh:
I plan to save all messages in an SQLite db.
For retrieving information it's nicer than having it in a text file.
    
This will also be good for later MAM
    
Regular text files will still be an option for users so that they caneasily grep them and do whatever they like.
    
Internally Profanity will only use the SQLite db.
So scroling through history should be easier to do (even without MAM). Getting the last ID of a certain chat should be doable, so once we start to implement MAM we can say "give me all new messages since ID X".
Searching should be easier to do too. 

And all the stuff that I currently don't think about.

We could also log the replace-id so that we can show previous text before a correction occured.

*TODO*: one database for account/jid? Or one table per account/jid?


jubalh:
Logging everything now.
mark_read will be done later. It marks whether a message has been read so we know how to display it in case the user has a setting for it.
archive_id will be added later, will be needed for MAM. So will be added when we implement MAM.

Now we probably should split the tables.

jubalh:
Ready for review/testing.

Issue #1294, opened 2020-03-30
Travis CI build fail


jubalh:
The build for openSUSE Tumbleweed currently fails with:

```
Step 11/17 : RUN ./bootstrap.sh

 ---> Running in 5623e9843763

aclocal: warning: couldn't open directory 'm4': No such file or directory

libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'build-aux'.

libtoolize: copying file 'build-aux/ltmain.sh'

libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.

libtoolize: copying file 'm4/libtool.m4'

libtoolize: copying file 'm4/ltoptions.m4'

libtoolize: copying file 'm4/ltsugar.m4'

libtoolize: copying file 'm4/ltversion.m4'

libtoolize: copying file 'm4/lt~obsolete.m4'

libtoolize: Consider adding '-I m4' to ACLOCAL_AMFLAGS in Makefile.am.

configure.ac:9: installing 'build-aux/compile'

configure.ac:9: installing 'build-aux/config.guess'

configure.ac:9: installing 'build-aux/config.sub'

configure.ac:8: installing 'build-aux/install-sh'

configure.ac:8: installing 'build-aux/missing'

Makefile.am: installing 'build-aux/depcomp'

Removing intermediate container 5623e9843763

 ---> 9e63698f896c

Step 12/17 : RUN ./configure --prefix=/usr --disable-dependency-tracking

 ---> Running in b268d2532325

ls: cannot access '.': Operation not permitted

configure: error: working directory cannot be determined

The command '/bin/sh -c ./configure --prefix=/usr --disable-dependency-tracking' returned a non-zero code: 2

The command "if [ "$TRAVIS_OS_NAME" = "linux" ]; then docker build -f Dockerfile."$BUILD_FLAVOR" -t profanity .; fi" failed and exited with 2 during .

Your build has been stopped.
```

openSUSE Tumblewed seems to be too cutting edge for Travis. It uses coreutils 8.32 (2020-03-05) which has the following change:

```
  stat and ls now use the statx() system call where available, which can
  operate more efficiently by only retrieving requested attributes.
```

We use Travis with Ubuntu 16.04. AFAIK only Ubuntu 18.10 is new enough to have a kernel and glibc version that support `statx()`.

Thanks to @Vogtinator for investigating the openSUSE docker images and finding the cause.

Vogtinator:
[...]

PR #1295, opened 2020-03-30
Use latest Docker in CI

wstrm:
Force a Docker upgrade for all Linux runners.

Fixes: #1294


jubalh:
Awesome! Thanks a lot!

PR #1296, opened 2020-04-01
Fallback for nproc for systems without GNU coreutils

wstrm:
My bad :(


jubalh:
Thanks for the quick fix! :)

PR #1297, opened 2020-04-02
Fixed memory leak in ProfMucWin

pasis:
Profanity remembers last message and its id for the message correction
feature. We must free them in window destructor.


jubalh:
Ohhh right! Thanks!

PR #1306, opened 2020-04-08
Add basic MAM support

jubalh:


jubalh:
Only contains basic support.
More work is planned and needed.
Don't enable it yet if you don't know what you are doing. Only for devs.

PR #1309, opened 2020-04-10
Fix use-after-free in stanza_create_caps_from_query_element()

pasis:
The function creates a form to find such strings as software, os, etc.
It remembers the strings allocated by form_create() and use them below
in caps_create(). The issue is that the form is destroyed before and as
result the strings are freed too.

As solution, allocate own copy of strings.


jubalh:
Thanks!

PR #1316, opened 2020-04-14
iq_handlers changes

DebXWoody:
Changes for iq_handlers

### DIFF
--- a/src/xmpp/iq.c
+++ b/src/xmpp/iq.c
@@ -158,6 +158,7 @@ static int
 _iq_handler(xmpp_conn_t *const conn, xmpp_stanza_t *const stanza, void *const userdata)
 {
     log_debug("iq stanza handler fired");
+    iq_autoping_timer_cancel();
 
     char *text;
     size_t text_size;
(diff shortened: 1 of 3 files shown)


src/xmpp/iq.c
```diff
@@ -241,9 +245,6 @@ iq_handlers_init(void)
         xmpp_timed_handler_add(conn, _autoping_timed_send, millis, ctx);
     }
 
-    iq_handlers_clear();
```
pasis, line 244: There is no difference where to call these functions within `iq_handlers_init()`. I would prefer to see definitions before the rest of code:
```C
iq_handlers_init(void)
{
    xmpp_conn_t * const conn = connection_get_conn();
    xmpp_ctx_t * const ctx = connection_get_ctx();
    /* The rest of the code */
```

src/xmpp/iq.c
```diff
@@ -158,6 +158,7 @@ static int
 _iq_handler(xmpp_conn_t *const conn, xmpp_stanza_t *const stanza, void *const userdata)
 {
     log_debug("iq stanza handler fired");
+    iq_autoping_timer_cancel();
```
pasis, line 161: Why do you cancel timer on every incoming IQ stanza? When does it suppose to start again?

mdosch: Every IQ you receive proofs you're connected to the server so they are as good as ping result I guess.

pasis: I misunderstood the timer semantics. I thought it is always the same timer, but it rather starts on every ping.

I won't insist on waiting for only ping reply. But consider the following when you decide what to push to master: receiving a stanza after sending a ping doesn't mean that the stanza arrived after ping had been sent. The stanza could reside in kernel's buffers. And in this scenario you will need 1 more ping interval to detect broken connection. This is not a big deal for profanity's usecases, so it is up to you to decide how to proceed.

DebXWoody:
I will create an issue for this, we can discuss first. I need my master branch  :-D

jubalh:
@DebXWoody @pasis so what is with this PR now? :)

DebXWoody:
Close.

Issue #1315, opened 2020-04-14
Autoping response timed out - Lost connection


DebXWoody:
Starting profanity. Profanity starts and joins the MUCs via autojoin. Profanity reconnects because of autoping timeout:

17:14:19 - Autoping response timed out after 20 seconds.
17:14:19 - Lost connection.

## Environment
Profanity master on Debian Buster

DebXWoody:
I changed the implementation of `iq.c`: `void iq_autoping_check(void)`

```
     gdouble elapsed = g_timer_elapsed(autoping_time, NULL);
     unsigned long seconds_elapsed = elapsed * 1.0;
     gint timeout = prefs_get_autoping_timeout();
+    log_warning("seconds_elapsed %d", seconds_elapsed);
+
     if (timeout > 0 && seconds_elapsed >= timeout) {
         cons_show("Autoping response timed out after %u seconds.", timeout);
-        log_debug("Autoping check: timed out after %u seconds, disconnecting", timeout);
+        log_warning("Autoping check: timed out after %u seconds, disconnecting", timeout);
         iq_autoping_timer_cancel();
-        session_autoping_fail();
+        //session_autoping_fail();
     }
```

Write the seconds_elapsed into logfile (warning) and disable the `session_autoping_fail()` - do not reconnect. - Because the session is not closed.


```
...
prof: WRN: seconds_elapsed 19
prof: WRN: seconds_elapsed 20
prof: WRN: Autoping check: timed out after 20 seconds, disconnecting
prof: DBG: iq_autoping_timer_cancel
prof: DBG: Autoping: Sending ping request: YTExNTQ1ZWEtNzRhOC00NWI0LWIzNWItYTQ2M2ViMjc5ZGVmMmYzOGE0MTNkNTM2MmY5MGM1YjFmYzM2ODg0YTk1OTA5MzQ2NmUxMjU5M2VlOWM2NjNkMzI0MmUxZmI0YWY2Mw==
14/04/2020 17:51:45: conn: DBG: SENT: <iq id="YTExNTQ1ZWEtNzRhOC00NWI0LWIzNWItYTQ2M2ViMjc5ZGVmMmYzOGE0MTNkNTM2MmY5MGM1YjFmYzM2ODg0YTk1OTA5MzQ2NmUxMjU5M2VlOWM2NjNkMzI0MmUxZmI0YWY2Mw==" type="get"><ping xmlns="urn:xmpp:ping"/></iq>
[...]

PR #1318, opened 2020-04-17
OMEMO Device List only for non anonymous MUCs

DebXWoody:
Profanity request the OMEMO Device List for all members, also if the MUC is
anonymouse. If the user is Admin / Owner, the device list will be requtest.

Issue #1315


DebXWoody:
I have changed the implementation of `muc.c:muc_members_add`, that this function will check the `anonymity_type` before profanity will requests the device list.
The device list will be requested, if the MUC type is `MUC_ANONYMITY_TYPE_NONANONYMOUS`.

jubalh:
Looks good to me.

PR #1319, opened 2020-04-18
Make _inp_edited() more robust

pasis:
When I shut my system down and profanity was killed, I got messages from valgrind that `g_utf8_get_char()` and `g_unichar_isprint()` use uninitialized memory. The scenario when memory can be uninitialized is that `bytes` is not set by `wcrtomb()`. I assume that the function returns an error which is `(size_t)-1`. Handle this situation.

Also, don't pass NULL to `wcrtomb()`, because we can call different external code in pthreads and there is no guarantee that the external libraries don't call this function.

### DIFF
--- a/src/ui/inputwin.c
+++ b/src/ui/inputwin.c
@@ -324,6 +324,10 @@ _inp_write(char *line, int offset)
 static int
 _inp_edited(const wint_t ch)
 {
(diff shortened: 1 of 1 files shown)


jubalh:
Thanks!

Issue #1327, opened 2020-04-25
Status with PGP shows offline in titlebar


DebXWoody:
## Expected Behavior
Profanity shows the current status of the connection. ([TLS] / [online])

## Current Behavior
src/event/server_events.c line 156 is calling ui_init()  🤨 First, profanity goes online (titlebar looks good). I have to enter my PIN. After this, this method is called. ui_init will set OFFLINE and TLS to no.
## Possible Solution

```
git diff origin/master bughunter -- src/event/server_events.c
diff --git a/src/event/server_events.c b/src/event/server_events.c
index 057af9bf..79ac0194 100644
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -153,7 +153,6 @@ sv_ev_roster_received(void)
         win_clear(win);
         theme_init(theme);
         prefs_free_string(theme);
-        ui_init();
         ui_resize();
         ui_show_roster();
     }

```

## Environment
Profanity, version 0.8.1dev.bughunter.d20b83cb

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Disabled
PGP support: Enabled (libgpgme 1.12.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (2.7.16)
GTK icons: Enabled

jubalh:
Thanks for your work @DebXWoody !

PR #1344, opened 2020-05-23
Bugfix: Status with PGP shows offline in titlebar

DebXWoody:
There is code to redraw the ui, because the user may enter a passphase for the
private key. There was also a ui_init, which shouldn't be called, because it
will set the status to the initial state, which is 'offline' and 'no tls'.

Issue: #1327

### DIFF
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -147,13 +147,12 @@ sv_ev_roster_received(void)
         }
         free(err_str);
 
-        // Redraw the screen after entry of the PGP secret key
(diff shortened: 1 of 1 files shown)


jubalh:
Did you test whether this works with pinentry-curses?

This code was added in 6034b833be41909982b0d2bdac0a1d8499ad8e76 for https://github.com/boothj5/profanity/issues/906.

jubalh:
Merging after ACK in MUC.

Issue #1329, opened 2020-04-25
OMEMO - Request only "current" item


DebXWoody:
If there is not only one "current" item list of the OMEMO devices. profanity will request all item lists. I think we should request for "current" only.

## Possible Solution

```
diff --git a/src/xmpp/omemo.c b/src/xmpp/omemo.c
index 780cfe17..3be4dad5 100644
--- a/src/xmpp/omemo.c
+++ b/src/xmpp/omemo.c
@@ -442,7 +442,7 @@ _omemo_receive_devicelist(xmpp_stanza_t *const stanza, void *const userdata)
     }
 
     xmpp_stanza_t *item = xmpp_stanza_get_child_by_name(items, "item");
-    if (item) {
+    if (item && g_strcmp0(xmpp_stanza_get_id(item), "current") == 0 ) {
         xmpp_stanza_t *list = xmpp_stanza_get_child_by_ns(item, STANZA_NS_OMEMO);
         if (!list) {
             return 1;
@@ -462,6 +462,10 @@ _omemo_receive_devicelist(xmpp_stanza_t *const stanza, void *const userdata)
             }
         }
     }
+    if (item && g_strcmp0(xmpp_stanza_get_id(item), "current") != 0 ) {
+        log_warning("OMEMO: User %s has a non 'current' device item list: %s.", from, xmpp_stanza_get_id(item));
+    }
+
     omemo_set_device_list(from, device_list);
 
     return 1;
```

## Environment
```
Profanity, version 0.8.1dev.bughunter.d20b83cb

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Disabled
PGP support: Enabled (libgpgme 1.12.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (2.7.16)
GTK icons: Enabled
```

jubalh:
@paulfariello ?

DebXWoody:
https://xmpp.org/extensions/xep-0384.html#devices

> NOTE: as per XEP-0060 §12.20, it is RECOMMENDED for the publisher to specify an ItemID of "current" to ensure that the publication of a new item will overwrite the existing item.

jubalh:
https://github.com/profanity-im/profanity/pull/1345 got merged.

PR #1345, opened 2020-05-23
OMEMO - Request only "current" item

DebXWoody:
NOTE: as per XEP-0060 §12.20, it is RECOMMENDED for the publisher to specify an
ItemID of "current" to ensure that the publication of a new item will overwrite
the existing item.
https://xmpp.org/extensions/xep-0384.html#devices

Issue: #1329

### DIFF
--- a/src/xmpp/omemo.c
+++ b/src/xmpp/omemo.c
@@ -442,7 +442,7 @@ _omemo_receive_devicelist(xmpp_stanza_t *const stanza, void *const userdata)
     }
 
     xmpp_stanza_t *item = xmpp_stanza_get_child_by_name(items, "item");
-    if (item) {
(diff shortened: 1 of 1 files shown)


src/xmpp/omemo.c
```diff
@@ -462,6 +462,10 @@ _omemo_receive_devicelist(xmpp_stanza_t *const stanza, void *const userdata)
             }
         }
     }
+    if (item && g_strcmp0(xmpp_stanza_get_id(item), "current") != 0 ) {
```
jubalh, line 465: Please put this in the else block above so strcmp() doesnt need to run twice :-)

jubalh:
I did 24e8243a9860c15227c756c5050ea3a5d25d5134 to address the if/else.
Thanks for your work!

PR #1349, opened 2020-05-31
configure.ac: add additional check for libstrophe

### LINKED ISSUE #1334: wrong error message in ./configure
Hi!
Trying to compile profanity. I already have compiled libmesode* files in `/home/eugene/.local/lib`
and libmesode.h in `/home/eugene/.local/include`.
I run this:

```
export libmesode_LIBS="-L/home/eugene/.local/libs -lmesode"  # pay attention to typo in directory. It is "libs", but must be "lib" 
export libmesode_CFLAGS="-I/home/eugene/.local/include"
./configure --prefix=/home/eugene/.local
```
## Expected Behavior

Expecting message like "-lmesode" not found.

## Current Behavior
```
checking for libmesode... yes
checking for ncursesw... yes
checking for wget_wch support in ncursesw... no
configure: error: ncurses does not support wide characters
```

## Possible Solution
Maybe adding extra check when trying to compile this
```
int main() {
  return 0;
}
```

config.log attached. There is `/usr/bin/ld: cannot find -lmesode`

[config.log](https://github.com/profanity-im/profanity/files/4597213/config.log)

pasis:
After the library is found by pkg-config, try to build a simple program
to check the installation. The motivation of this check is that the
chance that users manually install libstrophe/libmesode is higher than
for other libs.

Fixes #1334.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -130,10 +130,21 @@ AS_IF([test "x$PTHREAD_CC" != x], [ CC="$PTHREAD_CC" ])
 
 ### Check for libmesode, fall back to libstrophe
 PKG_CHECK_MODULES([libmesode], [libmesode >= 0.9.2],
-    [LIBS="$libmesode_LIBS $LIBS" CFLAGS="$CFLAGS $libmesode_CFLAGS" AC_DEFINE([HAVE_LIBMESODE], [1], [libmesode])],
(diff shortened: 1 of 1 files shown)


jubalh:
Thanks!
Travis reports:

```
checking for libmesode... yes

checking whether libmesode works... configure: error: libmesode is broken, check config.log for details

Error 1 with command './configure $features' on line 112. Exiting.
```

pasis:
right, we need to include mesode.h for libmesode...

pasis:
Thanks for finishing it.

PR #1351, opened 2020-06-02
use always c99; fixes "universal character names" error msg

toogley:
without this, i receive this error on openbsd 6.7:

src/event/server_events.c:1477:19: error: universal character names are only valid in C++ and C99
src/event/server_events.c:1478:19: error: universal character names are only valid in C++ and C99
gmake[1]: *** [Makefile:1925: src/event/server_events.o] Error 1
gmake[1]: Leaving directory '/home/toogley/profanity'
gmake: *** [Makefile:1212: all] Error 2

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -329,7 +329,7 @@ AC_CHECK_HEADERS([ncursesw/ncurses.h], [], [])
 AC_CHECK_HEADERS([ncurses.h], [], [])
 
 ### Default parameters
-AM_CFLAGS="-Wall -Wno-deprecated-declarations"
+AM_CFLAGS="-Wall -Wno-deprecated-declarations -std=c99"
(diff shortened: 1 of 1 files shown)


wstrm:
I've added something similar here:
https://github.com/profanity-im/profanity/blob/master/ci-build.sh#L81

Maybe you could remove those lines in this commit, too?

I have used `-std=gnu99 -fexec-charset=UTF-8` instead, the last part `-fexec-charset=UTF-8` was required so that I could write non-latin characters.

I'd say it's up to @jubalh if we want `gnu99` or `c99`.

pasis:
I'd suggest to go with gnu99

toogley:
i think it would be better to use as few gnu extensions as possible, so i think c99 is better than gnu99

toogley:
@wstrm 

> Maybe you could remove those lines in this commit, too?

Will do when i have time

PR #1360, opened 2020-06-10
Make legacy auth optional

pasis:
Some systems don't provide recent libstrophe releases. When older
version of libstrophe is detected, don't build legacy auth support.

To simplify this patch, report about unsupported legacy auth and
keep commands option as is.

/!\ This patch adds `ui/ui.h`. Is it ok?


jubalh:
Thanks!

> /!\ This patch adds ui/ui.h. Is it ok?

Since this will only be present in 0.9.X and removed for 0.10.0 I think that's okay.

Issue #1366, opened 2020-06-16
Handle all types of messages


pasis:
Profanity doesn't handle all message types and silently drops unknown ones.

According to https://tools.ietf.org/html/rfc6121#section-5.2.2, clients should handle all the types and additionally process messages without the type attribute as messages with `normal` type.

Current unknown types are `headline` and `normal`. Also, Profanity drops messages without the attribute.

DebXWoody:
What should happen, when we get a headline message?

`<message type="headline"><body>Text</body></message>`

Display in the console window?

jubalh:
I started with cleanup for message.c.
The following two commits take care of the allowed message types now:
46a00317ee85be02fc9174315db24297079d5c4d
3640900cf64f00591526444a80d9c412ba302a0a

Will still cleanup more, once I read the corresponding XEPs.

> Also, Profanity drops messages without the attribute.

What do you mean by that?

jubalh:
3640900cf64f00591526444a80d9c412ba302a0a takes care of 'headline'.
Feel free to reopen (will close to have a better overview).

PR #1369, opened 2020-06-21
Handle all types of messages

DebXWoody:
Issue: #1366

### DIFF
--- a/src/xmpp/message.c
+++ b/src/xmpp/message.c
@@ -86,7 +86,8 @@ static void _handle_captcha(xmpp_stanza_t *const stanza);
 static void _handle_receipt_received(xmpp_stanza_t *const stanza);
 static void _handle_chat(xmpp_stanza_t *const stanza, gboolean is_mam);
 static gboolean _handle_mam(xmpp_stanza_t *const stanza);
-
+static void _handle_headline(xmpp_stanza_t *const stanza);
+static void _handle_normal(xmpp_stanza_t *const stanza);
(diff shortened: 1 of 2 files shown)


jubalh:
Are you sure all the return 1 are good? I'm not sure whether some cases actually depend on the next.

DebXWoody:
Reset and removed the return 1 code. Created new (first) commit.
I need to check some details.

src/xmpp/message.c
```diff
@@ -166,7 +171,11 @@ _message_handler(xmpp_conn_t *const conn, xmpp_stanza_t *const stanza, void *con
         }
     }
 
-    _handle_chat(stanza, FALSE);
+    if (g_strcmp0(type, STANZA_TYPE_CHAT) == 0)  {
```
jubalh, line 174: Not sure whether all other messages have type chat.
Maybe it's better to use 
```
if (g_strcmp0(type, "normal") == 0 || type == NULL  )  {
    _handle_normal(stanza);
} else {
  _handle_chat(stanza, FALSE);
}
?

jubalh:
I took the _handle_headline() part and merged it in 3640900cf64f00591526444a80d9c412ba302a0a.
8d8ef675cd6e8d8ee756d247551d255edcea2a87 should then fix a memleak.

I omitted the handle_normal as I'm not sure that's right.

But I'll revisit that during message.c cleanup anyways. (Once I read all XEPs and understand what will be type normal and what will be type chat. It seems in practice its sometimes different that what is said in the XEP).

Issue #1361, opened 2020-06-12
[request merge] Define external commands to open and save URLs according to file types and protocols


pmaziere:
The following changes since commit 74ff38f0bda6e365bdedf29fbd53eedfa520ed74:

  Start new cycle (2020-06-11 10:07:51 +0200)

are available in the Git repository at:

  https://framagit.org/peetah/profanity.git/ revampUrlopen

for you to fetch changes up to 72487041253ae959117e056fce83c4d0017fba34:

  Display default value for /url associated commands (2020-06-12 09:35:55 +0200)

----------------------------------------------------------------
Pierre Mazière (6):
      add aesgcm to urls grabber
      refactor cmd_urlopen
      Add string and string list preferences with option
      Replace /urlopen with /url and adapt /executable
      Add /url autocompletion
      Display default value for /url associated commands

 src/command/cmd_ac.c     |  26 ++++++++++++++++++----
 src/command/cmd_defs.c   |  49 ++++++++++++++++++++++++-----------------
 src/command/cmd_funcs.c  | 190 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--------------
 src/command/cmd_funcs.h  |   3 ++-
 src/config/preferences.c | 105 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++----
 src/config/preferences.h |   5 +++++
 src/ui/console.c         |  14 +++++++-----
 src/ui/window_list.c     |   2 +-
 8 files changed, 343 insertions(+), 51 deletions(-)

jubalh:
Thanks! Will try to take a look soon :)

jubalh:
@pmaziere merged your branch in a test branch ontop of current master.
`/url open https://github.com/profanity-im/profanity` segfaulted.

jubalh:
`/executable` needs to have conisistent indenting:

```
19:30:20 - '/avatar' command (/executable avatar)                                   : xdg-open                                                                
19:30:20 - Default '/url open command' (/executable urlopen)                                 : xdg-open %u                                                    
19:30:20 - Default '/url save' command (/executable urlsave)                                 : curl -o %p %u   
```

pmaziere:
There is a default value in case `/executable` is not configured yet.
The segfault was caused by the URL missing a file extension.
This is now fixed, as well as the `/executable` output indenting.
[...]

Issue #1262, opened 2020-02-03
Send and retrieve encrypted files when using OMEMO


wstrm:
## Expected Behavior

The `/sendfile` command should send the file encrypted as AES-256 in GCM while using OMEMO, according to XEP-xxxx: OMEMO Media Sharing.

Profanity should also be able to decrypt files and save them to the file system as is already supported when retrieving unencrypted files.

## Current Behavior

When sending files with `/sendfile` the files are sent unencrypted and my friends gets upset that I'm not using OMEMO Media Sharing.

When receiving an encrypted files an `aesgcm://` URI is displayed. 

## Possible Solution

N/A

## Context

I am member of a large MUC that uses OMEMO, a lot of memes are shared an I want to save them seamlessly to my file system.

## Environment
N/A

wstrm:
> /close as duplicate #1070

I got told to create this issue

jubalh:
Yep, that's right. This issue is for tracking progress of that feature and so it can be assigned easily.
I added the link to it at the appropriate place in https://github.com/profanity-im/profanity/issues/1262.
Which is just the general feature tracker.

moppman:
There was some discussion the other day in one of the MUCs about profanity sending unencrypted files in an encrypted chat without warning users about it. That's why I made #1270 until we have actual file encryption :)

mdosch:
As I already said in the conversations MUC I'd rather have a way to 
override that.

jubalh:
#1270 got merged. `/omemo sendfile on` is now needed to allow sending unencrypted files using `/sendfile` while in an OMEMO session.

PR #1374, opened 2020-06-25
Rework /url and /executable for filetypes

jubalh:
This PR includes the (rebased, organized and squashed) commits from @pmaziere from his repo https://framagit.org/peetah/profanity/ revampUrlopen branch.
It is about https://github.com/profanity-im/profanity/issues/1361 with the purpose of  defining external commands to open and save URLs according to file types and protocols.

One benefit of this is that it will allow users to download and decrypt OMEMO file extensions using [omut](https://github.com/wstrm/omemo-utils).

### DIFF
--- a/src/command/cmd_ac.c
+++ b/src/command/cmd_ac.c
@@ -123,7 +123,7 @@ static char* _avatar_autocomplete(ProfWin *window, const char *const input, gboo
 static char* _correction_autocomplete(ProfWin *window, const char *const input, gboolean previous);
 static char* _correct_autocomplete(ProfWin *window, const char *const input, gboolean previous);
 static char* _software_autocomplete(ProfWin *window, const char *const input, gboolean previous);
-static char* _urlopen_autocomplete(ProfWin *window, const char *const input, gboolean previous);
+static char* _url_autocomplete(ProfWin *window, const char *const input, gboolean previous);
 static char* _executable_autocomplete(ProfWin *window, const char *const input, gboolean previous);
 
 static char* _script_autocomplete_func(const char *const prefix, gboolean previous, void *context);
@@ -259,6 +259,7 @@ static Autocomplete logging_group_ac;
 static Autocomplete color_ac;
 static Autocomplete correction_ac;
 static Autocomplete avatar_ac;
+static Autocomplete url_ac;
 static Autocomplete executable_ac;
 
 void
@@ -1007,9 +1008,14 @@ cmd_ac_init(void)
     autocomplete_add(avatar_ac, "get");
     autocomplete_add(avatar_ac, "open");
 
+    url_ac = autocomplete_new();
+    autocomplete_add(url_ac, "open");
+    autocomplete_add(url_ac, "save");
+
     executable_ac = autocomplete_new();
     autocomplete_add(executable_ac, "avatar");
     autocomplete_add(executable_ac, "urlopen");
+    autocomplete_add(executable_ac, "urlsave");
 }
 
 void
@@ -1328,6 +1334,7 @@ cmd_ac_reset(ProfWin *window)
     autocomplete_reset(color_ac);
     autocomplete_reset(correction_ac);
     autocomplete_reset(avatar_ac);
+    autocomplete_reset(url_ac);
     autocomplete_reset(executable_ac);
 
     autocomplete_reset(script_ac);
@@ -1489,6 +1496,7 @@ cmd_ac_uninit(void)
     autocomplete_free(color_ac);
     autocomplete_free(correction_ac);
     autocomplete_free(avatar_ac);
+    autocomplete_free(url_ac);
     autocomplete_free(executable_ac);
 }
 
@@ -1749,7 +1757,7 @@ _cmd_ac_complete_params(ProfWin *window, const char *const input, gboolean previ
     g_hash_table_insert(ac_funcs, "/correction",    _correction_autocomplete);
     g_hash_table_insert(ac_funcs, "/correct",       _correct_autocomplete);
     g_hash_table_insert(ac_funcs, "/software",      _software_autocomplete);
-    g_hash_table_insert(ac_funcs, "/urlopen",       _urlopen_autocomplete);
+    g_hash_table_insert(ac_funcs, "/url",           _url_autocomplete);
     g_hash_table_insert(ac_funcs, "/executable",    _executable_autocomplete);
 
     int len = strlen(input);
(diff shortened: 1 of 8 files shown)


jubalh:
@DebXWoody can you also review this please?

jubalh:
@pmaziere when I do:

```
/executable urlsave http curl -o %p %u
```

I get:
```
15:45:53 - '/avatar' command (/executable avatar)                                   : xdg-open
15:45:53 - Default '/url open' command (/executable urlopen)                        : xdg-open %u
15:45:53 - Default '/url save' command (/executable urlsave)                        : curlas -o %p %u
```

jubalh:
Honestly I'm not sure whether the benefit is worth the added complexity.

Especially once https://github.com/profanity-im/profanity/issues/1262 encrypted OMEMO filetransfer is implemented.

I think Profanity should not try to solve too many tasks.
Opening an URL should just be opened in a browser. Of course it would be nice to open movies with mpv or pictures with feh. But I'm not sure this is actually worth it.
It makes sense to make a distinction between open and save, but even there I think open would be enough and people should use their terminal to actually save something.

Maybe an alternative would just be to have `/url open` and `/url save`. So we don't have to check for filetype and "extension" of the URL since actually we cannot know what a URL will bring us. http://somedomain/test/ could be anything. Even with .html it could actually just be an image.

@pasis, @wstrm what do you think about this?

wstrm:
I would expect `/url open` to simply open the URL using `xdg-open`, or an alternative opener - that would handle the parsing of the protocol (I think this is in agreement w/ what you wrote @jubalh).

Using `xdg-open` one could create a file called `aesgcm.desktop` with something like (not tested, just a guess):
```
[Desktop Entry]
Name=OMEMO Media Sharing
Exec=omut -d "%u" | display
Type=Application
MimeType=x-scheme-handler/aesgcm;
```

wstrm:
Thinking about this a bit more, and the protocol scheme matching that this PR brings makes sense if we want to use this command for the OMEMO Media Sharing PR too. As we would have to match the `aesgcm://` scheme internally to decrypt and download the content (of course one could use something like omut to decrypt, too, but it would be weird if Profanity supported encryption but not decryption).

So I'd say it would be nice if Profanity supports downloading content for some supported schemes like `https://` and `aesgcm://` (would use the same download logic as both are `https://` in the end). libcurl could even let us support most protocols[1]. The actual content can then be opened by `xdg-open` (or any other handler) using `/url open` or just saved to disk with `/url save`.

It would be _cool_ to also match custom URL scheme's and allow for a custom executable that handles the URL (this PR brings hardcoded support for `aesgcm://` and `https://`, right?).

So maybe this is crazy/stupid, but let's keep the discussion going and maybe we come up with something good.

---

Here's an idea (mostly like this PR):
 * `/url save` - Matches scheme and either uses internal method or external if configured.
 * `/url open` - Simply invokes an external "opener" like `xdg-open`, Profanity doesn't do anything that matches file type.

The configured `/url save` executable takes `%u` and `%p`, and `/url open` takes `%u` (like in this PR).

If `/url open` is issued on a `ftp://` URL Profanity will save the file in a temporary location using the configured executable for `/url save` (by passing it with the `%p` format parameter), and then issue the configured opener (i.e. `xdg-open`) on it by passing `file:///tmp/profanity-asdf123` for the `%u` format parameter.

Example of `/url open`:

```
/url exec save "ftp" "wget %u %p"
/url exec open "xdg-open %u"

...

/url open ftp:///foo/bar
```

This would happen behind the scenes for the last command:

```
wget ftp:///foo/bar /tmp/profanity-asdf123 # hmm, would this happen in the background, or block?
xdg-open file:///tmp/profanity-asdf123
```

1. https://curl.haxx.se/libcurl/c/url2file.html
---
And just a note that [cmd_url_open](https://github.com/profanity-im/profanity/pull/1374/files#diff-7dee2bc3cffe25b1e10a999045047fc2R8942) could have used the somewhat new libcurl URL API to parse the scheme (and support any scheme and not only `aesgcm` and `https`): [https://ec.haxx.se/libcurl/libcurl-url](https://ec.haxx.se/libcurl/libcurl-url). I use this API in my PR in the [format_alt_url](https://github.com/profanity-im/profanity/pull/1375/files#diff-5d44d56d6e61e6888a5dec3e4c26b1d7R131) function (this API is in the Debian stable libcurl package).

jubalh:
Still a lot to do until we can merge this.

@pmaziere how does the user set the default handler? The one that is used if we don't know the file extension?

pmaziere:
> @pmaziere when I do:
> 
> ```
> /executable urlsave http curl -o %p %u
> ```
> 
> I get:
> 
> ```
> 15:45:53 - '/avatar' command (/executable avatar)                                   : xdg-open
> 15:45:53 - Default '/url open' command (/executable urlopen)                        : xdg-open %u
> 15:45:53 - Default '/url save' command (/executable urlsave)                        : curlas -o %p %u
> ```
> 
> instead of it just chaning the value.

Because you need to quote the command, otherwise the number of arguments passed to executable exceeds the max_args defined in cmd_defs.c, which triggers cons_executable_setting

pmaziere:
> Still a lot to do until we can merge this.
> 
> @pmaziere how does the user set the default handler? The one that is used if we don't know the file extension?

we could use * for this specific case:

`/executable urlopen * "xdg-open %u"`

> Why is require_save needed?

because some viewers are not able to open a remote file

pmaziere:
@wstrm 

>So I'd say it would be nice if Profanity supports downloading content for some supported schemes like https:// and aesgcm:// (would use the same download logic as both are https:// in the end). libcurl could even let us support most protocols[1]. The actual content can then be opened by xdg-open (or any other handler) using /url open or just saved to disk with /url save.

If I'm not mistaking, using libcurl means an additional dependency, which I tend to avoid unless really necessary, as well as  less freedom about the downloading choices, unless we are ready to provide preferences for all possible options that libcurl allows.
I find it more KISS  to call an external executable to do this job, which can then be tuned by users to the extent of their requirements (proxies, UA, etc. …)

> It would be cool to also match custom URL scheme's and allow for a custom executable that handles the URL (this PR brings hardcoded support for aesgcm:// and https://, right?).

I was not sure, security wise, that matching all `://`' containing strings and extracting the scheme was a good idea, therefore I kept it hardcoded for now. I was hoping for this discussion to drive the way to the right implementation for this feature.

> If /url open is issued on a ftp:// URL Profanity will save the file in a temporary location using the configured executable for /url save (by passing it with the %p format parameter), and then issue the configured opener (i.e. xdg-open) on it by passing file:///tmp/profanity-asdf123 for the %u format parameter.

That is exactly what is done for `aesgcm://` URLs in this PR.
And this is also what you can configure for any file type viewer which can not open remote files.

jubalh:
> I like NULLing a pointer once it has been freed since it allows me to easily see this pointer status when I read the code.
Moreover when this pointer is inadvertently reused, the fact that it is NULL makes the error obvious

I'm aware of that practise. But I think it clutters the code too much. And so far it is done nowhere in Profanity. So lets stay with consistent style.

jubalh:
> If I'm not mistaking, using libcurl means an additional dependency, which I tend to avoid unless really necessary

libcurl already is a dependency.

jubalh:
Doing:

```
/executable urlsave * "test %u"
/save
```

profrc has:
```
[executables]
```

So its empty for some reason.

jubalh:
Probably some limitation of `.ini` files.

`/executable urlsave html "test %u"` results in `url.save.cmd[html]=test %u`
`/executable urlsave * "test %u"` results in nothing.

jubalh:
If you run `/executable` it doesnt print the information for all schemes. Only one. Not for http/https and whatnot you can configure.

I really think this PR adds too much complexity which is not yet thought through.
I assume many things are not necessary, we don't need to do and support everything inside profanity.

We just need a `/url save` and `/url open` that does always do the same. The user can only configure one command that is run.
Not specific to file type. Internally we then check whether its OMEMO encrypted and in that case we decrypt the file when saving/before opening.

All the rest seems overkill and adds a lot of things that really need more thought.

I'll merge it now anyway.
So that @wstrm can base his work for OMEMO file transfer on it https://github.com/profanity-im/profanity/issues/1262.

PR #1375, opened 2020-06-27
Send and retrieve encrypted files when using OMEMO

wstrm:
Fixes: #1262


Neustradamus:
@wstrm: Have you progressed on it?

wstrm:
> @wstrm: Have you progressed on it?

No, not lately. Hopefully I'll get back to it soon.

src/config/preferences.c
```diff
@@ -1893,6 +1865,7 @@ _get_group(preference_t pref)
         return PREF_GROUP_LOGGING;
     case PREF_AVATAR_CMD:
     case PREF_URL_OPEN_CMD:
+        return PREF_GROUP_EXECUTABLES;
```
jubalh, line 1868: would fall through anyway :)

wstrm: Doh

wstrm:
@jubalh Show no mercy

src/command/cmd_defs.c
```diff
@@ -2274,55 +2274,52 @@ static struct cmd_t command_defs[] = {
...
-              { "policy always", "Set the global OMEMO policy to always, an error will be displayed if an OMEMO session cannot be initiated upon starting a conversation." },
-              { "sendfile on|off", "Allow /sendfile to send unencrypted files while in an OMEMO session." },
-              { "clear_device_list", "Clear your own device list on server side. Each client will reannounce itself when connected back." })
-      CMD_EXAMPLES(
-              "/omemo gen",
-              "/omemo start odin@valhalla.edda",
-              "/omemo trust c4f9c875-144d7a3b-0c4a05b6-ca3be51a-a037f329-0bd3ae62-07f99719-55559d2a",
-              "/omemo untrust loki@valhalla.edda c4f9c875-144d7a3b-0c4a05b6-ca3be51a-a037f329-0bd3ae62-07f99719-55559d2a",
-              "/omemo char *")
+        parse_args, 1, 3, NULL,
```
jubalh, line 2277: indentation got changed here. I know.. its a bit of a problem because we don't use default here.. :-/

src/command/cmd_funcs.h
```diff
@@ -128,6 +128,7 @@ gboolean cmd_status_get(ProfWin* window, const char* const command, gchar** args
 gboolean cmd_status_set(ProfWin* window, const char* const command, gchar** args);
 gboolean cmd_sub(ProfWin* window, const char* const command, gchar** args);
 gboolean cmd_theme(ProfWin* window, const char* const command, gchar** args);
+gboolean cmd_tiny(ProfWin* window, const char* const command, gchar** args);
```
jubalh, line 131: Seems this is a mistake to introduce old tiny?

src/omemo/omemo.h
```diff
@@ -33,13 +33,18 @@
...
 #include <glib.h>
+#include <gcrypt.h>
 
 #include "ui/ui.h"
 #include "config/account.h"
 
 #define OMEMO_ERR_UNSUPPORTED_CRYPTO -10000
 #define OMEMO_ERR_GCRYPT             -20000
 
+#define OMEMO_AESGCM_NONCE_LENGTH 12
```
jubalh, line 44: should this be set to AES128_GCM_IV_LENGTH  ? In case they always should be the same?

src/ui/console.c
```diff
@@ -2073,10 +2073,10 @@ cons_executable_setting(void)
     g_free(avatar);
 
     //TODO: there needs to be a way to get all the "locales"/schemes so we can
-    //display the defualt openers for all filetypes
-    gchar** urlopen = prefs_get_string_list_with_option(PREF_URL_OPEN_CMD, "");
+    //display the default openers for all filetypes
+    char* urlopen = prefs_get_string_with_option(PREF_URL_OPEN_CMD, "");
     cons_show("Default '/url open' command (/executable urlopen)                        : %s", urlopen[1]);
```
jubalh, line 2078: Won't that segfault with urlopen[1] ?

jubalh (changes requested):
See comments in code plus:
* Where do you use `omemo_free()`
* You added two empty files: src/tools/aesgcm_upload.c src/tools/aesgcm_upload.h

src/omemo/omemo.h
wstrm: Yeah it's just that it is mentioned separatedly for the OMEMO media sharing document, so I thought it could be viewed as "separate" from the standard -- but I can set it to AES128_GCM_IV_LENGTH

src/command/cmd_funcs.h
wstrm: Haha yes! Must be a bad rebase, I gues!

src/ui/console.c
wstrm: Could you elaborate?

src/command/cmd_defs.c
wstrm: I shouldn't do something here, right? :)

src/ui/console.c
wstrm: Aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaah

jubalh:
I like a lot what you did to http upload!

Some more reviewing:

`/executable` prints:

```
17:31:01 - Default '/avatar open' command (/executable avatar)                      : xdg-open
17:31:01 - Default '/url open' command (/executable urlopen)                        : false;xdg-open %u;
17:31:01 - Default '/url save' command (/executable urlsave)                        : (null)
```

What happens when a user sets `/executable urlsave aesgcm "omut -d -o %p %u"` which is still mentioned in `/help executable` and later wants to go back to default?
`/executable urlsave (<protocol>|DEF) <cmd>` is also mentioned in `/help executable` but setting `/executable urlsave DEF (null)` won't work.

So first I think `/help executable` needs an update. And second we need a way to reset to default.

jubalh (approved):
Looking good! Except that you don't expand ~. So `/url save https://profanity-im.github.io/faq.html ~/something` will make trouble.

jubalh:
I opened https://github.com/profanity-im/profanity/pull/1447 for the last comment.

jubalh:
@wstrm Thanks a lot for your work! This will be a nice christmas present for users!

Neustradamus:
@wstrm: Thanks for your job in Profanity!

I am not sure if https://github.com/wstrm/omemo-utils is compatible with all OMEMO 0.3.0 clients:
- "AES128_GCM_IV_LENGTH"
- "AES128_GCM_KEY_LENGTH"
- "AES128_GCM_TAG_LENGTH"

PR #1379, opened 2020-07-01
Cleanup files.c

jubalh:


jubalh:
@DebXWoody @wstrm in case you have time for a quick review.

PR #1381, opened 2020-07-02
Fix messages with receipt request not displayed

wstrm:
Messages from Conversations contains:
  ```<request xmlns='urn:xmpp:receipts'/>```

And would not be displayed in Profanity as it never reached
_handle_chat(..).


jubalh:
super, thanks! :)

PR #1383, opened 2020-07-03
Cleanup carbons/chat handling in message.c

jubalh:
We basically had a lot of duplicate code in _handle_chat() and _handle_carbons().
I try to improve the situation.

There is still more to be done. But I would like to get a review of it as is, just to make sure that I'm keeping the old behaviour.
Added some TODOs for more stuff that I need to check later on.

### DIFF
--- a/src/xmpp/message.c
+++ b/src/xmpp/message.c
@@ -84,8 +84,9 @@ static void _handle_muc_private_message(xmpp_stanza_t *const stanza);
 static void _handle_conference(xmpp_stanza_t *const stanza);
 static void _handle_captcha(xmpp_stanza_t *const stanza);
 static void _handle_receipt_received(xmpp_stanza_t *const stanza);
-static void _handle_chat(xmpp_stanza_t *const stanza, gboolean is_mam);
+static void _handle_chat(xmpp_stanza_t *const stanza, gboolean is_mam, gboolean is_carbon);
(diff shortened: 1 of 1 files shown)


jubalh:
@DebXWoody @wstrm maybe you can review? :)

src/xmpp/message.c
```diff
@@ -1307,17 +1234,34 @@ _handle_chat(xmpp_stanza_t *const stanza, gboolean is_mam)
...
     ProfMessage *message = message_init();
-    message->from_jid = jid;
     message->is_mam = is_mam;
+    message->from_jid = jid;
+    if (to_jid) {
+        message->to_jid = to_jid;
+    } else {
+        if (is_carbon) {
+            // happens when receive a carbon of a self sent message
+            // really? maybe some servers do this, but it's not required.
```
wstrm, line 1260: I know this comment existed before... It's talking to itself?

Maybe rewrite it and elaborate?

DebXWoody:
I just have had a quick look and didn't check the details, but I like the idea on the handle_carbon changes.

src/xmpp/message.c
jubalh: First line was present earlier, later then I added the second line because I have doubts about it. I still need to check it in more detail. So far it contains all I know :)

Issue #1384, opened 2020-07-03
Messages are not shown in ChatSecure


DebXWoody:
Messages are not shown in ChatSecure

PR #1385, opened 2020-07-03
OMEMO: Messages are not shown in ChatSecure 

DebXWoody:
In 0.9.x we fixed an issue, because OMEMO devices should be defined in "item"
with id "current". This should work, but it won't work if there is no "current".
If there is no "current" we will just use the first item.

Issue #1384

### DIFF
--- a/src/omemo/omemo.c
+++ b/src/omemo/omemo.c
@@ -98,7 +98,7 @@ struct omemo_context_t {
     GHashTable *pre_key_store;
     GHashTable *signed_pre_key_store;
     identity_key_store_t identity_key_store;
-    GHashTable *device_ids;
+//    GHashTable *device_ids;
     GString *identity_filename;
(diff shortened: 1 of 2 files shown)


DebXWoody:
Confirmed via PM. The fix + adding trust of fingerprint solved the ChatSecure issue.

src/omemo/omemo.c
```diff
@@ -98,7 +98,7 @@ struct omemo_context_t {
     GHashTable *pre_key_store;
     GHashTable *signed_pre_key_store;
     identity_key_store_t identity_key_store;
-    GHashTable *device_ids;
+//    GHashTable *device_ids;
```
wstrm, line 101: Remove this line? :)

jubalh:
Merged in bf170d6c9168f987ab4431e733bf1d120af54df3

Issue #1353, opened 2020-06-04
Rework logging


jubalh:
```
20:34:37 - DebXWoody: jubalh, We should change the logging of profanity. A normal user expects no problems and I think the user wish to see / check thinks on 
           the client. That means, the default level should be warning. Users which would like to see what is happening behind the screen, can set the log    
           level to INFO. We should print the most important things with info. Martin would be able to see what the client is doing. DEBUG should be used by  
           developers. Per default there won't be much in the logs and the user can switch to INFO. We can add some more informations into info level.
```

jubalh:
@DebXWoody do you consider this done?

PR #1386, opened 2020-07-03
Rework logging

DebXWoody:
Changed default logging from INFO to WARN.

Issue: #1353

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -232,8 +232,8 @@ log_level_from_string(char *log_level)
         return PROF_LEVEL_WARN;
     } else if (strcmp(log_level, "ERROR") == 0) {
         return PROF_LEVEL_ERROR;
-    } else { // default to info
-        return PROF_LEVEL_INFO;
+    } else { // default logging is warn
+        return PROF_LEVEL_WARN;
     }
 }
 
--- a/src/main.c
+++ b/src/main.c
@@ -78,7 +78,7 @@ main(int argc, char **argv)
     {
         { "version", 'v', 0, G_OPTION_ARG_NONE, &version, "Show version information", NULL },
         { "account", 'a', 0, G_OPTION_ARG_STRING, &account_name, "Auto connect to an account on startup" },
-        { "log", 'l', 0, G_OPTION_ARG_STRING, &log, "Set logging levels, DEBUG, INFO (default), WARN, ERROR", "LEVEL" },
+        { "log", 'l', 0, G_OPTION_ARG_STRING, &log, "Set logging levels, DEBUG, INFO, WARN (default), ERROR", "LEVEL" },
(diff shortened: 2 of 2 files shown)


src/log.c
```diff
@@ -232,8 +232,8 @@ log_level_from_string(char *log_level)
         return PROF_LEVEL_WARN;
     } else if (strcmp(log_level, "ERROR") == 0) {
         return PROF_LEVEL_ERROR;
-    } else { // default to info
-        return PROF_LEVEL_INFO;
+    } else { // default logging is warn
+        return PROF_LEVEL_WARN;
```
jubalh, line 236: shouldnt this one stay info?

src/main.c
```diff
@@ -78,7 +78,7 @@ main(int argc, char **argv)
     {
         { "version", 'v', 0, G_OPTION_ARG_NONE, &version, "Show version information", NULL },
         { "account", 'a', 0, G_OPTION_ARG_STRING, &account_name, "Auto connect to an account on startup" },
-        { "log", 'l', 0, G_OPTION_ARG_STRING, &log, "Set logging levels, DEBUG, INFO (default), WARN, ERROR", "LEVEL" },
+        { "log", 'l', 0, G_OPTION_ARG_STRING, &log, "Set logging levels, DEBUG, INFO, WARN (default), ERROR", "LEVEL" },
```
jubalh, line 81: why reordering this?

src/main.c
```diff
@@ -177,7 +177,8 @@ main(int argc, char **argv)
         return 0;
     }
 
-    prof_run(log ? log : "INFO", account_name, config_file, log_file, theme_name);
+    /* Default logging WARN */
+    prof_run(log ? log : "WARN", account_name, config_file, log_file, theme_name);
```
jubalh, line 181: so we only log warnings to the logfile, seems fine.

src/main.c
DebXWoody: Didn't reordering it, just moved the default from INFO to WARN.

src/log.c
DebXWoody: When I understand well, it's just called from main during init of profanity.

main.c
prof_run(log ? log : "INFO", account_name, config_file, log_file, theme_name);

prof_run calling: _init(log_level, config_file, log_file, theme_name)

profanity.c:_init(char *log_level, char *config_file, char *log_file, char *theme_name)
log_level_t prof_log_level = log_level_from_string(log_level);

if we would use NULL or "WHATEVER" it will still use the WARN level as default.

src/main.c
DebXWoody: Yes, and in the second round of clean-up we can add some helpful information with "info" for users and developer, which can be enabled very easy.

jubalh:
Thanks!

PR #1389, opened 2020-07-05
Manpage in adoc format

DebXWoody:
Write the manpages in adoc (AsciiDoc) format and create the manpage and html based on the adoc-file.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,0 +1,13 @@
+#
+# Manpage to build docs.
+#
+
+ADOC_FILE = profanity.1.adoc
+
+all: manpage
+
+manpage:
+	asciidoc -b docbook $(ADOC_FILE)
+	asciidoc -b html $(ADOC_FILE)
+	a2x --doctype manpage --format manpage $(ADOC_FILE)
+
--- a/docs/profanity.1.adoc
+++ b/docs/profanity.1.adoc
@@ -1,0 +1,78 @@
+PROFANITY(1)
+============
+
+NAME
+----
+profanity - a simple console based XMPP chat client.
+
+SYNOPSIS
+--------
+'profanity' [-vh] [-l 'level'] [-a 'account']
+
+DESCRIPTION
+-----------
+Profanity  is  a simple lightweight console based XMPP chat client. Its emphasis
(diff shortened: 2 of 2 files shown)


jubalh:
What is your goal with this? And what will the html be used for?

DebXWoody:
Improve the documentation. When I install a new application, I first check the manpage. The manpage can be improved with helpful. 

* Explain the files which are stored in .local/share/profanity (see example FILES at man 1 gpg)
* Explain the supported XEPs - maybe create a own man page `man profanity-xep`

If some domains are blocked, the website is not working. We should find a way to have a one place to hold the documentation and generated this for manpage and website and help within the application.

Instead of generating HTML:
https://github.com/profanity-im/profanity/blob/master/src/command/cmd_defs.c#L2772 
We can generate adoc format and those filed can used to generate html / manpage / DocBook / Latex / ....

mdosch:
BTW, the one time I wrote a manpage I settled for ronn which is pretty 
comfortable as you can write it in markdown. You can see an example 
there: https://salsa.debian.org/mdosch/go-sendxmpp/-/tree/master/man

DebXWoody:
md is also fine for me.

jubalh:
I think this is a fine idea.

It will be good to later have a `profanity-reference` which basically will be https://profanity-im.github.io/guide/latest/reference.html in manpage format. And it will be good if `profanity -docgen` will create that manpage and will also handle creating the corrsponding HTML.

Same might be for https://profanity-im.github.io/xeps.html.

But for the current usual manpage I don't see why you create a HTML from it like you do in this PR.

jubalh:
So I think the following needs to happen:
* create issue with detailed steps
* change this PR to remove the generation of html for usual manpage
* mention the newly created issue in the commit in this PR

Later:
* rewrite docgen to use adoc
* create both html for our website, and manpage for the user `profanity-reference`
* mention new manpage under 'see also' in regular manpage

jubalh:
https://github.com/profanity-im/profanity/pull/1445 was merged. Now for each profanity command a man page is generated.
Our existing manpage probably won't change much so for now I don't see much benefit in writing them in asciidoc. But since there is one small downside of having a new dependency on asciidoc I see this as a reason to not merge this for now.

Issue #1331, opened 2020-04-25
Feature request - XEP-0373: OpenPGP for XMPP (OX)


DebXWoody:
Feature request for profanity: Implementation of XEP-0373: OpenPGP for XMPP (OX)

Information:
* https://xmpp.org/extensions/xep-0373.html
* https://xmpp.org/extensions/xep-0374.html
* https://xmpp.org/extensions/xep-0380.html

DebXWoody:
Idea
Implementation like it is for pgp in the current version of profanity. Instead of `/pgp` command, we can use the command `/ox`. There are some difference between pgp (0027) and openpgp (0373).

* The message can be <signcrypt/>, <sign/> or <crypt/>, not encrypted, only
* signed presences is not part of XEP-0380 and not XEP-0373
* Lookup for a key can be done via XMPP URI (see https://xmpp.org/extensions/xep-0373.html#openpgp-user-ids)
* OX don't use ascii-armor: https://xmpp.org/extensions/xep-0373.html#openpgp-ascii-armor

DebXWoody:
There is a small example here: https://codeberg.org/Anoxinon_e.V./xmppc/src/branch/master/src/mode/openpgp.c#L69

DebXWoody:
I started to implement a `/ox` command for profanity:
https://github.com/DebXWoody/profanity/commits/feature/ox-devel

The command `/ox keys` shows all keys of the gnupg's keyring. Same like `/pgp keys`.

The command `/ox contacts` looks into the keyring and search for uid with a XMPP-URI.

```
OpenPGP keys:
56ECD7B83A173A355C7C18400BF9E946C3DD112D - xmpp:stefan@domain4.tld (not in roster)
2909AF6BF3D6DF3BA1B1805C539F89DC6083AA33 - xmpp:anna@domain3.tld
6D740FE2B1D55DCD74CAAD95AC1A1629095EEDAE - xmpp:alice@domain2.tld (not in roster)
A8431A0170B3EBB564CE294D0C1CE873ED588C2B - xmpp:alice@domain1.tld (not in roster)
7FA1EB8644BAC07E7F18E7C9F121E6A6F3A0C7A5 - xmpp:doctor.snuggles@domain.tld (not in roster)
```

DebXWoody:
Note: https://github.com/profanity-im/profanity/blob/master/src/event/client_events.c#L135

DebXWoody:
Current status for **basic** implementation of:

[...]

PR #1390, opened 2020-07-05
OX - announce public key

DebXWoody:
Issue: #1331


jubalh:
Nice!

Is this supposed two commits? The description of the commits sounds very much alike? Can we change the description of the first one so it reflects more detailed what it does? Or should we squash them into one?

jubalh:
Rebased, squashed and merged as 5a179572535805022580b264a99df60a47cb935d.

PR #1393, opened 2020-07-05
Plain chat messages not working

DebXWoody:
Plain chat messages not working for non-carbon + no OTR support.

### DIFF
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -73,6 +73,7 @@
 #include "ui/ui.h"
 
 static void _clean_incoming_message(ProfMessage *message);
+static void _sv_ev_incoming_plain(ProfChatWin *chatwin, gboolean new_win, ProfMessage *message, gboolean logit);
 
 void
 sv_ev_login_account_success(char *account_name, gboolean secured)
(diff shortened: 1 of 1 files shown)


DebXWoody:
On master we did some clean-up. The problem is at https://github.com/profanity-im/profanity/blob/0.9.patch/src/event/server_events.c#L625 (0.9.0). The implementation looks like:

- HAVE_LIBOTR is set - _sv_ev_incoming_otr
- HAVE_LIBOTR is not set - _sv_ev_incoming_plain

I think the `_sv_ev_incoming_otr` can handle otr and plain, because I didn't find a `_sv_ev_incoming_plain` if `HAVE_LIBOTR` is set.

On master for 0.10.0 the implementation is much better:
https://github.com/profanity-im/profanity/blob/master/src/event/server_events.c#L623

But, we just call `_sv_ev_incoming_otr` independent of `HAVE_LIBOTR`.
Unfortunately, `_sv_ev_incoming_otr` is doing nothing if `HAVE_LIBOTR` is not set:
https://github.com/profanity-im/profanity/blob/master/src/event/server_events.c#L538

I did some more clean-up at sv_ev_incoming_message and changed the implementation of `_sv_ev_incoming_otr`.

```
static void
_sv_ev_incoming_otr(ProfChatWin *chatwin, gboolean new_win, ProfMessage *message)
{
#ifdef HAVE_LIBOTR
  // OTR or plain
#else
        plain
#endif
}
```

The caller do not take care of `HAVE_LIBOTR`, call `_sv_ev_incoming_plain` if you are sure it's a plain message or call `_sv_ev_incoming_otr`. `_sv_ev_incoming_otr` can be used for otr / plain or for plain only.

jubalh:
Merged in 2a50bb6ad7d34848c9acb0d87469bb92e9e6e133

PR #1402, opened 2020-07-10
Ubuntu Latest (20.04 currently) in CI

wstrm:
Feel free to close if you do not want Ubuntu in the pipeline.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ matrix:
     - os: linux
       env: BUILD_FLAVOR=debian
     - os: linux
+      env: BUILD_FLAVOR=ubuntu
+    - os: linux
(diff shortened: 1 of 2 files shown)


jubalh:
I think i'll be very similar to our Debian one. Plus we also use libmesode git here. We could use libstrophe git or libstrophe package. (Though package might not make sense in some scenarios, but might help in others).

Not sure if it is worth the longer waiting time for travis to finish.

Some distribution that has more differences like Alpine would be more interesting to test.

wstrm:
OK, all of them are run in parallel so it doesn't add anything on the execution time (except if maybe Ubuntu decides to take a longer time), currently it's just the macOS build that takes extra time.

PR #1407, opened 2020-07-20
Change to .tar.zst archives that latest Arch devtools uses

wstrm:
This is required because of:
https://www.archlinux.org/news/now-using-zstandard-instead-of-xz-for-package-compression/


jubalh:
Great. Thanks!

Issue #1414, opened 2020-07-27
Memory leak


pasis:
Master (3e7776a9faa164bceaa1a1b6d8612fd68d075676)

```
==13112== 102 (48 direct, 54 indirect) bytes in 1 blocks are definitely lost in loss record 723 of 965
==13112==    at 0x4A340FF: malloc (vg_replace_malloc.c:309)
==13112==    by 0x13CC35: jid_create (jid.c:70)
==13112==    by 0x1469A8: _handle_chat (message.c:1245)
==13112==    by 0x147A4B: _message_handler (message.c:235)
==13112==    by 0x147A4B: _message_handler (message.c:152)
==13112==    by 0x6C6A0DC: handler_fire_stanza (in /usr/lib64/libstrophe.so.0.0.0)
==13112==    by 0x6C66BBA: _handle_stream_stanza (in /usr/lib64/libstrophe.so.0.0.0)
==13112==    by 0x6C74C5E: _end_element (in /usr/lib64/libstrophe.so.0.0.0)
==13112==    by 0x6FD9972: doContent (in /usr/lib64/libexpat.so.1.6.11)
==13112==    by 0x6FDA63B: contentProcessor (in /usr/lib64/libexpat.so.1.6.11)
==13112==    by 0x6FDCBBB: XML_ParseBuffer (in /usr/lib64/libexpat.so.1.6.11)
==13112==    by 0x6C69A13: xmpp_run_once (in /usr/lib64/libstrophe.so.0.0.0)
==13112==    by 0x140ABE: connection_check_events (connection.c:121)
```

jubalh:
I'm aware of that one. Saw it already some time ago but didn't see why it happens at that time. PR welcome.

pasis:
I will check the code when I have spare time

pasis:
Looks like the root cause is in `_handle_chat()`:
```C
    const gchar* to = xmpp_stanza_get_to(stanza);
    if (to) {
        to_jid = jid_create(to);
    }
...
    if (to_jid) {
        message->to_jid = to_jid;
...
    const gchar* to_text = xmpp_stanza_get_to(stanza);
    if (to_text) {
        message->to_jid = jid_create(to_text);
    }
```

pasis:
Closed via #1416.

PR #1416, opened 2020-08-06
Fix memory leak in _handle_chat()

pasis:
jid_create() for attribute "to" was called twice and the 1st object was lost.


pasis:
Fixes #1414.

jubalh:
Thanks!

PR #1424, opened 2020-09-04
NetBSD support

niacat:
The following changes were required:

- Added NetBSD as a platform that does not have a separate `libdl` to `configure.ac`.
- Changed a `==` condition in `configure.ac` to use `=` instead. These are identical, it's just that `=` is in POSIX and `==` isn't. NetBSD's `/bin/sh` only supports the POSIX syntax.
- Recognize NetBSD `libcurses` as a valid curses implementation.
  AFAIK, NetBSD is now the only operating system that does not primarily use ncurses. Its own curses library has support for wide characters and similar features -- there's also a [port of it to Linux](https://github.com/sabotage-linux/netbsd-curses).
- NetBSD curses is less permissive when handed null pointers than ncurses, apparently. I had to add a few checks to avoid null pointer deref.

Notes for building profanity on NetBSD:

- To get it to pick up `python-config`, either create a non-versioned symlink in `/usr/local/bin` or install `pkg_alternatives` - the executable is versioned in NetBSD, but this isn't really a problem for local builds or packaging.
- To test these changes, it's helpful to not have the ncurses or ncursesw packages installed. Alternatively, I moved the `*.pc` files out the way, and configured with: `MAKE=gmake ac_cv_header_ncursesw_ncurses_h=no ac_cv_lib_ncursesw_main=no ac_cv_lib_ncurses_main=no ./configure`
- As with other BSDs, GNU make is spelled `gmake`.
- Installing `autoconf-archive` probably helps

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -22,6 +22,7 @@ AC_CANONICAL_HOST
 PLATFORM="unknown"
 AS_CASE([$host_os],
     [freebsd*], [PLATFORM="freebsd"],
+    [netbsd*], [PLATFORM="netbsd"],
     [openbsd*], [PLATFORM="openbsd"],
     [darwin*], [PLATFORM="osx"],
     [cygwin], [PLATFORM="cygwin"],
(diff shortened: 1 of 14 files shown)


jubalh (approved):

jubalh:
Thanks for this !

PR #1426, opened 2020-09-30
Cleanup/newlibs

jubalh:


jubalh:
@ibotty do you see why Fedora fails detecting libstrophe 0.10.0 in this PR?

jubalh:
This Pr is about newlibs only. Fixing travis for Debian/Fedora should be done in a separate one.

Issue #1442, opened 2020-11-27
Disable profanity logging (not chatlogs) alltogeter or choosing where to write them


mrkiko:
Thus not having ~/.local/share/profanity/logs/profanity.log around.

Sorry if this is possible and I am missing the way to do it.
Thanks!!

jubalh:
You can use `profanity -f logfile1` to use `~/.local/share/profanity/logs/logfile1.log` instead. of default filename.

mrkiko:
Thanks!! I was looking for a way to use e.g.: /tmp/logfile1 ... ...

jubalh:
I don't remember the reason why I made `-f` use a filename in the log dir and not let the user define the whole filename. I think there was one.. I'll look into it.
Then one could also do something like `profanity -f /dev/null`.

wstrm:
@jubalh did you find any reason? Doing stuff like `profanity -f /dev/null` that you noted would be nice :)

jubalh:
@mrkiko please test with the new -f behaviour. We will notify people in the changelog of the behaviour change.

PR #1455, opened 2020-12-11
Use whole path as logfile when defined via -f

jubalh:
`profanity -f my` created ~/.local/share/profanity/logs/my.log`.

It would be nicer if one could define the actual path, so one can choose
another directory or even use /dev/null.

Fixes https://github.com/profanity-im/profanity/issues/1442

### DIFF
--- a/src/config/files.c
+++ b/src/config/files.c
@@ -116,20 +116,26 @@ char*
 files_get_log_file(const char* const log_file)
 {
     gchar* xdg_data = _files_get_xdg_data_home();
-    GString* logfile = g_string_new(xdg_data);
+    GString* logfile;
 
     if (log_file) {
-        g_string_append(logfile, "/profanity/logs/");
-        g_string_append(logfile, log_file);
+        gchar *log_path = g_path_get_dirname(log_file);
+        if (!mkdir_recursive(log_path)) {
+            log_error("Error while creating directory %s", log_path);
+        }
+        g_free(log_path);
+
+        logfile = g_string_new(log_file);
     } else {
+        logfile = g_string_new(xdg_data);
         g_string_append(logfile, "/profanity/logs/profanity");
-    }
 
-    if (!prefs_get_boolean(PREF_LOG_SHARED)) {
-        g_string_append_printf(logfile, "%d", getpid());
-    }
+        if (!prefs_get_boolean(PREF_LOG_SHARED)) {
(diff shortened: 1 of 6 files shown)


jubalh:
This isn't finished yet. Need some opinion here.

The problem, and probably also the reason why I chose to make -f use only the filename in a dir was to keep things simple (for me) :D 

We have `log_reinit()` which is run once we change `/log shared on|off`. It closes the current log file and uses a new one (toggling with PID attached to filename or not).

So either we need to rewrite `log_reinit()` which probably means another variable that will hold the information to build the proper filename and will use space all the time. Or we remove `log_reinit()` altogether and tell users they need to restart the client before `/log shared on|off` takes effect.

I'm for the latter since its much easier. Otherwise we need  to take into account all kind of scenarios like.

@wstrm whats your opinion?

wstrm:
> Or we remove log_reinit() altogether and tell users they need to restart the client before /log shared on|off takes effect.

As long as we tell the user (in the console) once they've used `/log shared on|off` that a restart is required I'm fine with it :)

jubalh:
@wstrm next question: does it make sense to do logfile rotation if the user provided a logfile with -f? I guess not. So it seems we need a way to detect whether a logfile is supplied anyways. And in this case don't rotate the logs?

src/config/files.c
```diff
@@ -116,20 +116,26 @@ char*
...
 {
     gchar* xdg_data = _files_get_xdg_data_home();
-    GString* logfile = g_string_new(xdg_data);
+    GString* logfile;
 
     if (log_file) {
-        g_string_append(logfile, "/profanity/logs/");
-        g_string_append(logfile, log_file);
+        gchar *log_path = g_path_get_dirname(log_file);
+        if (!mkdir_recursive(log_path)) {
```
wstrm, line 123: Hm, maybe I should move the download directory creation to this file instead then?

wstrm (approved):

src/config/files.c
jubalh: Right, that would be possible.

mrkiko:
Thanks! Works great!!

As per the logfile rotation - I think it may make sense. removing the logfile with rm would not help infact in e.g.: freeing space, or management.
But restarting the client isn't a problem for me.
thanks a lot for your great work and helping me out. this is a gift.

PR #1456, opened 2020-12-11
Fix color for win_print_outgoing()

jubalh:
Discovered by @aba-hollerer.

Mistake was introduced in b6b7dd5ad497a71e250b8b3cef0bb987314b141d
probably due to a wrong copy/paste.

First wasn't reproducible because I had `/receipts request on` and thus
win_print_outgoing_receipts() is used which has the correct
THEME_TEXT_ME.

### DIFF
--- a/src/ui/window.c
+++ b/src/ui/window.c
@@ -1213,7 +1213,6 @@ win_print_outgoing(ProfWin* window, const char* show_char, const char* const id,
     if (replace_id) {
         _win_correct(window, message, id, replace_id, myjid);
     } else {
-        //TODO my jid
         _win_printf(window, show_char, 0, timestamp, 0, THEME_TEXT_ME, "me", myjid, id, "%s", message);
     }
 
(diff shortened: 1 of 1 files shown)


wstrm (approved):

src/ui/window.c
```diff
@@ -1214,7 +1214,7 @@ win_print_outgoing(ProfWin* window, const char* show_char, const char* const id,
         _win_correct(window, message, id, replace_id, myjid);
     } else {
         //TODO my jid
```
wstrm, line 1216: What is this about?

jubalh: I'm not entirely sure anymore. At one point I rewrote window functions to have more infos inside the function like the whole jid and not just a part of it given as *char. I'm not sure whether this TODO is up to date or just forgot to remove it. In https://github.com/profanity-im/profanity/commit/f7fe87dd4e7e14112fbcb0ae8794d90ffee64f2f it was added and in outgoing_muc() too but there it got removed later. So I guess I just forgot to remove it it.

jubalh: I did a new commit which explains and removes this TODO :)

Issue #1359, opened 2020-06-10
Can't use eval_password


TheBrainScrambler:
I currently have an user, let's call it <user>. The password of this XMPP user is stored in a KeepassXC database, db.kdbx. I want profanity to get this password through the `eval_password` setting from the account.

If I do `keepassxc-cli show /path/to/db.kdbx /path/to/password -sa Password` I am granted with a prompt "Enter your password : " so I type it, and it prints the right password. Everything is ok here.

Now I enter profanity. I do `/account set <user> eval_password "keepassxc-cli show /path/to/db.kdbx /path/to/password -sa Password"`. Ok no problem. But when I `/connect`, profanity just freezes.

I thought that profanity would spawn a shell where the prompt appears, then you type the password, and then it is good. But no, actually profanity just freezes because the call to keepassxc-cli expects my password, which I have no way to enter.

## Expected Behavior

Profanity should give us a way to interact with prompts spawned by `eval_password`

## Current Behavior

Profanity just freezes ...

## Possible Solution

As stated above, maybe spawn some temporary shell in the entry field. The user wouldn't see any difference, but he would be typing in that shell.

## Steps to Reproduce (for bugs)

1. Have some KeepassXC file. (should be ending with .kdbx)
2. For a given user, type `/account set <user> eval_password "keepassxc-cli show /path/to/db.kdbx /path/to/password -sa Password`
3. `/connect`
4. Enjoy the freeze !

## Context

"I want profanity to get this password (the XMPP account's one) through the `eval_password` setting from the account."

## Environment
* Give us the version and build information output generated by `profanity -v`
```
Profanity, version 0.8.1

Build information:
XMPP library: libmesode
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.13.1)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.8.3)
GTK icons: Enabled
```

* Operating System/Distribution
Arch Linux
* glib version
Uh ... `pacman -Qs glib`
local/glib2 2.64.2-1
local/glibc 2.31-3

One of the two above I suppose

jubalh:
> I thought that profanity would spawn a shell where the prompt appears, then you type the password, and then it is good. But no, actually profanity just freezes because the call to keepassxc-cli expects my password, which I have no way to enter.

Spawn a shell? Where?

[...]

PR #1459, opened 2020-12-11
Restore old behavior when executing eval_password command

wstrm:


wstrm:
@jubalh I've done some testing and the Pinentry example in #1359 works.

However, the original command (without Pinentry) in that issue still doesn't work. I tried to revert the `account_eval_password` function to the old variant using `popen` -- but that didn't work either (I get an empty terminal that doesn't react on any input).

So for now this PR brings back the old functionality, but I think we want to support also commands not using e.g. Pinentry?

wstrm:
I'll mark this as ready and I can hopefully look into why xkeepass-cli doesn't work in the future.

This PR makes it work with Pinentry using TTY again.

jubalh:
Thanks!

PR #1461, opened 2020-12-11
Set input window size to max window size

### LINKED ISSUE #1458: Input win is shorter than max window size
Input win is one cell shorter than max window size. Therefore the last cell is not themed correctly if background color is set.

kaffeekanne:
Input window size was one cell off, therefore terminal themes with background color set were rendered incorrectly, fixes #1458


jubalh (approved):

jubalh:
Thank you!

PR #1465, opened 2020-12-21
Print logs on error in CI

wstrm:
Because we trap the error signal and run a error handler, logs were not
printed. I moved the printing of test-suite.log to the error handler and
also added so that config.log is printed on failure.


jubalh (approved):

jubalh:
Thanks!

Issue #1301, opened 2020-04-06
Inline display of images


jubalh:
We could use @hpjansson s [chafa](https://github.com/hpjansson/chafa/)/libchafa to display images inline.

## Expected Behavior
There should be a dedicated setting for this. For example `/inlineimage on|off`.
When `on` and a user sends a URL ending in a known image format (we will have to search through each incoming message).
We download the image to a temporary location (or a general media location) and use chafa to display the picture inside Profanity.

Similar this could be done for avatars, a new `/avatar view contact` could be added.

## Current Behavior
The URL is shown.

## Possible Solution
Chafa has also sixel support. I'm not sure whether we can detect whether a terminal supports this or should make it easy and just let the user have a setting for this.
So in case sixel support is present/set we can use sixel to display the image, otherwise we use one of chafas tricks.

See https://hpjansson.org/chafa/gallery/ what chafa looks like on the terminal.

It might be helpful to look at what vifm and ranger do:
https://github.com/vifm/vifm/commit/7e21fa69a5e78cdbea84b3bb6cf7cf749e99ce8f
https://github.com/ranger/ranger/issues/1434

Example what it looks like in ranger:
<img width="1440" alt="ranger" src="https://user-images.githubusercontent.com/1658215/78599448-704d0000-7851-11ea-8d13-eb185454f991.png">

hpjansson:
I'm available to help if needed. Note that Chafa can be made to generate Unicode, ASCII or sixel output using the same API. There isn't a lot of extra work to get sixel output that way, even the dimensions are still specified in character cells.

Also, if you need extra API (for instance, ways to access canvas contents), feel free to let me know. I'm already planning to add more accessors/drawing functions, but I haven't planned out the details yet, so input is very welcome.

jubalh:
@mdosch @wstrm 
Anybody knows how Conversations decides whether to display an image inline?

Displaying each image inline? Sounds like you can be spammed easily. So should we do it like `/url open <tab>`? That you can tab through the last links with image ending and upon hitting enter just that image will be displayed?

How do we detect whether an URL actually is just an image? We would need to download it first I assume since just the file ending won't be a guarantee.

mdosch:
Afaik conversations automatically downloads from contacts in your roster if the file size is below a configurable threshold.

jubalh:
See https://github.com/profanity-im/profanity/pull/1466#issuecomment-4180856793

PR #1466, opened 2021-01-11
Inline images

jubalh:
PR for https://github.com/profanity-im/profanity/issues/1301

So far this is just a test. Displaying an image `~/test.png` when Profanitys "about" is printed.
So far in ASCII art.

libchafa can also return colourful images using escape codes. But not sure how to use them inside ncurses yet.

I assume later we want a setting that either displays all images (URLs which we assume to be images based on ending) or we want it only explicitly with somethine like `/image` / `/imagepreview` / `/inlineimage` where one can tab through the latest URLs with image endings.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -69,6 +69,8 @@ AC_ARG_WITH([themes],
     [AS_HELP_STRING([--with-themes[[=PATH]]], [install themes (default yes)])])
 AC_ARG_ENABLE([icons-and-clipboard],
     [AS_HELP_STRING([--enable-icons-and-clipboard], [enable GTK tray icons and clipboard paste support])])
+AC_ARG_ENABLE([inline-images],
+    [AS_HELP_STRING([--enable-inline-images], [enable inline display of images])])
 
 ### plugins
 
@@ -191,9 +193,6 @@ PKG_CHECK_MODULES([glib], [glib-2.0 >= 2.62.0], [],
     [AC_MSG_ERROR([glib 2.62.0 or higher is required for profanity])])
 PKG_CHECK_MODULES([gio], [gio-2.0], [],
     [AC_MSG_ERROR([libgio-2.0 from glib-2.0 is required for profanity])])
-#TODO: only under cond
-PKG_CHECK_MODULES([pixbuf], [gdk-pixbuf-2.0], [],
-    [AC_MSG_ERROR([gdk-pixbuf-2.0 from glib-2.0 is required for profanity])])
 
 ### Check for other profanity dependencies
 AC_SEARCH_LIBS([fmod], [m], [],
@@ -205,10 +204,6 @@ PKG_CHECK_MODULES([curl], [libcurl], [],
 
 PKG_CHECK_MODULES([SQLITE], [sqlite3 >= 3.22.0], [],
     [AC_MSG_ERROR([sqlite3 3.22.0 or higher is required for profanity])])
-
-#TODO: only under cond
-PKG_CHECK_MODULES([chafa], [chafa >= 1.4.1], [],
-    [AC_MSG_ERROR([chafa needed])])
 
 AS_IF([test "x$enable_icons_and_clipboard" != xno],
     [PKG_CHECK_MODULES([GTK], [gtk+-3.0 >= 3.24.0],
@@ -328,6 +323,16 @@ if test "x$enable_omemo" != xno; then
                [AC_MSG_NOTICE([gcrypt >= 1.7.0 not found, OMEMO support not enabled])])])
 
    AM_COND_IF([BUILD_OMEMO], [AC_DEFINE([HAVE_OMEMO], [1], [Have OMEMO])])
+fi
+
+if test "x$enable_inline_images" = xyes; then
+    PKG_CHECK_MODULES([pixbuf], [gdk-pixbuf-2.0], [],
+        [AC_MSG_ERROR([gdk-pixbuf-2.0 from glib-2.0 is required for inline image support])])
+
+    PKG_CHECK_MODULES([chafa], [chafa >= 1.4.1], [],
+        [AC_MSG_ERROR([chafa >= 1.4.1 needed for inline image support])])
+
+   AC_DEFINE([HAVE_INLINE_IMAGE], [1], [Have inline image])
 fi
 
 AS_IF([test "x$with_themes" = xno],
--- a/configure.ac
+++ b/configure.ac
@@ -191,6 +191,9 @@ PKG_CHECK_MODULES([glib], [glib-2.0 >= 2.62.0], [],
     [AC_MSG_ERROR([glib 2.62.0 or higher is required for profanity])])
 PKG_CHECK_MODULES([gio], [gio-2.0], [],
     [AC_MSG_ERROR([libgio-2.0 from glib-2.0 is required for profanity])])
+#TODO: only under cond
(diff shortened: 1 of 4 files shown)


jubalh:
chafa has a test branch with added functionality for us:
https://github.com/hpjansson/chafa/tree/wip/cell-access
example: https://github.com/hpjansson/chafa/blob/wip/cell-access/tests/ncurses.c

Need to build this chafa version and see whether this helps progressing on this task.

binex-dsk:
i'm pretty sure using `viu` would work too, no? It displays stuff with colored blocks, or uses Kitty's image protocol if available, which i guess would be similar to what you're looking for, plus Kitty users could get stuff in high quality

jubalh:
`viu` is an cli image viewer. There are plenty of those. It outputs things on the terminal we can't use that to display a picture inside our chat windows. `viuer` is the library that `viu` uses. It is written in Rust. Profanity is written in C. So we we settled on libchafa.

hpjansson:
I merged the wip/cell-access branch into master some time ago. I don't expect it to change very much between now and release, so it should be safe to start using.

I'm also planning Kitty and iTerm2 support, but I'm not sure how you'd get ncurses to cooperate...

jubalh:
@binex-dsk a little similar to the QR code task. Do you think that would be something for you? ;)

binex-dsk:
> @binex-dsk a little similar to the QR code task. Do you think that would be something for you? ;)

I could possibly do basic block art in the same vain as the QR code. Not sure if colorful images would be my forte, but I'll look into it.

Also: are shell color codes agnostic to operating systems/shell standards?

jubalh:
> I could possibly do basic block art in the same vain as the QR code. Not sure if colorful images would be my forte, but I'll look into it.

In this PR I tried to use the library "chafa" actually it should do most of the work. We just need set the correct settings (what it will return) and then print this properly in the window.

Like `chafa_symbol_map_add_by_tags (symbol_map, CHAFA_SYMBOL_TAG_ASCII);` and there are various thing to set there. I also think @hpjansson added new options since I tried this last time.

> Also: are shell color codes agnostic to operating systems/shell standards?

I'm not entirely sure. But I don't think so.

> Also, I'm currently on a band trip without access to a computer. I'll see if I want to tackle this after next Sunday/Monday.

Enjoy!

binex-dsk:
> In this PR I tried to use the library "chafa" actually it should do most of the work. We just need set the correct settings (what it will return) and then print this properly in the window.
> 
> Like `chafa_symbol_map_add_by_tags (symbol_map, CHAFA_SYMBOL_TAG_ASCII);` and there are various thing to set there. I also think @hpjansson added new options since I tried this last time.
> 

Okay, I see. In that case I'll mess around with the settings and see what I can do. What is it you're looking for that this PR does not yet do (other than deciding on what to do for displaying them, of course)

And should I test this on the cell-access thing mentioned? Or on master?

hpjansson:
If your package manager has Chafa 1.8.0 or later, the cell access API will be there, so you can build against the devel package instead of master. In case it saves time.

binex-dsk:
If no one else takes this up within a few weeks I'll go ahead and do it... right now I've got a boatload of stuff to do: summer classes, driving practice, and soccer. I might experiment with the ncurses examples at some point though :)

mrusme:
Having inline images rendered via Chafa would really be awesome! I have only scratched the surface here, but this also requires integration with OMEMO in case of `aesgcm://` URLs, correct?

jubalh:
@binex-dsk how are your driving classes? Do you have time for this feature now maybe?

jubalh:
That's great news!

binex-dsk:
Quick update, I have been doing some basic libchafa testing (on its own, not with profanity) and although I haven't made that much progress, I AM getting there. I would be farther but, I had to go to urgent care, which (unfortunately) is more important than programming...

jubalh:
Thanks for the update!
Take your time and take care!

binex-dsk:
I've been messing with this for a while and so far I've not been able to figure out several things.

Here's the point I've reached:

image

I have yet to figure out:
- True color; term info IS properly detected but truecolor mode is not yet working. In my previous tests I never managed to get this to work either...
- Kitty pixel support; Also never managed to figure this out in previous tests.
- Wrong aspect ratio; Never a problem in my previous tests.

You can also notice that the pixels don't look right. It's a QR code so it should be all squares.

Commit is here:

https://github.com/binex-dsk/profanity/commit/8e5aa52c72fbb99b8ca4b50b8a7278521ad56b83

binex-dsk:
Looked into the ncurses example a bit more and I'm confused.

@hpjansson Would you mind explaining the `canvas_to_ncurses` function in the ncurses example? I believe this is the function we need for truecolor but I haven't managed to decipher it.

Also, how would we go about actual Kitty support? It seems like it would be really hard to make it work with kitty, though I don't know the internals of the Kitty protocol so maybe I'm missing something.

binex-dsk:
> Note that you need to link with the correct build of ncurses (e.g. ending in w6 for wide characters)

Aha, that's why the function didn't end up working after I plugged it in!

An interesting thing I'll need to look into is getting it to work with the standard `win_println`-style of printing--that is, with the time and indented as other messages are. Although it might be better to end up displaying it without the indent/time, and instead, image messages have text content "Image" or "Media Message" or similar, and then in the next line the image is displayed.

I'm unfamiliar with the internal workings of the profanity window thingy so I have no clue if it'll properly scroll if printed with `mvadd_wch`. Although a brief look at the print functions does suggest it will.

I suppose I'll look into all of this, perhaps more, tomorrow. Depends how much free time I've got. Which might not be a lot, because God knows if I'll end up needing to go to the ER again.

> I think this would be impossible, or at least not worth the effort, to support in an ncurses program unless there's a non-scrolling preview rectangle where you can just bypass ncurses and overwrite the contents using your own escape codes.

This is what I suspected. I have an idea on how that could work, but it would be clunky, awkward, perhaps user-unfriendly, and not to mention a HUGE pain to program.

Thanks for the assistance @hpjansson; this makes a lot more sense now!

binex-dsk:
After a test using `canvas_to_ncurses`, `mvadd_ch` does *not* properly print the image to the window... I think what we should do is add a way to print a wchar array, and in the `canvas_to_ncurses` function, we construct a wchar array and use that to print. But the former is beyond my area of expertise...

binex-dsk:
@jubalh perhaps you could help with the wchar printing? I have no idea how to implement this, as I'm not familiar with the internal print functions.

jubalh:
@binex-dsk ok I will try to look at it! :)

binex-dsk:
@jubalh how has the wchar printing progress coming? Perfectly okay if you haven't done anything yet

jubalh:
@binex-dsk sorry I haven't yet. I'm having some health issues. I will try to look at it when all is okay again :)

hpjansson:
Just from a quick look, it looks like `buffer.c` would be a good place to start. You could add code there to queue messages with images in them. Then `window.c` has a redraw function and print functions -- functionality could be added there to paint the contents of images from the buffer object.

hpjansson:
Also -- @jubalh, take care and get well soon!

binex-dsk:
> I'm having some health issues

Lmao same

Hope you get well! :)

jubalh:
By now I think inline images are not a great idea. Probably https://github.com/profanity-im/profanity/issues/1966 makes more sense.

PR #1481, opened 2021-02-14
Bugfix: Titlebar show name without room name

StefanKropp:
Error log:
GLib-CRITICAL - g_string_insert_len: assertion 'len == 0 || val != NULL' failed

Check if a room name exists, before adding the name into the title bar.


jubalh (approved):

jubalh:
Good catch! Thanks!

PR #1482, opened 2021-02-14
Feature: Date and Time

StefanKropp:


mwuttke97:
As a nice side effect, this also makes building on Debian 10 easier. This patch deprecates the Debian patch `use-glib-2.58.3.patch`. See <https://packages.debian.org/buster-backports/profanity>.

StefanKropp:
> As a nice side effect, this also makes building on Debian 10 easier.
yes, no need to change sources.

Otherwise, you need to apply a patch like this:
https://wiki.debian.org/de/Profanity

PR #1483, opened 2021-02-14
Compare two JIDs ( full and bare )

StefanKropp:


jubalh:
It's better to add new functions only when they are actually used. If I see it right so far you don't use these two helper functions yet. Usually to write clean code people suggest not to add helper functions in advanced because one "will need it later" since they could be forgotten and not used at all. Let's rather add them when they are actually needed.

jubalh:
```
ack --cc strcmp |grep jid | wc -l
53
```

So we use things like `if (g_strcmp0(confwin->roomjid, roomjid) == 0) {` quite often. And actually I don't see much value for having a separate function to do it. I think it's quite readable already.

PR #1484, opened 2021-02-15
Calm OpenBSD syslog

kaffeekanne:
On every write to `chatlog.db` syslog throws
```
profanity: vfprintf %s NULL in "INSERT INTO `ChatLogs` (`from_jid`,
`from_resource`, `to_jid`, `to_resource`, `message`, `timestamp`,
`stanza_id`, `archive_id`, `replace_id`, `type`, `encryption`) SELECT
'%s', '%s', '%s', '%s', '%s', '%s', '%s', '%s', '%s', '%s', '%s' WHERE
NOT EXISTS (SELECT 1 FROM `ChatLogs` WHERE `archive_id` = '%s')"
```
in `/var/log/messages`. Checking for `NULL` except in the fields the DB
is expected to throw errors for, satisfies OpenBSDs security measures

### DIFF
--- a/src/database.c
+++ b/src/database.c
@@ -334,14 +334,14 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
                  from_jid->resourcepart ? from_jid->resourcepart : "",
                  to_jid->barejid,
                  to_jid->resourcepart ? to_jid->resourcepart : "",
-                 escaped_message,
-                 date_fmt,
+                 escaped_message ? escaped_message : "",
+                 date_fmt ? date_fmt : "",
                  message->id ? message->id : "",
                  message->stanzaid ? message->stanzaid : "",
                  message->replace_id ? message->replace_id : "",
-                 type,
-                 enc,
-                 message->stanzaid)
+                 type ? type : "",
+                 enc ? enc : "",
+                 message->stanzaid ? message->stanzaid : "")
         == -1) {
         log_error("log_database_add(): SQL query. could not allocate memory");
         return;


src/database.c
```diff
@@ -330,18 +330,18 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
     char* escaped_message = str_replace(message->plain, "'", "''");
 
     if (asprintf(&query, "INSERT INTO `ChatLogs` (`from_jid`, `from_resource`, `to_jid`, `to_resource`, `message`, `timestamp`, `stanza_id`, `archive_id`, `replace_id`, `type`, `encryption`) SELECT '%s', '%s', '%s', '%s', '%s', '%s', '%s', '%s', '%s', '%s', '%s' WHERE NOT EXISTS (SELECT 1 FROM `ChatLogs` WHERE `archive_id` = '%s')",
-                 from_jid->barejid,
+                 from_jid->barejid ? from_jid->barejid : "",
```
jubalh, line 333: It's not correct to use this in some cases.
As you can see in https://github.com/profanity-im/profanity/blob/master/src/database.c#L118 some fields are expected to never be NULL. If one of those values is NULL we want sqlitedb to throw an error when we try to insert such a query. I think for this reason "" is not good but needs actually be NULL in those cases. Please change your code accordingly.

jubalh (changes requested):

src/database.c
jubalh: Also once you do these changes please also change the commit log. The subject line should not exceed 80 chars if possible. In the body you can then be more verbose. There I would also mention the actual file path of the syslog file and what is written in there (just 1 or 2 example lines) before this change.

jubalh:
You should have used `git commit --amend` for the second changes.

Now you need to do `git rebase -i HEAD~2` or something to squash the two commits together and then keep in mind: https://github.com/profanity-im/profanity/pull/1484#discussion_r576036893

jubalh (approved):

jubalh:
Thanks!

PR #1485, opened 2021-02-15
Add basic functionality to launch external editor

StefanKropp:


src/command/cmd_funcs.c
```diff
@@ -9296,33 +9296,74 @@ cmd_mam(ProfWin* window, const char* const command, gchar** args)
...
+        cons_show("Editor: Error during file creation");
+        return TRUE;
+    }
+    g_object_unref(fos);
+
+    // Fork / exec
     pid_t pid = fork();
     if( pid == 0 ) {
-        int x = execl("/usr/bin/vim", "/usr/bin/vim", filename, (char *) NULL);
+        int x = execl("/usr/bin/sensible-editor", "/usr/bin/sensible-editor", g_file_get_path(file), (char *) NULL);
```
jubalh: sensible-terminal is a Debian thing. We probably want to set an executable with `/executables` and default to `$EDITOR`.

src/command/cmd_funcs.c
```diff
@@ -9296,33 +9296,74 @@ cmd_mam(ProfWin* window, const char* const command, gchar** args)
 gboolean
 cmd_editor(ProfWin* window, const char* const command, gchar** args)
 {
-    const char* filename = "/tmp/profanity.txt";
+    /* Build temp file name */
+    GString * tempfile = g_string_new(g_get_tmp_dir());
```
jubalh, line 9300: Instead of creating a Gstring and appending several times. I would rather use `g_strdup_printf()`. It's just one call instead of plenty of lines :) You can format the string right there. And no need for `const char* filename = tempfile->str;` then too since it already returns a gchar.

src/command/cmd_funcs.c
```diff
@@ -9296,33 +9296,74 @@ cmd_mam(ProfWin* window, const char* const command, gchar** args)
...
+    if (uuid) {
+        xmpp_free(ctx, uuid);
+    }
+    g_string_append(tempfile, ".txt");
+
+    // tempfile should be something like
+    // /tmp/profanity-f2f271dd-98c8-4118-8d47-3bd49c8e2e63.txt
+    const char* filename = tempfile->str;
+
+    // Check if file exists and create file
```
jubalh, line 9318: Is it really appropriate to exit if the file exists? I'm not sure whether actually we would just open it as is or even rewrite? Is there a reason why we need several such files even? Maybe we should just always write to `/tmp/profanity-externaledit` or something? If a person chooses to use the editor command often it could mean he creates plenty of files in /tmp. And if this is on a server or somewhere where profanity runs for weeks it could mean that it uses lots of space. If the system is configured to only clean /tmp upon restart that will mean wasting plenty of space for a long time on a server for example.

jubalh: Just now saw that you delete the file at the end. So the space thing doesn't make sense. But makes me wonder even more why we need a random filename :)

jubalh (changes requested):

src/command/cmd_funcs.c
jubalh: If you don't have time to do this one just leave it and I'll do it myself after merging.

src/command/cmd_funcs.c
jubalh: PM from @StefanKropp "we need random in case multiple instanced of profanity are running"

kaffeekanne: Too bad, continuing typing would be nice. But let that be a feature request later.

src/command/cmd_funcs.c
```diff
@@ -9286,3 +9292,91 @@ cmd_mam(ProfWin* window, const char* const command, gchar** args)
...
+               &creation_error);
+    if ( creation_error ) {
+        cons_show_error("Editor: Error during file creation");
+        return TRUE;
+    }
+    g_object_unref(fos);
+
+    char* editor = prefs_get_string(PREF_COMPOSE_EDITOR);
+    if (!g_file_test(editor, G_FILE_TEST_EXISTS)) {
+        cons_show_error("Editor: editor %s not exists", editor);
```
kaffeekanne, line 9338: Should be `"Editor: editor %s does not exists"`.

src/config/preferences.c
```diff
@@ -2323,6 +2326,8 @@ _get_default_string(preference_t pref)
         return "xdg-open";
     case PREF_URL_OPEN_CMD:
         return "xdg-open %u";
+    case PREF_COMPOSE_EDITOR:
+        return "/usr/bin/vim";
```
kaffeekanne, line 2330: May be we should consider using `/usr/bin/vi` as default. That should exist on almost any *nix machine. For example OpenBSD installs to `/usr/local/bin` but there is a basic `vi` in `/usr/bin`.

kaffeekanne:
I think, that PR works quite well, don't know about `free()` and if there are some leaks. If that is not the case, lets commit it and improve it via PRs from master. There are things that can be improved by less talented C developers. 😃

src/command/cmd_funcs.c
jubalh: or maybe "Editor: %s does not exist"

src/config/preferences.c
jubalh: does `$EDITOR` work?

jubalh: Maybe also just use `xdg-open` I think it should work since you use `.txt` extension, right? Then its similar to other executables and user can set to their fav editor in `/executable`.

jubalh:
Also:

Test1:
* start profanity
* /msg someone
* /quit

Test2:
* start profanity
* /msg someone
* write a message using /editor
* /quit

In Test2 the terminal will not be cleared properly. There is still part of profanity visible. Maybe something ncurses related.

src/command/cmd_funcs.c
StefanKropp: Maybe /usr/bin/vi is a good idea. I would not use $EDITOR right now. I have not define a specification of $EDITOR / $VISUAL Maybe VISUAL would be much better than EDITOR.
I think we can change it later.

src/command/cmd_funcs.c
StefanKropp: > Too bad, continuing typing would be nice. But let that be a feature request later.
I think we can think about more features. Maybe /editor <file> also /editor correct to use the message correction in editor mode,.... we will see :-)

src/config/preferences.c
StefanKropp: > Maybe also just use `xdg-open` I think it should work since you use `.txt` extension, right? Then its similar to other executables and user can set to their fav editor in `/executable`.

Maybe we can recommend it. But I think vi will be on all systems. xdg-open is maybe not installed in headless systems.

src/command/cmd_funcs.c
jubalh: done in 8be46422d2ca5afff675ef3f7a2117c097a642a7

src/command/cmd_funcs.c
jubalh: done in e1249d9bbbe6a7a0741c8a83697f7bf962f4f82a

src/config/preferences.c
jubalh: true. also right now it doesnt seem to work with graphical editors (did a quick test with gedit only)

src/command/cmd_funcs.c
jubalh: done in 8be46422d2ca5afff675ef3f7a2117c097a642a7

jubalh:
I rebased + added some commits.

src/config/preferences.c
Mic92: Why not "vim" instead of "/usr/bin/vim"? This breaks on freebsd, GuixSD and NixOS because vim is not in `/usr/bin/vim`.
Also users are no longer able to have self-compiled vims.

DebXWoody: > Why not "vim" instead of "/usr/bin/vim"? This breaks on freebsd, GuixSD and NixOS because vim is not in `/usr/bin/vim`.
> Also users are no longer able to have self-compiled vims.

User can set the editor via `~/.config/profanity/profrc`.

```
[ui]
compose.editor=/path/editor
```

Mic92: Sure but what is the advantage of the absolute path?

jubalh: It doesn't "break" on freebsd etc. You just need to configure it ;). We need absolute path because of the function we use to call it.

PR #1487, opened 2021-02-15
OX: Bugfixes for OX implementation

StefanKropp:
- [x]  /ox char O leads to OX not implemented
- [ ] /ox k<tab> autocompletion starts cycling somewhere, but not with and option starting with the letter k
- [ ] /ox <tab> autocompletion does not start with letter a announce
- [x] /ox discover on jid without OX leads to OX: ERROR:
- [x] /ox request <jid> leads to JID and Fingerprint is required, so further documentation

### DIFF
--- a/src/xmpp/ox.c
+++ b/src/xmpp/ox.c
@@ -43,6 +43,9 @@
 #include "pgp/gpg.h"
 
 #ifdef HAVE_LIBGPGME
+
+#define KEYID_LENGTH 40
+
 static void _ox_metadata_node__public_key(const char* const fingerprint);
 static int _ox_metadata_result(xmpp_conn_t* const conn, xmpp_stanza_t* const stanza, void* const userdata);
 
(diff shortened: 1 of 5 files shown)


kaffeekanne:
Unsorted list of issues:

- [ ] `/ox char O` leads to `OX not implemented`
- [ ] `/ox k<tab>` autocompletion starts cycling somewhere, but not with and option starting with the letter `k`
- [ ] `/ox <tab>` autocompletion does not start with letter `a` `announce`
- [ ] `/ox discover` on jid without OX leads to `OX: ERROR:`
- [ ] `/ox request <jid>` leads to `JID and Fingerprint is required`, so further documentation
- [ ] Check for case inconsequencies of feedback messages `Public Key` and `Fingerprint` should be lowercase, i think

Feel free to comment or copy that list to the PR message. I will adjust the list and to actual testing of the feature. These cosmetic issues are just a quick start.

src/xmpp/ox.c
```diff
@@ -224,7 +229,7 @@ _ox_metadata_node__public_key(const char* const fingerprint)
 {
     log_info("Annonuce OpenPGP metadata: %s", fingerprint);
     assert(fingerprint);
-    assert(strlen(fingerprint) == 40);
+    assert(strlen(fingerprint) == KEYID_LENGTH );
```
jubalh, line 232: I dont think we want to assert? MAybe just print a notice and abort?

PR #1493, opened 2021-02-27
Draw wintitle using fputs and fprintf instead of /bin/echo.

akce:
This change simplifies wintitle so that it uses libc fputs and fprintf to draw the X-window titles instead of shelling out to /bin/echo.

I've tested wintitle show and goodbye on Void linux (glibc) with both rxvt-unicode and xterm.

### DIFF
--- a/src/ui/core.c
+++ b/src/ui/core.c
@@ -82,7 +82,6 @@
 #include "otr/otr.h"
 #endif
 
-static char* win_title;
 static int inp_size;
 static gboolean perform_resize = FALSE;
 static GTimer* ui_idle_time;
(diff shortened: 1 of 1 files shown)


src/ui/core.c
```diff
@@ -1079,51 +1078,35 @@ ui_contact_offline(char* barejid, char* resource, char* status)
 void
 ui_clear_win_title(void)
 {
-    printf("%c]0;%c", '\033', '\007');
+    fputs("\e]0;\a", stdout);
```
jubalh, line 1081: What's the advantage of this?

src/ui/core.c
```diff
@@ -1079,51 +1078,35 @@ ui_contact_offline(char* barejid, char* resource, char* status)
...
+    fflush(stdout);
 }
 
 void
 ui_goodbye_title(void)
 {
-    int result = system("/bin/echo -ne \"\033]0;Thanks for using Profanity\007\"");
-    if (result == -1)
-        log_error("Error printing title on shutdown");
+    fputs("\e]0;Thanks for using Profanity\a", stdout);
```
jubalh, line 1088: Anyone knows why echo was used at all?

src/ui/core.c
akce: It's slightly more efficient. ie, there's no reason here to use a printf to construct a string when the args are constants.

src/ui/core.c
akce: Not i.

src/ui/core.c
jubalh: Ah of course. Sorry I was confused due to different languages, thinking about print and printf.

jubalh (approved):

jubalh:
Thank you! I hope more improvements are coming! :)

PR #1242, opened 2019-12-19
Use OMEMO for offline MUC members

paulfariello:

PR #1495, opened 2021-03-05
MUC: Show offline members in sidebar

xenrox:
For small, private rooms I like to always see a list of members, even if they are offline. This works in e.g. conversejs.

I mainly got the idea from https://github.com/profanity-im/profanity/pull/1242

This PR is not fully working as intended: When first starting profanity it only shows a small subset of members. As soon as you let profanity redraw its window (change to fullscreen), the list becomes mostly correct (besides some double entries).

What is still missing: Option to toggle setting, code duplication, filter out online users from list

Alternative idea: Try to read what `/affiliation list` sees


jubalh:
> This PR is not fully working as intended: When first starting profanity it only shows a small subset of members. As soon as you let profanity redraw its window (change to fullscreen), the list becomes mostly correct (besides some double entries).

There is a function to force a redraw.

xenrox:
I am redrawing now at a point where the program should have the correct list returned by `muc_members`. Otherwise OMEMO encryption would not work in rooms.

xenrox:
Since dismail seems to be down right now, here is a status update:

Code deduplication is done and the offline list should be filtered correctly now.
The only thing that could be missing is a setting to toggle displaying the list.
If you want that, please tell me what command should be used.

jubalh:
> If you want that, please tell me what command should be used.

Maybe according to `/roster show|hide offline` we could have an `/occupants show|hide offline`.

xenrox:
I think that everything should be done now.

jubalh (approved):
Looks very good! Thanks for this patch!
The setting isn't saved per channel nor generally right now but for me that's okay.

jubalh:
Thanks! I hope you contribute more to Profanity! Do you plan to? ;)

xenrox:
> The setting isn't saved per channel nor generally right now but for me that's okay.

Generally should work right now with `/occupants default show offline` followed by `/save` or do you mean something different?

> Thanks! I hope you contribute more to Profanity! Do you plan to? ;)

Probably, since XMPP is my primary messenger now, especially if I find a bug or another interesting feature to work on.

jubalh:
> Generally should work right now with /occupants default show offline followed by /save or do you mean something different?

Oh seems I missed that haha.

> Probably, since XMPP is my primary messenger now, especially if I find a bug or another interesting feature to work on.

Glad to hear!

PR #1498, opened 2021-03-09
Remove "file://" from file names

mwuttke97:
Rationale: When copying an image in some application, a URL instead of a
path is copied to the clipboard.


jubalh (approved):

jubalh:
Thanks @mwuttke97

PR #1500, opened 2021-03-10
Fix HTTP upload

### LINKED ISSUE #1499: Profanity doesn't work with prosodys new http_upload module
Profanity doesn't work with prosodys new [http upload module](https://prosody.im/doc/modules/mod_http_file_share). As go-sendxmpp had the same problem before (which got fixed by @Zash) I assume profanity is also violating this part of [XEP-0363](https://xmpp.org/extensions/xep-0363.html):

> In addition to the Content-Length and Content-Type header the client MUST include all allowed headers that came with the slot assignment.

## Expected Behavior
Profanity should upload the file.

## Environment
* Debian Testing/Unstable

```
Profanity, version 0.10.0dev.master.96580f91

Build information:
XMPP library: libstrophe
PGP support: Enabled (libgpgme 1.14.0-unknown)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.9.2)
GTK icons: Enabled
```

mwuttke97:
This should hopefully fix #1499.


src/tools/http_upload.c
```diff
@@ -252,11 +272,21 @@ http_file_put(void* userdata)
     curl_easy_cleanup(curl);
     curl_global_cleanup();
     curl_slist_free_all(headers);
+
     if (fh) {
         fclose(fh);
     }
     free(content_type_header);
     free(output.buffer);
+    if (auth_header) {
```
jubalh, line 281: free(NULL) is a noop. So no need for `if`

jubalh:
```
Fix string STANZA_NS_HTTP_UPLOAD
":0" was missing at the end.
```
is an inappropriate description for the first commit.

In version 0.1.1 of XEP-0363 the namespace was just `urn:xmpp:http:upload namespace`. Only in a later version `urn:xmpp:http:upload:0` was introduced.

Regarding the second commit it seems that 0.5.0 of XEP-0363 allowed for these headers.

I would like to change the description to reflect what is actually changed. "Fix" sounds like something was broken. But it's more like adhering/updating to the newer version.

jubalh:
It's very good that you found this. In https://profanity-im.github.io/xeps.html we claim to support the latest version. Obviously we didn't. Until your PR ;)

jubalh:
So it seems to really support the namespace bump to `:0` then we need to change more in the stanza.

mwuttke97:
Thanks for all the feedback! Indeed, I didn't notice that the XEP actually was updated. I will implement the corresponding updates.

mdosch:
I experimented a little bit (without C, libstrophe or profanity code 
knowledge) and got to a point where profanity sends a (hopefully) valid 
stanza and gets a slot. But I don't know yet how to parse the reply in 
iq.c.

jubalh (changes requested):
Usually in git people use something like `Co-authored-by: name <name@domain.org>` for a commit that has more than one author :)

mwuttke97:
Do you know why the CI fails?

src/tools/http_upload.c
```diff
@@ -225,6 +244,7 @@ http_file_put(void* userdata)
     curl_easy_setopt(curl, CURLOPT_INFILESIZE_LARGE, (curl_off_t)(upload->filesize));
     curl_easy_setopt(curl, CURLOPT_UPLOAD, 1L);
 
+
```
jubalh, line 247: maybe one line too much :)

jubalh:
> Do you know why the CI fails?

No idea. Annoying.. Maybe python2 vs 3..

jubalh (approved):
Looks good. Except the one empty line too much. But doesnt' matter much :)

PR #1501, opened 2021-03-11
Add command to change password of logged in user

xenrox:
I used this to learn how to read XEPs and how stanzas work.
I followed https://xmpp.org/extensions/xep-0077.html#usecases-changepw

Open questions:
- How to name command/functions
- Should we confirm the password (ask user to enter it twice)
- Check connection encryption?

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -9298,3 +9298,30 @@ cmd_mam(ProfWin* window, const char* const command, gchar** args)
 
     return TRUE;
 }
+
+gboolean
+cmd_change_password(ProfWin* window, const char* const command, gchar** args)
+{
+    jabber_conn_status_t conn_status = connection_get_status();
+
+    if (conn_status != JABBER_CONNECTED) {
+        cons_show("You are not currently connected.");
+        return TRUE;
+    }
+
+    char* user = connection_get_user();
+    char* passwd = ui_ask_password(false);
+    char* confirm_passwd = ui_ask_password(true);
+
+    if (g_strcmp0(passwd, confirm_passwd) == 0) {
+        iq_register_change_password(user, passwd);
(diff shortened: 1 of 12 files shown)


jubalh:
>    How to name command/functions

`changepassword` looks fine to me.

>   Should we confirm the password (ask user to enter it twice)

Might be handy!

>  Check connection encryption?

 I wouldn't bother. User knows his settings and is logged in already.

xenrox:
Confirmation is in and CI is hopefully fixed as well (really miss SSHing into the build).

src/command/cmd_funcs.c
```diff
@@ -9298,3 +9298,30 @@ cmd_mam(ProfWin* window, const char* const command, gchar** args)
...
+
+    if (g_strcmp0(passwd, confirm_passwd) != 0) {
+        cons_show("Aborted! The new password and the confirmed password do not match.");
+        return TRUE;
+    }
+
+    iq_register_change_password(user, passwd);
+
+    free(user);
+    free(passwd);
```
jubalh, line 9324: Freein confirm_passwd is missing.

jubalh (changes requested):

src/command/cmd_funcs.c
xenrox: Now freed.

jubalh:
Two more problems:
You don't free in the case that passwords don't match.
You should turn it around and only send the IQ if they match. Then you dont need the return and can use the free each time.

Also a message upon success would be good :)

jubalh (approved):

jubalh:
Local `make check` passed. Code seems good. I thought maybe we should not allow empty passwords.. but you never know :-)

jubalh:
Thanks @xenrox !

PR #1502, opened 2021-03-11
Upload: Fix unused return

xenrox:


jubalh (approved):
We could omit the `headers = curl_slist_append(headers, expires_header);` in the case that asprintf fails. Because the header won't make much sense like this anyways. But since asprintf() is unlikely to fail and if it does we have more problems.. Let's leave it like this. It's easier.

jubalh:
Good catch! Thank you!

Issue #1211, opened 2019-10-21
Add support for XEP-0045 7.13 "Requesting voice"


mdosch:
I would appreciate support for [XEP-0045 7.13](https://xmpp.org/extensions/xep-0045.html#requestvoice):

> It is not possible for a visitor to speak (i.e., send a message to all occupants) in a moderated room. To request voice, a visitor SHOULD send a <message/> stanza containing a data form to the room itself, where the data form contains only a "muc#role" field with a value of "participant".

Some MUCs have switched to moderated and deactivated PMs due to spam and trolling. So you have to find some moderator in another MUC with no disabled PMs to request voice which is no good UX (and sometimes annoys the moderators "why don't you use the voice request of your client?"). Afaik Gajim is the only client supporting this so far.

jubalh:
Thanks for the request. Makes sense to have this.
I'll close this until we are down to 50 ;)
Feel free to reopen then.

PR #1504, opened 2021-03-12
MUC: Add voice request

xenrox:
closes https://github.com/profanity-im/profanity/issues/1211


xenrox:
This is not as well tested as I would like because I don't have a client that supports answering those requests, but the `xmlconsole` output looks good, and the server returns a nice error message, when you send requests too fast.

xenrox:
I should probably add a check if the user already has an affiliation.

jubalh (approved):
Looking good. Thank you!

PR #1505, opened 2021-03-12
Fix: Chatlog for MUC

StefanKropp:
Setting message->stanzaid


jubalh:
See 3884a4d35bce3e306fc1b7e02d55184bc7a90f19

jubalh:
This was most likely also the reason why there were NULL entries for
this in the DB which DebXWoody mentioned in the MUC.

Thus comparison was with NULL and no new entries were added to the
database.

PR #1514, opened 2021-03-26
Include config.h in omemo files

jubalh:
So that off_t is defined in all files (so far http_upload.c and
http_download.c too) the same way.


jubalh:
Will check the other files too. @mwuttke97 you could test these already.

Issue #1185, opened 2019-09-08
Alert user if sending encrypted messages when none/not all fingerprints are trusted


quite:
The original item in #1070 read "Alert user if sending encrypted messages with not all keys trusted".

Given that messages are only ever encrypted with keys that are trusted, I interpret this as: the user should be alerted if they try to sent a message to a peer, for which fewer than all fingerprints have been trusted (not counting any explicitely untrusted fingerprints!).

The message could include both the number of fingerprints trusted and non-trusted (excluding untrusted).

It would serve as a reminder to the user to remember to trust a fingerprint before sending a message. But also to explicitely untrust fingerprints that should not be encrypted for.

Now that I'm writing this, I realise that perhaps this is a bit harsh? Should the user perhaps only be alerted if there are no trusted fingerprints, and the dummy message "You received a..." is all that got through (to any device).

jubalh:
@paulfariello :-)

kaffeekanne:
I think on new new not decided keys, the user should get an alert inside the chat window, and the message should not be sent. This forces the user to do something about that keys. Then the message can be resend from the command history. This way the receiver will get the message on all trusted devices. Otherwise maybe, we wont send to the new mobile phone and the information is read much later on a desktop client. Yeah, there still is a dummy message, but do normal users always ask about that or just wait and hope messaging might heal itself.

jubalh:
@paulfariello could we also just make a check at the beginning when the OMEMO session is started. And then print a message like "fingerprint X" is not trusted. Or won't that catch all cases? For example Profanity is running on a server. Someone gets a new phone, puts conversations on it and gets a new omemo fingerprint. The session will still be initialized so no message?

jubalh:
@DebXWoody you did this recently, right?

DebXWoody:
Am Mittwoch, den 09.06.2021 um 14:50:31 -0700 schrieb ***@***.***:
> @DebXWoody you did this recently, right?
> 

No, this hasn't been implemented. I think @mwuttke97 has implemented a
message, if we didn't find a key at all.

Personally, I also think we shouldn't print a message during sending.
Maybe I have a buddy which I decided that I wouldn't trust his mobile
phone.

I saw that we get a `headline` message from users. I will try to
implement this. It should be possible that we print a message, "User X
has got a new device (device id): Fingerprint.", just once in the
console.

We should also add a flag to the known devices:

 * unknown
 * trusted - has been trusted by user
[...]

PR #1515, opened 2021-03-29
 [OMEMO] Fix sending messages

### LINKED ISSUE #1436: OMEMO error with my other client
## Expected Behavior
When I send an OMEMO-encrypted message to a group chat or to a contact directly, they should be able to see the message (and my other devices should as well).  I am using profanity on one device and conversations on another, and my other contact is using just conversations.

## Current Behavior
When I send an OMEMO-encrypted message to a group chat that includes a certain one of my contacts, or to that person directly, they only get a "message not encrypted for this device" message.  I also get the same message on my other device when using a group chat, but it works (on my device) when sending a message to that contact directly.

## Possible Solution
No idea.

I have no idea how to reproduce this, other than that it is very reproducible with my contact and me.  I can provide more debugging information if needed.

1. Profanity opens the group chat automatically, and it defaults to omemo
2. I send a message in the group chat's tab

## Context
I am still able to talk encrypted with my contact through my conversations instance, but it is a little annoying and worrying.

I was also able to do some debugging with the help of the profanity MUC.  I found that this message gets output to the profanity log:
```
prof: WRN: OMEMO: cannot find device ids for <contact jid>
```
I also found a weird thing when I enabled debug logging.  There are two other people in this group chat, not including me.  I see the stanza sent in the debug log, and there are four `key` xml tags.  Each of the devices of the contact that works is listed twice, with none for my other device or the one device of the contact that gets the error message.  Not sure if this is intended, or if this could be (part of) the problem.

## Environment
```
Profanity, version 0.9.5
Copyright (C) 2019 - 2020 Michael Vetter <jubalh@iodoru.org>.

Build information:
XMPP library: libmesode
PGP support: Enabled (libgpgme 1.14.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.8.6)
GTK icons: Disabled
````
* Operating System/Distribution: Arch Linux
* glib version: glib2 2.66.2-1

mwuttke97:
Fix #1436

- [x] Fix the race condition (don't start two sessions in parallel)
- [x] Warn the user in the chat window (not only in the log file) if there where errors when encrypting the messages (related: #1185)
- [x] Don't encrypt to own device

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -2151,22 +2151,21 @@ cmd_msg(ProfWin* window, const char* const command, gchar** args)
 
         ProfChatWin* chatwin = wins_get_chat(barejid);
         if (!chatwin) {
-            // NOTE: This will also start the new OMEMO session
-            // and send a MAM request.
+            // NOTE: This will also start the new OMEMO session and send a MAM request.
             chatwin = chatwin_new(barejid);
         }
         ui_focus_win((ProfWin*)chatwin);
 
         if (msg) {
-            // FIXME [OMEMO] We can't be sure whether the
-            // bundles have already been receieved. Thus, it is
-            // possible (and probable) that the recipent can't
-            // encrypt the message.
+            // NOTE: In case the message is OMEMO encrypted, we can't be sure
+            // whether the key bundles of the recipient have already been
+            // received. In the case that *no* bundles have been received yet,
+            // the message won't be sent, and an error is shown to the user.
+            // Other cases are not handled here.
             cl_ev_send_msg(chatwin, msg, NULL);
         } else {
 #ifdef HAVE_LIBOTR
-            // Start the OTR session after this (i.e. the
-            // first) message was sent
+            // Start the OTR session after this (i.e. the first) message was sent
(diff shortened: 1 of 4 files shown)


jubalh:
@mwuttke97 please see the comments above.

For the commit where you do the change with the otr (`is_otr_secure = otr_is_secure(barejid);`) please add in the commit body why you think this change is needed and where you copied it from.

mwuttke97:
On 07/04/2021 09:26, Michael Vetter wrote:
> @mwuttke97 <https://github.com/mwuttke97> please see the comments above.
>
> For the commit where you do the change with the otr (|is_otr_secure = 
> otr_is_secure(barejid);|) please add in the commit body why you think this change is needed and where you copied it from.
>

I've fixed the typos and improved the messages. (Apparently, I pressed CTRL+N or CTRL+P a couple of times too often.)

Regarding the `is_otr_secure` thing. I clarified the comment and the error message a bit. I think the code+comment are quite clear and don't deserve an extra explanation / commit message. However, I am not sure whether these contradicting settings (OTR + OMEMO configured for one contact) are reachable.

jubalh (approved):

PR #1523, opened 2021-04-18
Fix serversoftware examples

mdosch:
s/software/serversoftware


jubalh:
oh sorry! I just got messaged about this via PM and then did it. Only saw this PR now :X

PR #1526, opened 2021-04-18
OMEMO Rework

StefanKropp:


jubalh:
Cool!
I think the `Rework` in each commit isn't necessary. Just state what you changed in subject line :)
And in the message body additionally to what you did (`Moved omemo_devicelist_subscribe after loading OMEMO files`) I would also like to understand *why* you did it :)

jubalh:
@StefanKropp I updated the commit messages. Please pull before pushing.

tests/unittests/omemo/stub_omemo.c
```diff
@@ -114,3 +114,11 @@ omemo_encrypt_file(FILE* in, FILE* out, off_t file_size, int* gcry_res)
     return NULL;
 };
 void omemo_free(void* a){};
+
+uint32_t
+omemo_device_id() {
+	return 123;
+}
+
```
jubalh, line 122: why so many empty lines?

StefanKropp:
> Please squash the last two commits into one and replace `SENT` with `SEND`.
> Do we want all this as info? Some looks more like debug?
I recommend we use this branch as a kind of "testing" brunch. In debug mode there are do much information.
When we understood the code and found the problem, we can "clean-up" the code and provide a clear commit for those changes.

jubalh:
Ok, so we will do this before the merge :)

jubalh:
Done again and merged in: https://github.com/profanity-im/profanity/pull/1539

PR #1528, opened 2021-04-28
Apply pre_chat_message_display hook to history

dustinlagoy:
I find it helpful to have the plugin hook pre_chat_message_display apply to messages shown in a chat's history (via _chatwin_history) and not just for new incoming messages (in chatwin_incoming_msg). There is just one line here but I can try and add some tests or anything else if someone thinks it would be useful.

Could this break any current plugins?

On a related note, a similar change could be made to pre_room_msg_display

### DIFF
--- a/src/ui/chatwin.c
+++ b/src/ui/chatwin.c
@@ -519,6 +519,7 @@ _chatwin_history(ProfChatWin* chatwin, const char* const contact_barejid)
 
         while (curr) {
             ProfMessage* msg = curr->data;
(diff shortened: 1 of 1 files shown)


jubalh:
I see why you need this. I'm not sure whether it's the best solution because I don't know the plugin code at all. It could be. It could also be that we would need to mark either via parameter or new function that the messages are from history and not new ones.

Let's merge this for now and see what happens :-)

jubalh (approved):

PR #1529, opened 2021-04-29
Access roster from plugins

dustinlagoy:
This adds two functions to the "prof" plugin API:
```c
char* prof_get_name_from_roster(const char *barejid);
```
and
```c
char* prof_get_barejid_from_roster(const char *name);
```
as well as the equivalent functions for the Python API.

I can add some tests if this functionality is otherwise mergable and of course fix anything else too!

There is a potential inconsistency here because these functions inherit the convention of the `xmpp/roster_list.c` functions that they call. `prof_get_name_from_roster` returns the input barejid if it is not found in the roster, but `prof_get_barejid_from_roster` returns NULL if the name is not in the roster. This makes sense to me but is potentially confusing.

### DIFF
--- a/apidocs/c/profapi.h
+++ b/apidocs/c/profapi.h
@@ -142,7 +142,7 @@ char* prof_get_current_nick(void);
 Retrieve the nickname for a given barejid if it is in the roster.
 @return the users nickname e.g. "eddie", or NULLL if the barejid is not in the roster.
 */
-char* prof_get_nick_from_roster(const char *barejid);
+char* prof_get_name_from_roster(const char *barejid);
 
 /**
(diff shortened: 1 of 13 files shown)


jubalh:
@dustinlagoy cool! It's highly appreciated that you improve the plugin code! I hope you can do more like this.

BTW Do you mind transforming this to a draft PR and setting it to a regular PR once you want your first reviews? It's easier for me this way to see when I need to do something.

jubalh:
@dustinlagoy ping :)

jubalh (approved):

jubalh:
Thanks a lot @dustinlagoy !

Issue #1516, opened 2021-03-30
Feature request: Option to control intype separately for console and titlebar


mdosch:
Right now you can either enable or disable typing notifications for console and titlebar together:

> 30.03.21 21:07:53 - Synopsis
> 30.03.21 21:07:53 - /intype on|off
> 30.03.21 21:07:53 - 
> 30.03.21 21:07:53 - Description
> 30.03.21 21:07:53 - Show when a contact is typing in the console, and in active message window.

I would like to be able to enable or disable typing notifications for console and titlebar separately, e.g. like that:

> /intype titlebar|console on|off

PR #1532, opened 2021-05-08
Have separate settings for intype (console/titlebar)

jubalh:
See https://github.com/profanity-im/profanity/issues/1516
This PR also sets enables intype by default.


jubalh:
@mdosch looking for feedback.

PR #1534, opened 2021-05-12
Use direct messages instead of MUC-PMs for non-anonymous MUCs

jubalh:


jubalh:
I tested:
* Writing to a person in an anonymous muc -> start mucpm
* Writing to a person in an non-anonymous muc -> start direct message

Would be great if someone could test with OMEMO.

mdosch:
Worked for me (including OMEMO).

Issue #1538, opened 2021-05-26
[OMEMO] profanity should fix the PEP access model if wrongly configured by another client


g4570n:
Me and my friend use the same version of profanity, we both have two OMEMO keys, one from another client and one created in profanity. When we do `/omemo fingerprint <account>` we only get the fingerprint of another client, but the profanity of the other is not listed. 

But if we both do: `/omemo fingerprint` in the console, if the hash of profanity and that of the other client we use come out

I followed the steps detailed here: https://github.com/profanity-im/blog/blob/master/content/post/omemo.md , but it only lists one of the IDs of your mobile device but not that of your profanity client.

When we want to use Omemo and he writes to me from his profanity client I get this message: "You received a message encrypted with OMEMO but your client doesn't support OMEMO." `¯\_(ツ)_/¯`

```
$ uname -a
OpenBSD 6.9 GENERIC#464 amd64

$ profanity --version
Profanity, version 0.10.0
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2021 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libmesode
Desktop notification support: Disabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.15.1)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.8.8)
GTK icons: Disabled
```

ffuentese:
Also it's not possible to add the other device's fingerprint because it's "unknown"

DebXWoody:
Hello,
I'm working on OMEMO bug-squashing. Feel free to join the party ;-) I would like to check some things with you.

Is the device listed in the file `~/.local/share/profanity/omemo/<YOUR_JID>/known_devices.txt`?

g4570n:
> Is the device listed in the file `~/.local/share/profanity/omemo/<YOUR_JID>/known_devices.txt`?

At the beginning of this file are listed mine, both, Gajim's and profanity's. But of my buddy's devices, there is only one listed there

DebXWoody:
Use `/xmlconsole` to open the XML Console.

Prepare the following line ( set `from` and `to` attributes) (one line!)

`
[...]

Issue #1530, opened 2021-05-07
Can't send OMEMO messages after reconnect


mdosch:
Whenever profanity reconnects I can't send OMEMO encrypted messages afterwards. I can provide you a log file (but I won't publish it) if you contact me via XMPP or Email: martin@mdosch.de

## Expected Behavior
OMEMO still works after reconnect.

## Current Behavior
Profanity tells me it doesn't have any key to encrypt to the contact. I have to do `/omemo end` and `/omemo start` to get it working again.

## Steps to Reproduce (for bugs)

1. Start profanity
2. Configure autoping
3. Send OMEMO encrypted messages to a contact
3. Switch of WiFi or unplug Ethernet to force a connection loss
4. Wait till profanity realizes it's disconnected
5. Reconnect the machine to the internet
6. Wait for the reconnect to succeed
7. Try to send another OMEMO encrypted message to the contact from 3.

## Environment

Profanity, version 0.10.0dev.master.1ef700f9
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2021 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.14.0-unknown)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.9.2)
GTK icons/clipboard: Enabled

DebXWoody:
I think we need to do a "omemo session start" for all open windows where the chat mode is "OMEMO".

Issue #1070, opened 2019-04-17
OMEMO feature tracker


jubalh:
Original issue requesting OMEMO support: https://github.com/boothj5/profanity/issues/658
General OMEMO support was implemented in https://github.com/boothj5/profanity/pull/1039

Some features are missing.
This tracker bug is there to give an overview of what we have and what we are working on.

- [x] OMEMO policy https://github.com/boothj5/profanity/pull/1062
- [ ] filter fingerprints that we autocomple https://github.com/boothj5/profanity/issues/1068
- [ ]  /omemo trust [tab]... in a 1:1 chatwindow should only cycle through partners fingerprints. currently cycles through all fingerprints
- [ ] autocompletion for untrust command
- [ ] Handle error with publish-option for device_list
- [ ] Alert user if sending encrypted messages with not all key trusted
- [ ] Send files encrypted [proto xep](https://xmpp.org/extensions/inbox/omemo-media-sharing.html) [xep-0396](https://xmpp.org/extensions/xep-0396.html)
- [ ] OMEMO encryption for MUC PM
- [ ] OMEMO MUC should encrypt to all jid present in affiliation. Not only those currently connected.
- [ ] Messages are not decrypted unless we trust the sending key. We should probably decrypt everything but mark it.

- [ ] Update website with OMEMO usage docu https://github.com/boothj5/profanity/issues/1048

kaffeekanne:
Even if it might not be implemented i would add trusting without mutual roster entry/subscription. Background: You would not want to cludder up your roster with every aquaintance of a greater MUC, although you share common interes/goal that leads you to taking part in a E2EE MUC.

Since MUCs need to be non anonymous you get the JID of the MUC participants anyway and could manage the keys the way it is done already.

paulfariello:
@kaffekanne this issue should have been solved when I added support for `publish-option` in device list and bundle. Can you test it?

kaffeekanne:
Do you have a branch or PR for it already?

moppman:
Is there a way to trust your account's own fingerprints from other clients you use? 
Right now, omemo messages are only encrypted for profanity and the chat partner, but not for my other clients.
If I try `/omemo fingerprint` in the main window, my account's fingerprints are listed. However, I can't trust them with `/omemo trust <fingerprint>`.
`profanity.log` says:
`prof: ERR: OMEMO: cannot encrypt key for my@account device id <id> ` for all my other devices.

optmzr:
[...]

PR #1539, opened 2021-05-28
Improved OMEMO logging

DebXWoody:
The most loggings has been set to debug. Just in case of error / problem the
logging should be done with log-level ERROR or WARNING.

We will see later, which loggings should be switched to log-level INFO.

This improvement has been done to find OMEMO bugs.

See also: #1070 #1530 #1538


jubalh (approved):

jubalh:
Awesome! Very good commit and description and everything! Top, thanks!

PR #1540, opened 2021-05-28
Display OMEMO device ID

DebXWoody:
Display the OMEMO device ID which has been generated, when the user generated
OMEMO crytographic materials via /omemo gen.


jubalh (approved):

jubalh:
Thanks!

Issue #1541, opened 2021-05-28
OMEMO excluding sending device in groupchats


DebXWoody:
## Expected Behavior
https://xmpp.org/extensions/xep-0384.html#encrypt says

_(i.e. including their own devices node, but excluding itself)_

## Current Behavior

This part has been done in e8664e273 for 1-to-1 `chat`, but not for `groupchat`.

Sending a message to a groupchat:

```
28/05/2021 20:19:35: prof: DBG: [OMEMO] Sending from device sid 907785502
28/05/2021 20:19:35: prof: DBG: [OMEMO] Sending to device rid 907785502
...
```

## Possible Solution
Need to be fixed in the part above "Encrypt keys for the recipients".

I will take this task.

PR #1542, opened 2021-05-28
OMEMO: Don't encrypt to yourself (MUC)

DebXWoody:
As defined in XEP-0384 the application should not encrypt the message to own
devices. Within a groupchat, yourself are a recipients as well.

We will check the recipients and filter out the own device of the own jid.

This Pull Request will fix Issue: #1541

### DIFF
--- a/src/omemo/omemo.c
+++ b/src/omemo/omemo.c
@@ -791,6 +791,19 @@ omemo_on_message_send(ProfWin* win, const char* const message, gboolean request_
                 .device_id = GPOINTER_TO_INT(device_ids_iter->data)
             };
 
+            // Don't encrypt for this device (according to
+            // <https://xmpp.org/extensions/xep-0384.html#encrypt>).
+            // Yourself as recipients in case of MUC
+            Jid* me = jid_create(connection_get_fulljid());
+            if ( !g_strcmp0(me->barejid, recipients_iter->data) ) {
+                if (GPOINTER_TO_INT(device_ids_iter->data) == omemo_ctx.device_id) {
+                    jid_destroy(me);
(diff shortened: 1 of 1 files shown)


DebXWoody:
Needs test and review

src/omemo/omemo.c
```diff
@@ -791,6 +791,19 @@ omemo_on_message_send(ProfWin* win, const char* const message, gboolean request_
                 .device_id = GPOINTER_TO_INT(device_ids_iter->data)
             };
 
+            // Don't encrypt for this device (according to
+            // <https://xmpp.org/extensions/xep-0384.html#encrypt>).
+            // Yourself as recipients in case of MUC
+            Jid* me = jid_create(connection_get_fulljid());
```
jubalh, line 797: You could use `connection_get_barejid()` which returns a `char*` with the barejid. Then you don't need to build it yourself.

jubalh (changes requested):

mwuttke97 (approved):
I tested the same logic (but w/o the fixed memory leak) in the branch `StefanKropp/omemo-rework`.  I encountered no problems so far, but haven't tested this in a MUC.

src/omemo/omemo.c
DebXWoody: Done

DebXWoody:
What about MAM? :thinking:

jubalh (approved):

jubalh:
> What about MAM? thinking

Gets filtered out via MUC reflection. We add the message already to the SQL DB when we are sending out.
And when displaying in a window we never get from MAM directly (just somtimes send a MAM request from which we will the SQL DB), always from SQL DB.

DebXWoody:
Is there something still open or can we merge?

jubalh:
Merged :)

PR #1543, opened 2021-05-29
Feature - attention flag

DebXWoody:
The command `/wins attention` will display all important (flagged) chats and groupchat.
The user can toggle the attention flag via `ALT+F`.


DebXWoody:
# Usage
Just open a chat window or a groupchat window. You can toggle the attention flag via ALT+F.

```
21:23:17 ! Attention flag has been deactivated
21:23:18 ! Attention flag has been activated
```
With the `/wins attention` command you can see all windows where the attention flag is set - displayed in the console window.

With ALT+M you should be able to circle the flagged windows.

- [ ] There is a bug, it's just working with 2 rooms

mdosch:
Works well but if you do `/wins attention` you'll have an empty line before the list. If no windows are flagged it'll print only an empty line. Maybe it should print "Windows with attention flag:" if there are flagged windows and "No flagged windows." if there are non there instead of an empty line. 

2021-05-30-194350_scrot

mdosch:
Just realized the `[ATTENTION]` in the header bar is shown for 1-1 but not for MUCs. Is this intentional?

DebXWoody:
> Just realized the `[ATTENTION]` in the header bar is shown for 1-1 but not for MUCs. Is this intentional?

Yes, I just implemented in 1-to-1 `chat` and not in `groupchat`. 
Maybe the idea of  `[ATTENTION]` is not bad, I can implement in `groupchat` also.

jubalh:
I'm unsure whether we really want the notice inside the window or whether the titlebar would be enough. I think it clutters the window too much. But we can change this later.

jubalh (approved):
Did a couple of commits too. Should be fine now.

jubalh:
Thanks @DebXWoody !

PR #1548, opened 2021-06-01
OMEMO - Unable to publish own device list

DebXWoody:
If we are not able to publish our own device list, we just ignored it.
This commit will show at least an error message and informs the user that this
device is currently not in the list of devices.

See Issue: #1538

Next task will be to handle the <precondition-not-met/> conflict properly.


jubalh (approved):
I changed some sentences

Issue #1231, opened 2019-11-25
Wrong message display


mdosch:
I received MUC messages which were displayed wrong. The date was also "injected" in the message.

## Expected Behavior
The date should not be "injected" in the message.

## Current Behavior

![2019-11-25-181008_scrot](https://user-images.githubusercontent.com/5103003/69562382-5819ca00-0faf-11ea-835f-191b72e270dd.png)

## Context

```
grep Kter profanity.log.99 
25/11/2019 17:38:44: xmpp: DBG: RECV: <message lang="en" to="REMOVED/rF9sqBAM" type="groupchat" from="conversations@conference.siacs.eu/Licaon_Kter [web]"><inactive xmlns="http://jabber.org/protocol/chatstates"/><no-store xmlns="urn:xmpp:hints"/><no-permanent-store xmlns="urn:xmpp:hints"/></message>
25/11/2019 17:38:44: prof: ERR: Message received without body for room: conversations@conference.siacs.eu/Licaon_Kter [web]
25/11/2019 17:44:08: xmpp: DBG: RECV: <presence lang="en" type="unavailable" to="REMOVED/rF9sqBAM" from="conversations@conference.siacs.eu/Licaon_Kter [web]"><x xmlns="http://jabber.org/protocol/muc#user"><item role="none" affiliation="none"/></x></presence>
25/11/2019 17:44:08: prof: DBG: Unavailable presence handler fired for conversations@conference.siacs.eu/Licaon_Kter [web]
25/11/2019 17:44:08: prof: DBG: Room presence received from conversations@conference.siacs.eu/Licaon_Kter [web]
25/11/2019 18:02:59: xmpp: DBG: RECV: <message id="812fe59f-55d4-4e26-a749-2e084ec9cbef" lang="en" to="REMOVED/rF9sqBAM" type="groupchat" from="operators@muc.xmpp.org/Licaon_Kter"><origin-id id="812fe59f-55d4-4e26-a749-2e084ec9cbef" xmlns="urn:xmpp:sid:0"/><body>Martin: i think the massage comes from /me saying &quot;buy coins from Martin@domain&quot;</body><stanza-id id="2019-11-25-45c772389fbcdf1d" by="operators@muc.xmpp.org" xmlns="urn:xmpp:sid:0"/></message>
25/11/2019 18:02:59: prof: DBG: Attempting notification: Licaon_Kter in operators (win 20)
[...]

PR #1557, opened 2021-06-08
Fix wrong message display when mention is triggered

jubalh:
Fix https://github.com/profanity-im/profanity/issues/1231


jubalh:
@mdosch please test.

Issue #1525, opened 2021-04-18
Show XEP-0353 Jingle Message Initiation


mdosch:
Even if profanity can't (yet?) do a/v calls I'd like to be able to see when a contact tries to initiate an [XEP-0353 Session](https://xmpp.org/extensions/xep-0353.html#intent).

The usecase: Often my phone is muted in another room or my jacket and I might not realize I'm getting called via XMPP even if I'm sitting in front of my computer using one of the best XMPP clients. :)

PR #1560, opened 2021-06-09
XEP-0353: Display a notice when receiving a call

jubalh:
Display a notice in the console when someone tries to call us.

Implement https://github.com/profanity-im/profanity/issues/1525


mdosch:
Works for me. :)

PR #1561, opened 2021-06-11
Escape all parameters in SQL statements

CIacademic:
Some more parameters might contain quotes, so just escape all of them with %q by using sqlite3_mprintf.

For more information about %q see https://sqlite.org/printf.html


jubalh:
Cool, I wasn't aware of this. Thanks!

Issue #1551, opened 2021-06-05
OMEMO - Check implementation - headline / devicelist


DebXWoody:
When I remember well, profanity asks the buddy`s via <iq/> to get the OMEMO device list.

The OMEMO device list will be send via headline from the buddy to the user.

`
<message type="headline" to="user1@domain/profanity.37lb" from="user2@domain"><event xmlns="http://jabber.org/protocol/pubsub#event"><items node="eu.siacs.conversations.axolotl.devicelist"><item id="current"><list xmlns="eu.siacs.conversations.axolotl"><device id="179620673"/><device id="260570434"/><device id="898388139"/><device id="1185166165"/><device id="176266165"/></list></item></items></event><addresses xmlns="http://jabber.org/protocol/address"><address jid="user2@domain/abc" type="replyto"/></addresses></message>
`

- [ ] Check the implementation of `headline` to get the device list via `headline`

DebXWoody:
It would be very nice if we will be able to get notifications about device list changes. I will take a look.

jubalh:
When do we get this send?
What do we want to display?

jubalh:
Any news?

jubalh:
Not sure what should be done here.
Feel free to reopen/comment in case there is still work to be done.

PR #1562, opened 2021-06-12
OMEMO headline

DebXWoody:
Online users will send a list of devices via PubSub `headline`. We will get this just after connection from the online users, but also when a user generates a key for a new device, we will be informed.

Bugfix: This pull request fix the handling of messages to be able to get the `event` messages for the device list.

Feature: Inform the user (console) when a new device as been found. 
#1551

### DIFF
--- a/src/omemo/omemo.c
+++ b/src/omemo/omemo.c
@@ -541,6 +541,27 @@ omemo_set_device_list(const char* const from, GList* device_list)
         jid = jid_create(connection_get_fulljid());
     }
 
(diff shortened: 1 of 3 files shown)


DebXWoody (commented):
Abbrechen!!!!! Ich will hier keine doofen Fenster!!!

jubalh:
I'm not sure what's the plan with this one.
I think there were other commits regarding the headline done by several people (including you?) since this PR got created.

Feel free to reopen/continue working on this if still applicable!

Issue #1434, opened 2020-10-22
Feature request: XEP-0377: Spam Reporting


mdosch:
I'd appreciate to see [XEP-0377: Spam Reporting](https://xmpp.org/extensions/xep-0377.html) supported in profanity.

Maybe a command `/blocked report` could be introduced which would add the spammer to the blocklist and sends the report.

jubalh:
So far untested.

PR #1569, opened 2021-06-30
Add XEP-0377: Spam Reporting

jubalh:
Report and block:
`/blocked add someone@domain.org report-abuse This is not nice`
`/blocked add someone@domain.org report-spam This is not nice`

Regular block:
`/blocked add someone@domain.org`

Implement https://github.com/profanity-im/profanity/issues/1434


jubalh:
@mdosch can you test? :)

jubalh:
Maybe `blocked add` should stay like it was and we should use:
`blocked report abuse|spam <jid> <msg>`

Makes autocompletion easier.

jubalh:
Seems like https://modules.prosody.im/mod_spam_reporting.html uses an older version of the XEP (which was only recently updated).

PR #1570, opened 2021-07-01
Restore last status string when connecting

xenrox:


src/config/accounts.c
```diff
@@ -844,8 +844,6 @@ accounts_set_last_status(const char* const account_name, const char* const value
     if (accounts_account_exists(account_name)) {
         if (value) {
             g_key_file_set_string(accounts, account_name, "presence.laststatus", value);
-        } else {
-            g_key_file_remove_key(accounts, account_name, "presence.laststatus", NULL);
```
jubalh, line 848: I'm not really in this part of the code so I might ask some stupid questions.
Here is the first one: why remove this?

src/config/accounts.c
```diff
@@ -952,6 +950,17 @@ accounts_get_login_presence(const char* const account_name)
     return result;
 }
 
+char*
+accounts_get_login_status(const char* const account_name)
+{
+    gchar* setting = g_key_file_get_string(accounts, account_name, "presence.login", NULL);
+    gchar* status = NULL;
+    if (strcmp(setting, "last") == 0) {
```
jubalh, line 958: maybe use g_strcmp0() in case `setting` is NULL?

jubalh (commented):
Few small questions.
BTW do I see this right that `/accout set <account> status last` never worked then?

xenrox:
> Few small questions.
> BTW do I see this right that `/accout set <account> status last` never worked then?

No restoring the status works quite well, the problem here is the status message.

src/config/accounts.c
xenrox: I used normal `strcmp` because that was done [here](https://github.com/profanity-im/profanity/blob/22cda3dbea49d8ac9382117427ec4889bf999ff8/src/config/accounts.c#L939), but `g_strcmp0` would probably be better.

src/config/accounts.c
xenrox: Well that is the tricky part and a reason why I thought that maybe setting status messages should be refactored. 
I am not sure why that part is there, but my guess would be for overwriting an older status message like: `/status set online` would overwrite an older setting `/status set away eating`. But currently using `/status set online` automatically implies `/status set online online`.
The reason why it was removed is that this function is called without a value before we try to restore the last status which will remove the last status message from our config file.

xenrox: Seems like the function is for auto away: https://github.com/profanity-im/profanity/commit/54e225aa5330b7a3c20bda38264e5ad0cfdfcdc5
Probably so that something like "away (ready to talk)" does not happen. But I don't get why that should remove the setting from the file.

jubalh (approved):
Awesome! Thanks for your work!

PR #1573, opened 2021-07-01
Initialize jid

xenrox:
Someone broke compiling :O


jubalh:
That's only to silence the compiler.
The real error is that I forgot a return in:
```
        } else {
            cons_bad_cmd_usage(command);
        }
```

jubalh:
Thanks!

PR #1574, opened 2021-07-01
Add in-band account registration

### LINKED ISSUE #199: In-band account registration
Allow for in-band account registration as per:
http://xmpp.org/extensions/xep-0077.html

binex-dsk:
Closes #199

currently broken but making PR anyway because i dont feel like doing this anymore for now

checklist:
- [x] Add `auth` param like `/connect` does
- [x] Switch `host` and `user` positions?
- [ ] support for forms
- [ ] Feature & error checking -- ensure in-band registration is enabled
- [ ] Support for simple JID registration; `/register user@xmpp.serv.er` instead of `/register user xmpp.serv.er`
- [ ] Possibly remove the `auth` parameter actually, because there's no authentication occuring

bugs:
- [x] SIGABRT after confirming password, in `xmpp_run_once` (caused by accidental freeing of one of the args in `cmd_register`)
- [ ] The raw connection to the server finishes AFTER the registration process (!!)
- [ ] Does not actually register at all - caused by the previous issue
- [ ] Connection is left in a `JABBER_RAW_CONNECTED` state afterwards--must be set to `JABBER_DISCONNECTED` (and, well, it should be disconnected) after registration has finished, but it's not possible to do this unless bug 2 is fixed. 
- [ ] After the register command has been run once, if it is run again, a segfault occurs, in `xmpp_conn_release(conn.xmpp_conn)` -> `conn_disconnect(conn)` -> `conn->ctx, "xmpp", "Closing socket.")` -> `xmpp_log(ctx, XMPP_LEVEL_DEBUG, area, fmt, ap)` -> `ctx->log->handler(ctx->log->userdata, level, area, buf)`

OTHER STUFF/NOTES
- added a `connection_connect_raw` function for connecting to a server without authenticating as a user, as well as corresponding `session_connect_raw` and `cl_ev_connect_raw` functions
  * `JABBER_RAW_CONNECTING` and `JABBER_RAW_CONNECTED` states were also added, self explanatory as they indicate a raw, non-authenticated connection was made
- some added functions are very similar to other already existing ones--maybe possible to merge them?

### DIFF
--- a/src/xmpp/iq.c
+++ b/src/xmpp/iq.c
@@ -2658,14 +2658,15 @@ _mam_rsm_id_handler(xmpp_stanza_t* const stanza, void* const userdata)
 void
 iq_register_new_account(const char* const user, const char* const password)
 {
-    //char* id = connection_create_stanza_id();
     xmpp_ctx_t* const ctx = connection_get_ctx();
     xmpp_stanza_t* iq = stanza_register_new_account(ctx, user, password);
 
     const char* id = xmpp_stanza_get_id(iq);
-    iq_id_handler_add(id, _register_new_account_result_id_handler, NULL, NULL);
+    iq_id_handler_add(id, _register_new_account_result_id_handler, NULL, NULL); // FIXME: function doesn't seem to ever run?
 
+    log_debug("HI hi sending registration stanza");
     iq_send_stanza(iq);
+    log_debug("registration stanza has been sent");
     xmpp_stanza_release(iq);
 }
 
--- a/src/xmpp/iq.c
+++ b/src/xmpp/iq.c
@@ -150,6 +150,7 @@ static int _command_list_result_handler(xmpp_stanza_t* const stanza, void* const
 static int _command_exec_response_handler(xmpp_stanza_t* const stanza, void* const userdata);
 static int _mam_rsm_id_handler(xmpp_stanza_t* const stanza, void* const userdata);
 static int _register_change_password_result_id_handler(xmpp_stanza_t* const stanza, void* const userdata);
+static int _register_new_account_result_id_handler(xmpp_stanza_t* const stanza, void* const userdata);
 
 static void _iq_free_room_data(ProfRoomInfoData* roominfo);
 static void _iq_free_affiliation_set(ProfPrivilegeSet* affiliation_set);
@@ -2655,6 +2656,36 @@ _mam_rsm_id_handler(xmpp_stanza_t* const stanza, void* const userdata)
 }
 
 void
+iq_register_new_account(const char* const user, const char* const password)
+{
+    //char* id = connection_create_stanza_id();
+    xmpp_ctx_t* const ctx = connection_get_ctx();
+    xmpp_stanza_t* iq = stanza_register_new_account(ctx, user, password);
+
+    const char* id = xmpp_stanza_get_id(iq);
+    iq_id_handler_add(id, _register_new_account_result_id_handler, NULL, NULL);
+
+    iq_send_stanza(iq);
+    xmpp_stanza_release(iq);
+}
+
+static int
+_register_new_account_result_id_handler(xmpp_stanza_t* const stanza, void* const userdata)
+{
+    const char* type = xmpp_stanza_get_type(stanza);
+    if (g_strcmp0(type, "error") == 0) {
+        char* error_message = stanza_get_error_message(stanza);
+        cons_show_error("Server error: %s", error_message);
+        log_debug("Registration error: %s", error_message);
+        free(error_message);
+    } else {
+        cons_show("Registration successful.");
+        log_debug("Registration successful.");
+    }
+    return 0;
+}
+
+void
 iq_register_change_password(const char* const user, const char* const password)
 {
     xmpp_ctx_t* const ctx = connection_get_ctx();
@@ -2807,3 +2838,4 @@ iq_muc_register_nick(const char* const roomjid)
     xmpp_stanza_release(iq);
     xmpp_stanza_release(query);
 }
+
--- a/src/command/cmd_defs.c
+++ b/src/command/cmd_defs.c
@@ -2638,6 +2638,32 @@ static struct cmd_t command_defs[] = {
       CMD_NOEXAMPLES
     },
 
+    { "/register",
+      parse_args, 2, 6, NULL,
+      CMD_NOSUBFUNCS
+      CMD_MAINFUNC(cmd_register)
+      CMD_TAGS(
+              CMD_TAG_CONNECTION)
+      CMD_SYN(
+              "/register <host> <username> [port <port>] [tls force|allow|trust|legacy|disable]")
+      CMD_DESC(
+              "Register an account on a server.")
+      CMD_ARGS(
+              { "<host>", "Server to register account on." },
+              { "<username>", "Username to register with." },
+              { "port <port>", "The port to use if different to the default (5222, or 5223 for SSL)." },
+              { "tls force", "Force TLS connection, and fail if one cannot be established. This is the default behavior." },
+              { "tls allow", "Use TLS for the connection if it is available." },
+              { "tls trust", "Force TLS connection and trust server's certificate." },
+              { "tls legacy", "Use legacy TLS for the connection. This forces TLS just after the TCP connection is established. Use when a server doesn't support STARTTLS." },
+              { "tls disable", "Disable TLS for the connection." })
+      CMD_EXAMPLES(
+              "/register valhalla.edda odin",
+              "/register vanaheimr.edda freyr port 5678",
+              "/register 127.0.0.1 me tls disable",
+              "/register my.xmppserv.er someuser port 5443 tls force")
+    },
+
     // NEXT-COMMAND (search helper)
 };
 
@@ -3008,3 +3034,4 @@ command_mangen(void)
     g_free(header);
     g_list_free(cmds);
 }
+
--- a/src/command/cmd_defs.c
+++ b/src/command/cmd_defs.c
@@ -2645,23 +2645,25 @@ static struct cmd_t command_defs[] = {
       CMD_TAGS(
               CMD_TAG_CONNECTION)
       CMD_SYN(
-              "/register <host> <username> [port <port>] [tls force|allow|trust|legacy|disable]")
+              "/register <username> <server> [port <port>] [tls force|allow|trust|legacy|disable] [auth default|legacy]")
       CMD_DESC(
               "Register an account on a server.")
       CMD_ARGS(
-              { "<host>", "Server to register account on." },
               { "<username>", "Username to register with." },
+              { "<server>", "Server to register account on." },
               { "port <port>", "The port to use if different to the default (5222, or 5223 for SSL)." },
               { "tls force", "Force TLS connection, and fail if one cannot be established. This is the default behavior." },
               { "tls allow", "Use TLS for the connection if it is available." },
-              { "tls trust", "Force TLS connection and trust server's certificate." },
+              { "tls trust", "Force TLS connection and trust the server's certificate." },
               { "tls legacy", "Use legacy TLS for the connection. This forces TLS just after the TCP connection is established. Use when a server doesn't support STARTTLS." },
-              { "tls disable", "Disable TLS for the connection." })
+              { "tls disable", "Disable TLS for the connection." },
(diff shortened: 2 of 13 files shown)


jubalh:
Very good start!
Instead of merging master into your branch you should instead rebase your branch ontop of master: `git rebase master` (while the branch is checked out).

binex-dsk:
git is currently being extremely funny and i have no clue how to fix this

binex-dsk:
nvm i fixed it

jubalh:
Ok great! Are you making good progress?

binex-dsk:
Debugged it for a while but it seems to occur after the event loop runs a couple of times, not just instantaneously, where it then SIGABRTs with an IOT instruction. It's crashing on an `xmpp_run_once` function in `xmpp/connection.c`. I'm 99% sure that it's caused by the raw connection, as that's literally the only thing that's different. Clearly it's failing, because no account is actually registered; but if i remember correctly there was something in logs indicating a request was sent. because of this, it COULD be the requesting itself that's broken, not connect_raw. Maybe it connects successfully, but when it tries to request, it fails, and for whatever reason, that causes it to fail.

well i've currently got some stuff to do, I'll test this further in a little bit. Instead of debugging i'm just gonna screw around with `cmd_register` and see what happens

binex-dsk:
I'll be writing my updates/findings here as i go

so the problem seems to be that by the time it makes the registration request, it's not actually connected yet. I've tried out a couple things but nothing seems to be able to fix that.

However after some testing it seems the SIGABRT was caused by accidentally freeing one of the arguments, so it no longer occurs yay!

however, when attempting to register again, it segfaults... on `xmpp_conn_release(conn.xmpp_conn)` -> `conn_disconnect(conn)` -> `conn->ctx, "xmpp", "Closing socket.")` -> `xmpp_log(ctx, XMPP_LEVEL_DEBUG, area, fmt, ap);` -> `ctx->log->handler(ctx->log->userdata, level, area, buf);`. I have literally no clue why this happens whatsoever.

anyways, I've confirmed that it's definitely not registering because of it connecting AFTER the registration. Confirmable by commenting out the lines related to registration in `cmd_register`, and, of course, logging when it leaves the function:
```
04/07/2021 18:37:08: prof: INF: we are leaving the registration process
...
04/07/2021 18:37:08: prof: DBG: Connection handler: XMPP_CONN_RAW_CONNECT
```

Also, it currently leaves the connection in a `JABBER_RAW_CONNECTED` state (added this new state so that the application doesn't think it's logged in) even after registration is over, meaning commands like `/connect` will fail. This can be fixed once the whole registration before connection thing is fixed. I'll be pushing my changes here shortly, but that's decent enough findings.

binex-dsk:
git is sperging out extremely hard right now wtf

binex-dsk:
so i did manage to put together a hacky "fix" for registration not working. I just copied a bunch of stuff from the libstrophe register example and "adapted" it. However, as i said, it's **really** hacky so i need to find a way to change it into the standard stanza-iq thing whatever it's called.

BUT that's not really possible right now. The fix still works, but for whatever reason, now there's a segfault after registration in `ret = parser_feed(conn->parser, buf, ret);` -> `return XML_Parse(parser->expat, chunk, len, 0);` -> ... -> `parser->stanzacb(parser->stanza, parser->userdata);` -> `handler_fire_stanza(conn, stanza);` -> `hash_add(conn->id_handlers, id, head);` -> `entry = xmpp_alloc(ctx, sizeof(hashentry_t));` -> `return ctx->mem->alloc(size, ctx->mem->userdata);`. Can not figure out what's wrong right now--but i think it has something to do with something being freed, though I'm not sure what.

even weirder, before it DIDN'T segfault the first time, but still segfaulted after attempting to register a second time. I have no idea what change I made during this that caused this new segfault, but I've been trying to fix these 2 segfaults for 2 hours with no luck on either of them, and I'm tired, physically and mentally, and I really need to finish up some work soon. I probably won't be touching this until that work is done.

@jubalh in the meantime, can you do some testing on your own/look through the code to see stuff that might be wrong? That would be very nice and cool because I'm in an extreme amount of pain right now and please and thank you.

src/xmpp/iq.c
```diff
@@ -2654,6 +2655,37 @@ _mam_rsm_id_handler(xmpp_stanza_t* const stanza, void* const userdata)
...
 }
 
+void
+iq_register_new_account(const char* const user, const char* const password)
+{
+    xmpp_ctx_t* const ctx = connection_get_ctx();
+    xmpp_stanza_t* iq = stanza_register_new_account(ctx, user, password);
+
+    const char* id = xmpp_stanza_get_id(iq);
+    iq_id_handler_add(id, _register_new_account_result_id_handler, NULL, NULL); // FIXME: function doesn't seem to ever run?
```
jubalh, line 2665: You have `iq_register_new_account()` outcommented in `cmd_register()` so the function can't run (if I interpret your comment right here).

jubalh:
@binex-dsk I only took a quick look so far. But looks great! You are doing very good!
I left one comment because you don't call the IQ function (yet?).

src/xmpp/iq.c
binex-dsk: Yeah, that was intentional. That comment in the function meant that the `_register_new_account_result_id_handler` function was seemingly never running, indicating it wasn't properly getting the response. I commented out the `iq_register_new_account` to make way for the new stuff I did with the `_register_handler` and all that, which actually registers after connecting like it should.

Left the `iq_register_new_account` there in case the current implementation can be adapted to the standard iq function thing.

Plus, I've verified that it does actually properly register. the current problems that need fixing are those 2 segfaults.

binex-dsk:
hi ok i finished up all that work. SInce then I have done a very minimal amount of testing, but from what I remember figured nothing out

I did have a hunch that it might be something funny with it trying to authenticate or something but I don't actually think it is. Also there were some TLS issues on my server after switching to a new domain (prosody was being incredibly obese), so I wasn't really able to test for a bit I think. However I managed to make prosody lose weight and start working again so I will have to do a bit more testing later

and speaking of which, I will probably be able to start testing and devving stuff this weekend. Had to do some other stuff for school and still need to until Saturday. Anyways yeah i'll do stuff then

I'm *pretty* sure that both segfaults are caused by something being freed before it should, but like i said will need to do more testing. But considering my experience with segfaults yeah, probably is just that again

jubalh:
Sounds good!

binex-dsk:
ok, looking through segfault backtrace, here are some things I noticed:
- In `xmpp_run_once`, `conn->jid` and `conn->lang` appear to be garbled, and `conn->domain` is a nullptr indicating they were freed at some point (`bound_jid` is also a nullptr--probably why the jid is garbled):
  * image
  * Additionally, `conn->parser->{stanza, inner_text}` are `nullptr`s:
  * image
- Later, in `_end_element` (`libstrophe/src/parser_expat.c:212`), the stanza is used:
  * image
- Because the stanza is a `nullptr`, in `_handle_stream_stanza`, when allocating the `stanza` parameter, it ends up creating, uh, this:
  * image
- Then in `handler_fire_stanza`, because the `stanza` is whatever that mess was, when it gets the `id` and `head` variables:
  * image
  * `head` ends up as a `nullptr`, and `id` ends up as a jumbled mess (though oddly similar to `stanza->children->data`):
  * image
- Those 2 variables are passed to `hash_add`, and I think this part is self explanatory:
  * image
  * image
- Obviously, because the key is a jumbled mess, `entry` is NULL, so it runs `xmpp_alloc`...
  * image
  * And, for whatever reason, `ctx->mem` is just a bunch of 0's:
  * image

So with that extremely long and probably really unnecessary rundown on this one segfault, the problems are as I predicted, something being freed. And it appears to be the JID and stanza that get freed, but the stanza (i believe) is causing this segfault. Both need to be fixed though.

binex-dsk:
Actually wait I just found another thing out. In `connection_check_events`, `conn->xmpp_{conn, ctx}` are both `NULL`:

image

The connection appears to disconnect before the segfault:

```
18/07/2021 16:36:19: xmpp: DBG: RECV: </stream:stream>
18/07/2021 16:36:19: xmpp: DBG: Closing socket.
18/07/2021 16:36:19: prof: DBG: Disconnected
<segfault here>
```

And guess what's in `connection_disconnect`?

image

Yep.

So, the issue actually appears to be that we're disconnecting and then receiving some message from the server. However, there's something weird going on. Here, where the requested username already exists on the server, the stanza appears to be this:
```
<iq type='error' id='h0G3Bsiu3Gpweeu708ae77b5b30d2cfae8676057fcd622d98f62f84'><error type='cancel'><conflict xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/><text xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'>The requested username already exists.</text></error></iq>instructions>Choose a username and password for use with this service.</instructions><field type='hidden' var='FORM_TYPE'><value>jabber:iq:register</value></field><field label='Username' type='text-single' var='username'><required/></field><field label='Password' type='text-private' var='password'><required/></field></x></query></iq>```

When registration was successful:
```
<iq type='result' id='pSbEIwUlLDh9fEda7322a25fb4d5a0ec944a4b9e1cab1af6e9c8eb1'/>instructions>Choose a username and password for use with this service.</instructions><username/><password/><x type='form' xmlns='jabber:x:data'><title>Creating a new account</title><instructions>Choose a username and password for use with this service.</instructions><field type='hidden' var='FORM_TYPE'><value>jabber:iq:register</value></field><field label='Username' type='text-single' var='username'><required/></field><field label='Password' type='text-private' var='password'><required/></field></x></query></iq>
```

So... it also appears that the stanza buffer isn't even cleared properly. You see, in both cases, the beginning part was the most recent message; either that an error occured or it succeeded. Then, after, you can see the registration form: `/>instructions>`.

Well from what I can see in all this, the problem appears to be that the event loop is run AFTER the disconnect.

So the connection gets disconnected in `iq_reg2_cb`... and then, immediately after, the `hash_add` function in `handler_fire_stanza` is run. So this means it's all caused by a premature disconnect, NOT something being freed as i thought! Therefore, removing the disconnect solves this.

But then once you type something after registration, it aborts. And, well, this is caused by the program thinking it's connected and authenticated (when it's actually not). As such there needs to be a way to disconnect afterwards... but i dont know how lol

binex-dsk:
I FIXED REGISTRATION LETS FUCKING GOOOOOOOOOOOOOOOOOO

turns out the only issue was using `connection_disconnect()` instead of `xmpp_disconnect`

Anyways now that it actually works i can maybe start focusing on some of the other stuff now

jubalh:
> I FIXED REGISTRATION LETS FUCKING GOOOOOOOOOOOOOOOOOO

Cool!

binex-dsk:
ok after some long (5 seconds) and careful consideration I've decided that forms shouldn't be included in this PR

Forms are used for account data (i.e. email, name, etc) and to my knowledge profanity doesn't support account data forms yet. it should probably be part of a larger PR for implementing that, and forms aren't really in my scope currently. Plus, because it's so big, it'd probably significantly delay the 0.12.0 milestone, and having registration in 0.12.0 would be pretty cool

as such I don't think forms should be in this PR. Also I don't know if simple JID registration is necessary or really even possible without injecting every known form of cancer into the code, might remove that from task list as well

@jubalh If that's fine with you, then I can rebase, push auth param removal, and then you can merge the PR

jubalh:
> ok after some long (5 seconds) and careful consideration I've decided that forms shouldn't be included in this PR

What kind of forms are we talking about here? Because we already have forms support.
See `src/xmpp/form.c`

> If that's fine with you, then I can rebase, push auth param removal and remove excess functions, and then you can merge the PR

Sure! Let's rebase and cleanup. And then I will try to review soonish, maybe also asking other people to review. And then we can merge! Thanks for your work.

binex-dsk:
I was talking about the data forms mentioned in XEP-0077, see examples 8-10. I know forms are already implemented, but to my knowledge, account data forms aren't yet, like email, gender, name, etc, so a separate PR should be made for that and it should also implement it into registration. If it does exist, then I'm still not gonna add it because I'm lazy lol

jubalh:
Hi @binex-dsk! We released 0.11.1 and are now working on 0.12.0 where this functionality could also land.
What is the current state? Are you finished with implementing and everything works and just a review is needed?

binex-dsk:
@jubalh I completely forgot honestly, I'll do that in a little bit. Probably an invalid disconnect, or it failed to recognize the error

PR #1577, opened 2021-07-12
Fix examples in python API doc

raspbeguy:
Probably a refactoring that didn't go well.


jubalh (approved):

jubalh:
Thanks! Good catch!
BTW the title of the PR was much better than the title of the commit message. You should have used the same title there :)
`Fix examples in python API doc` instaed of `Fix typos in functions names` since it was not really typos.

Seems API was changed in 2016 4f7af4e and documentation wasn't adapted.

PR #1583, opened 2021-07-17
Fix memory leak in accounts_get_login_status()

pasis:
`g_key_file_get_string()` returns a newly allocated string and must be freed with g_free(). `accounts_get_login_status()` uses this API internally and also for return value.

Also reduce copy-paste in a separated commit. Please, review this part carefully.

### DIFF
--- a/src/config/accounts.c
+++ b/src/config/accounts.c
@@ -958,6 +958,10 @@ accounts_get_login_status(const char* const account_name)
     if (g_strcmp0(setting, "last") == 0) {
         status = accounts_get_last_status(account_name);
     }
+
+    if (setting) {
+        g_free(setting);
+    }
     return status;
 }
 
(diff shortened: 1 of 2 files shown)


src/config/accounts.c
```diff
@@ -958,6 +958,10 @@ accounts_get_login_status(const char* const account_name)
     if (g_strcmp0(setting, "last") == 0) {
         status = accounts_get_last_status(account_name);
     }
+
```
jubalh, line 961: I don't think we need an if here, right?

pasis: yes, g_free() and free() handle NULL pointers. I just made in the same way as in other places in this file.

jubalh: I usually try to get rid of them when I encounter them :)

jubalh:
Thanks!

PR #1587, opened 2021-07-20
Fix memory leak in chatwin when load history (#1585)

### LINKED ISSUE #1585: Memory leak when open chat window with history
If I open chat window with `/msg <NAME>` that contains history, I see memory leaks. Also in the chat window I see only opponent's messages in the history (however, this may be unrelated).

## Expected Behavior
No memory leaks.

## Current Behavior
```
==9474== 429 bytes in 10 blocks are definitely lost in loss record 3,022 of 3,317
==9474==    at 0x483879F: malloc (vg_replace_malloc.c:307)
==9474==    by 0x68FBB2A: strdup (in /lib64/libc-2.32.so)
==9474==    by 0x142855: log_database_get_previous_chat (database.c:245)
==9474==    by 0x175264: _chatwin_history.part.0 (chatwin.c:517)
==9474==    by 0x175397: _chatwin_history (chatwin.c:516)
==9474==    by 0x175397: chatwin_new (chatwin.c:70)
==9474==    by 0x17A6B7: _cmd_msg_chatwin (cmd_funcs.c:2124)
==9474==    by 0x17ED11: cmd_msg (cmd_funcs.c:2206)
==9474==    by 0x17B0B6: _cmd_execute (cmd_funcs.c:8312)
==9474==    by 0x17AFAD: cmd_process_input (cmd_funcs.c:154)
==9474==    by 0x17AFAD: cmd_process_input (cmd_funcs.c:131)
==9474==    by 0x142BC8: prof_run (profanity.c:119)
==9474==    by 0x13E69C: main (main.c:185)
```

## Possible Solution
No.

1. `/msg <NAME>`

## Environment
Master branch 45fd229c11bed87fe2413785d7b7f571b79225a7.

pasis:
_chatwin_history() reassigns msg->plain without freeing previous string. This leads to memory leak. As a temporary solution, free replaced string.

Fixes #1585.

### DIFF
--- a/src/ui/chatwin.c
+++ b/src/ui/chatwin.c
@@ -519,7 +519,11 @@ _chatwin_history(ProfChatWin* chatwin, const char* const contact_barejid)
 
         while (curr) {
             ProfMessage* msg = curr->data;
+            char *msg_plain = msg->plain;
             msg->plain = plugins_pre_chat_message_display(msg->from_jid->barejid, msg->from_jid->resourcepart, msg->plain);
+            // This is dirty workaround for memory leak. We reassign msg->plain above so have to free previous object
+            // TODO: Make a better solution, for example, pass msg object to the function and it will replace msg->plain properly if needed.
+            free(msg_plain);
             win_print_history((ProfWin*)chatwin, msg);
             curr = g_slist_next(curr);
         }


jubalh:
Ah now i remember!

`Make a better solution, for example, pass msg object to the function and it will replace msg->plain properly if needed.`

Indeed. There was a time when all functions in profanity had a long lits of arguments. I added ProfMessage to get rid of them and group arguments together that made sense. I replaced all the API calls but omitted the plugins code and wanted to deal with that later.

This is the reason why for arguments we still pass all the parameters individually.

Will make a note.

jubalh:
Thanks!

PR #1591, opened 2021-08-19
Fix OMEMO devicelist access model by reconfiguring it

### LINKED ISSUE #1538: [OMEMO] profanity should fix the PEP access model if wrongly configured by another client

paulfariello:
Fix #1538

Handle precondition-not-met returned when inserting our device in the device list. Try to reconfigure it properly and give up if it fails again.

Also add support for missing device list node. It rely on auto-create. We should ensure this feature is enabled and fallback to a proper create with config otherwise.

### DIFF
--- a/src/omemo/omemo.c
+++ b/src/omemo/omemo.c
@@ -386,12 +386,11 @@ omemo_publish_crypto_materials(void)
 
     char* barejid = connection_get_barejid();
 
-    /* Ensure device list is properly configured */
-    omemo_devicelist_configure_and_request();
 
     /* Ensure we get our current device list, and it gets updated with our
      * device_id */
     g_hash_table_insert(omemo_ctx.device_list_handler, strdup(barejid), _handle_own_device_list);
+    omemo_devicelist_request(barejid);
 
     omemo_bundle_publish(true);
 
--- a/src/omemo/omemo.c
+++ b/src/omemo/omemo.c
@@ -386,10 +386,12 @@ omemo_publish_crypto_materials(void)
 
     char* barejid = connection_get_barejid();
 
+    /* Ensure device list is properly configured */
+    omemo_devicelist_configure_and_request();
+
     /* Ensure we get our current device list, and it gets updated with our
      * device_id */
     g_hash_table_insert(omemo_ctx.device_list_handler, strdup(barejid), _handle_own_device_list);
-    omemo_devicelist_request(barejid);
 
     omemo_bundle_publish(true);
 
--- a/src/omemo/omemo.c
+++ b/src/omemo/omemo.c
@@ -386,7 +386,6 @@ omemo_publish_crypto_materials(void)
 
     char* barejid = connection_get_barejid();
 
-
     /* Ensure we get our current device list, and it gets updated with our
      * device_id */
     g_hash_table_insert(omemo_ctx.device_list_handler, strdup(barejid), _handle_own_device_list);
--- a/src/xmpp/omemo.c
+++ b/src/xmpp/omemo.c
@@ -563,7 +563,8 @@ _omemo_devicelist_publish_result(xmpp_stanza_t* const stanza, void* const userda
 
         xmpp_stanza_t *pubsub_error = xmpp_stanza_get_child_by_ns(error, STANZA_NS_PUBSUB_ERROR);
         if (!pubsub_error) {
-            // TODO
+            log_error("[OMEMO] Unknown error while publishing our own device list");
+            cons_show_error("Unknown error while publishing our own device list");
             return 0;
         }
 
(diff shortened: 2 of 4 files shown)


src/omemo/omemo.c
```diff
@@ -386,6 +386,7 @@ omemo_publish_crypto_materials(void)
 
     char* barejid = connection_get_barejid();
 
+
```
jubalh, line 389: ? :)

src/xmpp/omemo.c
```diff
@@ -507,11 +542,105 @@ _omemo_receive_devicelist(xmpp_stanza_t* const stanza, void* const userdata)
...
+
+        xmpp_stanza_t *error = xmpp_stanza_get_child_by_name(stanza, "error");
+        if (!error) {
+            log_error("[OMEMO] Missing error element in device list publication result");
+            return 0;
+        }
+
+        xmpp_stanza_t *pubsub_error = xmpp_stanza_get_child_by_ns(error, STANZA_NS_PUBSUB_ERROR);
+        if (!pubsub_error) {
+            // TODO
```
jubalh, line 566: todo what?

jubalh (changes requested):

jubalh:
Awesome! Thanks for this!

src/xmpp/omemo.c
```diff
@@ -507,11 +542,105 @@ _omemo_receive_devicelist(xmpp_stanza_t* const stanza, void* const userdata)
...
+            return 0;
+        }
+
+        if (g_strcmp0(xmpp_stanza_get_name(pubsub_error), "precondition-not-met") == 0) {
+            static gboolean reconfigured = false;
```
jubalh, line 571: I'm unsure whether this works as intended.

src/xmpp/omemo.c
```diff
@@ -507,11 +542,105 @@ _omemo_receive_devicelist(xmpp_stanza_t* const stanza, void* const userdata)
...
+
+    if (g_strcmp0(type, STANZA_TYPE_ERROR) == 0) {
+        log_error("[OMEMO] cannot configure device list to an open access model: Result error");
+        return 0;
+    }
+
+    log_debug("[OMEMO] node configured");
+
+    // Try to publish
+    char* barejid = connection_get_barejid();
```
jubalh, line 638: barejid needs to be freed. memleak.

src/xmpp/omemo.c
jubalh: we will after the first run always be in `true` but it could be that we switch accounts without restarting profanity. or it could be that we need to republish because while profanity was running another client did something to the devicelist and we need to republish.

jubalh:
> It rely on auto-create. We should ensure this feature is enabled and fallback to a proper create with config otherwise.

I asked around and heard that the auto-create is pretty common. So we probably don't have to bother with adding the fallback.

jubalh (approved):

jubalh:
There is the corner case with the static variable mentioned in the review above.

Instead of using a static variable we maybe could also pass it along the userdata parameter. That's a long line of calls. According to paul:
* _omemo_devicelist_publish_result
*  omemo_devicelist_configure
*  _omemo_devicelist_configure_submit
*  _omemo_devicelist_configure_result 
* omemo_devicelist_request
*  _omemo_receive_devicelist
*  omemo_set_device_list 
* _handle_own_device_list 
*  omemo_devicelist_publish 
*  _omemo_devicelist_publish_result

omemo_devicelist_request and _handle_own_device_list seem to be not so easy.

The corner cases probably will be quite rare anyways.

paulfariello:
The issue with the static boolean could arise in two cases:
  - deconnection from a client that already reconfigured it's own device list. Followed by a connection to another account that also have a misconfigured device list.
  - running client that has already reconfigured it's own device list. With another client misconfiguring the existing devicelist and removing the running client from the list.
Each time the effect would be that profanity won't be able to reconfigure correctly the devicelist.

Issue #1423, opened 2020-09-02
Messages from MUC history show wrong time stamp


mdosch:
When joining a MUC the timestamp for messages retrieved via MUC history is shown in UTC not in local time.

## Expected Behavior

The MUC history messages should be shown in local time as other messages are also.

## Current Behavior

Messages retrieved via MUC history show the time in UTC while messages received while being joined show the local time (UTC+2 in my case).

## Possible Solution

Convert the time for MUC history messages into local time for display.

## Steps to Reproduce (for bugs)

1. Leave the MUC in profanity
2. Write a message in another client at e.g. 08:30
3. Join with profanity
4. See the message sent in 2 with a timestamp of 06:30 (in case of TZ UTC+2)
5. Send a new message → timestamp shows 8:30

## Environment

Debian Bullseye (Testing)

```
rofanity, version 0.9.5dev.master.98b9cf06
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2020 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.14.0-unknown)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.8.5)
GTK icons: Enabled
```

MarcoPolo-PasTonMolo:
I want to try to solve this, where should I look?

DebXWoody:
On Sa, 02.10.2021 10:33:15, MarcoPolo-PasTonMolo wrote:
> I want to try to solve this, where should I look? 

I'm not sure, but I would start to look into

static void                                                                                                              
_handle_chat(xmpp_stanza_t* const stanza, gboolean is_mam, gboolean is_carbon, const char* result_id, GDateTime* timestamp)

and 

static gboolean                                                                                                          
_handle_mam(xmpp_stanza_t* const stanza) 

file: src/xmpp/message.c

MarcoPolo-PasTonMolo:
[...]

Issue #1254, opened 2020-01-16
Profanity shows old messages as new when connecting to some MUC


Misaflo:
## Expected Behavior

Old messages are not shown as new but are in the right place of the timeline.

## Current Behavior

In some MUC (on server running Openfire 4.5.0), after connect old messages are shown as just received with the current date and time (not in Profanity MUC).

## Context

Here some log (`profanity --log=DEBUG`):

    15/01/2020 21:29:11: xmpp: DBG: RECV: <message id="410757451959" to="me@domain.fr/profanity.8G1h" type="groupchat" from="muc@conference.domain.fr/other"><body>allez bonne soirée.</body><stanza-id id="71106678-f2e1-4054-89f7-17394a61b7af" by="muc@conference.domain.fr" xmlns="urn:xmpp:sid:0"/><delay xmlns="urn:xmpp:delay" stamp="2020-01-15T18:21:58.455Z" from="other@domain.fr/resource/></message>

There is: `stamp="2020-01-15T18:21:58.455Z"` (witch is the right hour, 18:21:58).
But in profanity I see the hour of the connection to the MUC (21:29:11).

This bug concern dev version (and does not occur in the stable release 0.7.1).

## Environment

    profanity --version    
    Profanity, version 0.7.1dev.makepkg.44204635
    Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
    Copyright (C) 2019 - 2020 Michael Vetter <jubalh@iodoru.org>.
    License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>
    
    This is free software; you are free to change and redistribute it.
    There is NO WARRANTY, to the extent permitted by law.
    
    Build information:
    XMPP library: libmesode
    Desktop notification support: Enabled
    OTR support: Enabled (libotr 4.1.1)
    PGP support: Enabled (libgpgme 1.13.1)
    OMEMO support: Enabled
    C plugins: Enabled
    Python plugins: Enabled (3.8.1)
    GTK icons: Enabled

Tested on Arch Linux and Debian Buster.

jubalh:
@Misaflo can  you please test with latest master?

Misaflo:
Yes, no change: the datetime is still wrong (datetime of the connection).

jubalh:
@Misaflo can you test with latest master again?

Misaflo:
No change :-(

jubalh:
[...]

PR #1600, opened 2021-10-03
Fix/history timestamp

MarcoPolo-PasTonMolo:
The fix was to get the newest stanza delay instead of the oldest. This might reintroduce #1254 
Fixes: #1423 

Edit:
Oldest stanza is fetched like before but now the timestamp is generated by creating date_time_from_iso8601 and then to_local, instead of trusting that the timestamp is in utc.
The procedure previously was stamp = time_val_from_iso8601() then date_time = date_time_from_utc(stamp) then local_date_time = to_local(date_time)

### DIFF
--- a/src/xmpp/stanza.c
+++ b/src/xmpp/stanza.c
@@ -1180,9 +1180,9 @@ _stanza_get_delay_timestamp_xep0203(xmpp_stanza_t* const delay_stanza)
 
         if (stamp && (g_time_val_from_iso8601(stamp, &utc_stamp))) {
 
-            GDateTime* utc_datetime = g_date_time_new_from_timeval_utc(&utc_stamp);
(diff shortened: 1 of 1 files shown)


weiss:
> The fix was to get the newest stanza delay instead of the oldest.

The fix to what problem exactly? In what situation is Profanity displaying a wrong timestamp? (Is it easily reproducible?)

weiss:
I mean you're referencing #1423, but that's just about timezones, not about picking the 'wrong' delay tag, no?

jubalh (approved):

jubalh:
Thanks!
This looks good.
If you do a commit in a PR/branch and then revert them later on the same PR/branch please remove the commit entirely. This is done with `git rebase` and then a `drop` instead of `pick`.
I did this for you now and pushed the commit here.

PR #1604, opened 2021-10-16
Add (experimental) theme snikket.

mdosch:
This theme is not an official snikket theme. I just had
some fun creating an orange and blue based theme and
call it snikket.


jubalh:
Thanks!

PR #1616, opened 2021-11-18
gcc-12 fixes

trofi:


jubalh:
Hmm I fail to see how this is a fix. It might be removing code not needed for gcc12. But what about about other compilers? I would think that not all compilers check this themselves and it's not part of any standard.
gcc12 will probably optimize this out anyways, so I don't see any benefit from removing the check.

Am I missing something here?

trofi:
> Hmm I fail to see how this is a fix. It might be removing code not needed for gcc12. But what about about other compilers? I would think that not all compilers check this themselves and it's not part of any standard. gcc12 will probably optimize this out anyways, so I don't see any benefit from removing the check.
> 
> Am I missing something here?

Caveat: I am not familiar with the code base and not familiar with zero-sized C VLAs.

It's probably worth setting a base on what the problem is. `profanity` build fails on `gcc-12` with the following error:

```
$ LANGUAGE=en LANG=C make
make  all-am
make[1]: Entering directory '/home/slyfox/dev/git/profanity'
depbase=`echo src/plugins/python_api.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
gcc -DHAVE_CONFIG_H -I. -I./src  -I/nix/store/wavh03b88r7c5w36kv1hnq1fdp000din-gpgme-1.16.0-dev/include -I/nix/store/0a1kwlava8awba8xl7zivd166h33j3k2-libassuan-2.5.5-dev/include -I/nix/store/lpj7fa0k4679hbl97cd60g7f69vpc1w1-libgpg-error-1.42-dev/include  -D_DEFAULT_SOURCE -D_XOPEN_SOURCE=600 -I/nix/store/qi4amvp612pliyc3cm5rqz2c5ghdh192-ncurses-6.3-dev/include -I/nix/store/62wi0agjl0jfnakfzsxrv6f1h6y6bjnr-glib-2.70.1-dev/include/glib-2.0 -I/nix/store/afshry47b0gb3mnkvyqzcn13dxbwma59-glib-2.70.1/lib/glib-2.0/include -I/nix/store/62wi0agjl0jfnakfzsxrv6f1h6y6bjnr-glib-2.70.1-dev/include -I/nix/store/62wi0agjl0jfnakfzsxrv6f1h6y6bjnr-glib-2.70.1-dev/include/glib-2.0 -I/nix/store/afshry47b0gb3mnkvyqzcn13dxbwma59-glib-2.70.1/lib/glib-2.0/include -I/nix/store/9lh0d4fm3vbrvm126m4hrp39pkcgnp6m-curl-7.79.1-dev/include -I/nix/store/62wi0agjl0jfnakfzsxrv6f1h6y6bjnr-glib-2.70.1-dev/include -I/nix/store/62wi0agjl0jfnakfzsxrv6f1h6y6bjnr-glib-2.70.1-dev/include/glib-2.0 -I/nix/store/afshry47b0gb3mnkvyqzcn13dxbwma59-glib-2.70.1/lib/glib-2.0/include -I/nix/store/8kga2609pmkiqbmvcaaxbx65b3ywwlqh-libnotify-0.7.9-dev/include -I/nix/store/d2ckfk2ry9avx5vr641nh62v75p2f0xv-gdk-pixbuf-2.42.6-dev/include/gdk-pixbuf-2.0 -I/nix/store/7d4amhp7ndqy0wxb84d9j3khw01y234g-python3-3.9.6/include/python3.9 -I/nix/store/62wi0agjl0jfnakfzsxrv6f1h6y6bjnr-glib-2.70.1-dev/include -I/nix/store/62wi0agjl0jfnakfzsxrv6f1h6y6bjnr-glib-2.70.1-dev/include/glib-2.0 -I/nix/store/afshry47b0gb3mnkvyqzcn13dxbwma59-glib-2.70.1/lib/glib-2.0/include -I/nix/store/d2ckfk2ry9avx5vr641nh62v75p2f0xv-gdk-pixbuf-2.42.6-dev/include/gdk-pixbuf-2.0 -I/nix/store/qi765xc809jvxzhx8rn1gw4sp7lgkqh0-gtk+3-3.24.30-dev/include/gtk-3.0 -I/nix/store/ln3z8vrgjgav2x27wk5w3m9jhvc7ys91-atk-2.36.0-dev/include/atk-1.0 -I/nix/store/bywcsawd5ia5cnl9qxf2xgs5fscbsscx-cairo-1.16.0-dev/include/cairo -I/nix/store/fbgycrg54lwzl3jqqdjmdxgpg77v0gnx-freetype-2.11.0-dev/include/freetype2 -I/nix/store/fbgycrg54lwzl3jqqdjmdxgpg77v0gnx-freetype-2.11.0-dev/include -I/nix/store/k1mbb9d8rv40wg90fin6dzplvw1xvhdx-pango-1.48.10-dev/include/pango-1.0 -I/nix/store/m0gaamw205hcwjcrzclx1vavafcv67qq-harfbuzz-3.0.0-dev/include/harfbuzz -I/nix/store/nbimy3y3ln1xqw9hz69hzr3qxnf2c5xq-sqlite-3.36.0-dev/include -DTHEMES_PATH="\"/usr/local/share/profanity/themes\"" -DICONS_PATH="\"/usr/local/share/profanity/icons\"" -DGLOBAL_PYTHON_PLUGINS_PATH="\"/usr/local/share/profanity/plugins\"" -DGLOBAL_C_PLUGINS_PATH="\"/usr/local/lib/profanity/plugins\""  -Wall -Wno-deprecated-declarations -std=gnu99 -Wunused -Werror -I./src -g -O2 -pthread -I/nix/store/b9y45879v60xkasffriz32ykch8jq7vd-libstrophe-0.11.0/include -I/nix/store/qczsiwm600jrpbx0nq0sw6s6qmqwvaxs-libotr-4.1.1-dev/include -MT src/plugins/python_api.o -MD -MP -MF $depbase.Tpo -c -o src/plugins/python_api.o src/plugins/python_api.c &&\
mv -f $depbase.Tpo $depbase.Po
src/plugins/python_api.c: In function ‘python_api_register_command’:
src/plugins/python_api.c:199:31: error: the comparison will always evaluate as ‘true’ for the address of ‘c_arguments’ will never be NULL [-Werror=address]
  199 |         while (c_arguments[i] != NULL && c_arguments[i][0] != NULL) {
      |                               ^~
src/plugins/python_api.c:161:15: note: ‘c_arguments’ declared here
  161 |         char* c_arguments[args_len == 0 ? 0 : args_len + 1][2];
      |               ^~~~~~~~~~~
cc1: all warnings being treated as errors
make[1]: *** [Makefile:1985: src/plugins/python_api.o] Error 1
make[1]: Leaving directory '/home/slyfox/dev/git/profanity'
make: *** [Makefile:1251: all] Error 2
```

To clarify just in case: it's not a cosmetic warning. It's a build failure. I think it's worth fixing. Thus the above changes are at least "the fix" in this regard. `-Werror ` turns all compiler warnings into errors. Compiler generates warnings for standard-compliant code that could be helpful to user. If warnings are not helpful they should probably disabled explicitly or `-Werror` should not be used.

In this case I think the warning is helpful and probably correct: it tells us that code as written probably does not do what it's expected to do. The commits above should not change code's current behaviour.

If we pick `src/plugins/python_api.c` as an example:

```c
src/plugins/python_api.c:199:31: error: the comparison will always evaluate as ‘true’ for the address of ‘c_arguments’ will never be NULL [-Werror=address]
  199 |         while (c_arguments[i] != NULL && c_arguments[i][0] != NULL) {
      |                               ^~
src/plugins/python_api.c:161:15: note: ‘c_arguments’ declared here
  161 |         char* c_arguments[args_len == 0 ? 0 : args_len + 1][2];
      |               ^~~~~~~~~~~
```

`c_arguments` is a 2-dimensional C array on stack. Which should be a syntactic sugar to single-dimensional arrays with nicer arithmetics.

I think `c_arguments[i]` can't be `NULL` because it's still a pointer within `c_arguments`. Or could it somehow be something outside? Why removal is not a correct fix here?

Or does this code rely on the fact that `char* c_arguments[0][2];` could be legitimately dereferenced as `c_arguments[0]` and be `NULL`?

sjaeckel:
I'd say this code is broken in multiple ways...

In case where `args_len == 0`
1. this is UB as the C standard says

> 6.7.6.2 Array declarators
> ...
> 5 If the size [in an array declarator] is an expression that is not an integer constant expression: [...] each time it is evaluated it shall have a value greater than zero.

2. also it writes to unallocated memory on the stack when it does the following

```c
        c_arguments[args_len][0] = NULL;
        c_arguments[args_len][1] = NULL;
```
as even de-referencing `c_arguments[0]` requires the array to have one element

-> The array should always be declared as `char* c_arguments[args_len + 1][2];`


Also the code after the patch is equivalent to the code before the patch

```
$ cat foo.c
#include <stdio.h>
int main() { 
        char* foo[1][2] = {0};
        foo[0][0] = 1;
        foo[0][1] = 2;
        printf("%p\n%p\n%p\n%p\n", foo, foo[0], &foo[0][0], foo[0][0], &foo[0][1], foo[0][1]);
        return 0; 
}
$ gcc -w foo.c -o foo && ./foo 
0x7fff6bacf010
0x7fff6bacf010
0x7fff6bacf010
0x1
0x7fff6bacf018
0x2
```
[^1]
```c
while (c_arguments[i] != NULL && c_arguments[i][0] != NULL)
// c_arguments[i] != NULL - this check is unnecessary as correctly analyzed by gcc
//                          since the array is allocated on the stack and can therefor never be NULL.
//                          In theory it also doesn't hurt, but it will always evaluate to `true`.
// c_arguments[i][0] != NULL - this checks correctly whether the stored pointer is non-NULL
```

@jubalh now the question is:
> But what about about other compilers? I would think that not all compilers check this themselves and it's not part of any standard.

What do you mean by 'this' in "not all compilers check this themselves" which is not part of any standard?

Or am I missing something here?

[^1]: yeah I know, the code spits warnings but that's why I pass `-w` :P

jubalh:
Hi guys, will reply when I have more time (and read more carefully).

Quick info: the plugin code was unmaintained for a long time. No developer almost touched it/looked at it since many releases. Personally i don't use plugins so i never took a look at the code.
Only for 0.12.0 i thought it would be time to actually take a look and in case clean it up.

So this PR comes at the perfect time! :D

> What do you mean by 'this' in "not all compilers check this themselves" which is not part of any standard?

I meant checking the first dimension before trying to access the 2nd.

sjaeckel:
> I meant checking the first dimension before trying to access the 2nd.

in the case of `python_api.c` it's an array on the stack and in the case of `callbacks.c` it's a fixed size array in a struct -> the check is superfluous in both cases.

This check also won't detect out-of-bounds array access as that's never possible (from the array itself) when using arrays in C.

@trofi could you please add a commit that fixes the array declaration in `python_api.c` as mentioned in my comment above?

trofi:
> @trofi could you please add a commit that fixes the array declaration in `python_api.c` as mentioned in my comment above?

Added as https://github.com/profanity-im/profanity/pull/1616/commits/f0a39a4b660cc27d40288216cb9e5a8611109c56

jubalh:
Thanks guys!

sjaeckel:
IMO that's also a bug that has been fixed

Issue #1614, opened 2021-11-08
Can't get OTR to work


updavi:
After upgrading Profanity as part of the upgrade from OpenBSD 6.9 tp 7.0, OTR chat has stopped working. It's no longer possible to start an encrypted session, either manually with `/otr start` or automatically with `/otr policy always`.

Reviewing the `profanity.log` file, I see the following error when I try to start an OTR session:
```
WRN: Invalid carbon received, from: my@contact.name/XMPP
```
This happens with contacts who I previously chatted over OTR with, using Profanity 0.10.0.

This happens for all my XMPP accounts in Profanity, not just one,

## Expected Behavior

OTR session starts

## Current Behavior

Nothing happens in the UI, no OTR session starts. Message about "invalid carbon" is logged to profanity.log

## Possible Solution

No obvious work arounds

## Steps to Reproduce (for bugs)

1. On OpenBSD, `pkg_add profanity` and configure the client
2. Attempt to start an OTR session someone whose client supports OTR (and ideally where you previously commuicated with them using OTR)

## Context

OTR encrypted communication with Profanity is no longer possible

## Environment
* Libotr 4.1.1
* Profanity 0.11.0
* XMPP library libstrophe
* OpenBSD 7.0

jubalh:
I doubt there changed anything regarding the release.

OTR is just brittle and it's hard for people to use, which is why so many clients removed OTR support.

I still like and use OTR. It works well for me.

I'm not entirely sure what breaks OTR. But basically it's designed for a one client to one client world. Multiclient support was not intended.

So please test what happens if you are online with only one client, and your partner too. Does it work then?

It could be that one party goes online with another client, maybe with a higher prio, and then we try to send to this client, breaking OTR.

See setting `/resource`.

jubalh:
BTW https://github.com/profanity-im/profanity/releases/tag/0.11.1 is the latest release.

updavi:
> BTW https://github.com/profanity-im/profanity/releases/tag/0.11.1 is the latest release.

I'm using OpenBSD 7, which packages 0.11.0

My conversation partners are all using one client only, and I am also only using one client.

I can't downgrade to 0.10.0 because that is not available on my system.

`/resource` shows that all display options are turned on, but I don't see anything to give me more information.

jubalh:
> I'm using OpenBSD 7, which packages 0.11.0

They should update. It's a point release.

> I can't downgrade to 0.10.0 because that is not available on my system.

[...]

PR #1618, opened 2021-11-25
Fix carbons criteria

jubalh:
We came into the carbons checking code when we received `<private
xmlns="urn:xmpp:carbons:2"/>`. Which actually marks a message to _not_
be a carbon.

In this code we also make sure that carbons only come from us.
If not we don't call the message handler code.

So we should actually only check for `<sent xmlns='urn:xmpp:carbons:2'>`
and `<received xmlns='urn:xmpp:carbons:2'>`.

Thanks pukkamustard and Holger.

Fixes https://github.com/profanity-im/profanity/issues/1614


pukkamustard:
Seems to do the trick. Thanks for the quick fix!

PR #1619, opened 2021-11-30
Add meson build system support

jubalh:
It's not yet decided whether we will switch to meson or support both autotools and meson.
Even when switching we will support both systems for some releases so that packagers have time to adjust and we can slowly iron out the meson build files.

Run:
`meson build` or `meson setup --wipe build`
Followed by `ninja -C build` or `cd build && ninja`.

Take a look at meson_options.txt for the various flags. Enable/disable them with `meson build -Domemo=disabled`.


meson.build
```diff
@@ -0,0 +1,117 @@
...
+  conf.set_quoted('HAVE_LIBNOTIFY', '1')
+endif
+# TODO: check for otr3 and otr4
+otr_dep = dependency('libotr', version: '>= 4.0', required : get_option('otr'))
+if otr_dep.found()
+  conf.set_quoted('BUILD_OTR', '1')
+endif
+pgp_dep = dependency('gpgme', required : get_option('pgp'))
+if pgp_dep.found()
+  conf.set_quoted('BUILD_PGP', '1')
```
eli-schwartz, line 60: "notifications" is a feature object. When configured with `-Dnotifications=disabled` it logs:

```
Dependency libnotify skipped: feature notifications disabled
[...]
    notifications      : disabled
```
and builds without support.

gpg is a boolean option, which IMHO it shouldn't be. When configured with `-Dpgp=false` it logs:

```
Run-time dependency gpgme found: YES 1.16.0
[...]
    pgp                : false
```

which is contradictory and misleading, because it was available and detected, just not *required*. So it tries to build with PGP support.

BUILD_PGP is the Makefile.am define, not the C define. This should set HAVE_LIBGPGME.

Later, in `core_sources_c`, check if pgp_dep.found() and add 'src/pgp/gpg.c' as an additional source.

meson.build
```diff
@@ -0,0 +1,117 @@
...
+
+# optional dependencies
+notify_dep = dependency('libnotify', required : get_option('notifications'))
+if notify_dep.found()
+  conf.set_quoted('HAVE_LIBNOTIFY', '1')
+endif
+# TODO: check for otr3 and otr4
+otr_dep = dependency('libotr', version: '>= 4.0', required : get_option('otr'))
+if otr_dep.found()
+  conf.set_quoted('BUILD_OTR', '1')
```
eli-schwartz, line 56: Same logic here, I guess.

meson.build
```diff
@@ -0,0 +1,117 @@
...
+thread_dep = dependency('readline')
+m_dep = cc.find_library('m', required : true)
+
+
+# this will detect ncursesw etc (-I / -l flags)
+curses_dep = dependency('curses', required: false)
+# get ncurses definitions
+check_headers = [
+  ['ncursesw/menu.h', 'HAVE_NCURSESW_MENU_H'],
+  ['ncurses/menu.h', 'HAVE_NCURSES_MENU_H'],
```
eli-schwartz, line 28: You can use `'ncursesw/menu.h'.underscorify().to_upper()` later on in conf.set(), so you only list each header once rather than twice.

src/meson.build
```diff
@@ -0,0 +1,95 @@
+# main sources
+core_sources_c = [ 'src/main.c',
+                   'src/profanity.c',
```
eli-schwartz, line 3: Instead of making an array of strings, and including the current directory (`src/`) you can use:

```meson
core_sources_c = files(
    'main.c',
    'profanity.c',
    ...
)
```

The returned File object automatically includes the leading `src/` for you.

src/meson.build
```diff
@@ -0,0 +1,95 @@
...
+# optional sources
+
+if pgp_dep.found()
+  sources += 'src/pgp/gpg.c'
+endif
+
+otr_internal_dep = declare_dependency()
+if otr_dep.found()
+  otr_sources = [ 'src/otr/otrlibv4.c',
+                  'src/otr/otr.c',
```
eli-schwartz, line 89:
```meson
   otr_sources = files('otr/otrlibv4.c',
                       'otr/otr.c',
  )
```

meson.build
jubalh: So basically in our case all the options in meson_options.txt should be `type: feature` right?

src/meson.build
jubalh: what about the files that also belong to the core sources like `src/xmpp/contact.c`? Can I add them via `xmpp/contact.c` or will I need a src/xmpp/meson.build file for this?

meson.build
eli-schwartz: Not entirely.

Feature options are tristate options, booleans are dual-state.
- When used in `if get_option('foobar')` then booleans are on/off switches.
- When used as `required: get_option('foobar')`, then booleans act like "enabled"/"auto" switches. In this case there is no off (disabled) switch.

If you aren't looking up a dependency, just enabling parts of the build, then booleans are correct. For example, c-plugins is either on or off, it has no dependencies so "auto" makes no sense.

You can also do booleans like this:

```meson
if get_option('pgp')
    pgp_dep = dependency('gpgme')
else
    # dummy always-not-found dependency
    pgp_dep = dependency('', required: false)
endif
```

This would be either enabled or disabled, with no "auto" option. But auto is nice.

src/meson.build
eli-schwartz: The former.

```meson
core_sources_c = files(
    'main.c',
    'profanity.c',
    'xmpp/contact.c',
    'event/server_events.c',
    'ui/window.c',
    'command/cmd_defs.c',
    'config/accounts.c',
    ...
)
```

meson.build
jubalh: Thanks for pointing to https://git.sr.ht/~eschwartz/pacman/tree/master/item/meson.build#L123 I think for now I like it like it is :)

meson.build
```diff
@@ -0,0 +1,118 @@
...
+otr_dep = dependency('libotr', version: '>= 4.0', required : get_option('otr'))
+if otr_dep.found()
+  conf.set_quoted('BUILD_OTR', '1')
+endif
+pgp_dep = dependency('gpgme', required : get_option('pgp'))
+if pgp_dep.found()
+  conf.set_quoted('HAVE_LIBGPGME', '1')
+endif
+# TODO: HAVE_LIBSIGNAL_LT_2_3_2
+omemo_libsignal_dep = dependency('libsignal-protocol-c', version: '>= 2.3.2', required : get_option('pgp'))
```
eli-schwartz:
```meson
omemo_libsignal_dep = dependency('libsignal-protocol-c', required : get_option('pgp'))
if omemo_libsignal_dep.version().version_compare('< 2.3.2')
    conf.set_quoted('HAVE_LIBSIGNAL_LT_2_3_2', 1)
endif
```

meson.build
```diff
@@ -0,0 +1,118 @@
...
+omemo_gcrypt_dep = dependency('libsignal-protocol-c', version: '>= 1.7.0', required : get_option('pgp'))
```
eli-schwartz: Typo, should be gcrypt not libsignal-protocol-c.

jubalh: Thanks!

jubalh:
After some internal discussions we decided to stay with autotools for now.
Instead of working on meson support we cleaned up our build files in https://github.com/profanity-im/profanity/pull/1641.

PR #1622, opened 2021-12-22
use wildcard for man pages in Makefile.am

a02c1175-5220-4e75-b7a1-18e20548305f:
use wildcard for man pages in Makefile.am

otherwise it did not build on Debian, see https://salsa.debian.org/xmpp-team/profanity/-/blob/debian/master/debian/patches/fix-make-globbing.patch


jubalh:
Debian maintainers are the best! ;)

jubalh:
Thanks!

PR #1623, opened 2021-12-24
Upgrade to OpenBSD 7.0 in CI

wstrm:
GCC is disabled in 7.0, so instead we install it explicitly and point the CC variable to egcc (which is the binary of the gcc package installed with pkg_add).


jubalh (approved):

jubalh:
Thank you!

PR #1625, opened 2021-12-31
database.c: fix inserting messages to chat logs if archive_id is empty

### LINKED ISSUE #1589: Chat history lost after closing window
## Expected Behavior
When closing a chat window and re-opening one, the entire chat history should be displayed when `history=true` is set in the configuration file.

## Current Behavior
During a chat, messages are displayed in the window. Once the window is closed and that I re-open the same one, chat history is blank. `history=true` is set in the configuration file.

## Possible Solution
Chat is logged, as per the configuration file, in `.local/share/profanity/chatlogs/`. However, the SQLite database is not updated.
Closing Profanity, deleting the database in `.local/share/profanity/database/` and re-starting Profanity creates a new empty one. Then I initiate a chat with few messages. The first one is stored in the database, but not the following ones. If I close and re-open the window, only the first message (the one saved in the database) is displayed in the chat window history.

So, this is probably an issue with updating the database. It started with Profanity 0.11. The database from 0.10 worked and those messages are still displayed in the chat history.

1. Start a new conversation with `/msg nick`
2. Send and/or receive a few messages
3. Close this chat window with `/close`
4. Re-initiate a conversation with the same person `/msg nick`

## Context
It is impossible to keep a history of any conversations. 

## Environment
* Operating System/Distribution: Parabola GNU/Linux
* glib version: 2.33
* Window Manager: sway 1.6.1
* Terminal Emulator: termite 15

`profanity -v`

Profanity, version 0.11.0

Build information:
XMPP library: libmesode
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.9.6)
GTK icons/clipboard: Disabled

JurajMlich:
The original intention of the code was that in case archive_id is not set, NULL should be inserted. What is
inserted however is an empty string. This causes the condition to not insert messages with non-unique
archive_id insert only one message in total and ignore all further ones (if NULL was there, the condition
would work properly). And this in turn causes chat history not work properly.

This commit makes the SQL condition work properly and therefore fixes chat history.

Fixes #1589.

See https://github.com/profanity-im/profanity/issues/1589#issuecomment-1002290945 for longer explanation.

### DIFF
--- a/src/database.c
+++ b/src/database.c
@@ -328,7 +328,7 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
         type = (char*)_get_message_type_str(message->type);
     }
 
-    query = sqlite3_mprintf("INSERT INTO `ChatLogs` (`from_jid`, `from_resource`, `to_jid`, `to_resource`, `message`, `timestamp`, `stanza_id`, `archive_id`, `replace_id`, `type`, `encryption`) SELECT '%q', '%q', '%q', '%q', '%q', '%q', '%q', '%q', '%q', '%q', '%q' WHERE NOT EXISTS (SELECT 1 FROM `ChatLogs` WHERE `archive_id` = '%q')",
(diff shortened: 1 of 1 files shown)


jubalh (approved):

jubalh:
Thanks a lot @JurajMlich !

PR #1631, opened 2022-02-01
Fix #1628

### LINKED ISSUE #1628: Implement see-other-host during login
When connecting to sure.im, about 3/4 times results in "Login Failed". Here is the log:
```
09/01/2022 22:05:24: tls: DBG: Subject=/CN=tigase.im
09/01/2022 22:05:24: tls: DBG: Issuer=/C=US/O=Let's Encrypt/CN=R3
09/01/2022 22:05:24: conn: DBG: QUEUED: <?xml version="1.0"?><stream:stream to="sure.im" xml:lang="en" version="1.0" xmlns="jabber:client" xmlns:stream="http://etherx.jabber.org/streams" from="rdg@sure.im">
09/01/2022 22:05:24: conn: DBG: Added queue element: 0x5608de2ba290
09/01/2022 22:05:24: conn: DBG: SENT: <?xml version="1.0"?><stream:stream to="sure.im" xml:lang="en" version="1.0" xmlns="jabber:client" xmlns:stream="http://etherx.jabber.org/streams" from="rdg@sure.im">
09/01/2022 22:05:24: xmpp: DBG: Finished writing queue element: 0x5608de2ba290.
09/01/2022 22:05:25: xmpp: DBG: RECV: <stream:stream id="tigase-error-tigase" from="sure.im" version="1.0" lang="en">
09/01/2022 22:05:25: xmpp: DBG: RECV: <error xmlns="http://etherx.jabber.org/streams"><see-other-host xmlns="urn:ietf:params:xml:ns:xmpp-streams">ec2-54-201-227-188.us-west-2.compute.amazonaws.com</see-other-host></error>
09/01/2022 22:05:25: xmpp: DBG: RECV: </stream:stream>
09/01/2022 22:05:25: xmpp: DBG: Closing socket.
09/01/2022 22:05:25: prof: DBG: Connection handler: XMPP_CONN_DISCONNECT
09/01/2022 22:05:25: prof: DBG: Connection handler: Login failed
09/01/2022 22:05:25: prof: DBG: Connection handler: No reconnect timer
09/01/2022 22:05:25: prof: INF: Login failed
```

As per the RFC, it should reconnect to the given server:
> the initiating entity SHOULD simply attempt to reconnect at that address

https://xmpp.org/rfcs/rfc6120.html#streams-error-conditions-see-other-host

sjaeckel:
This PR fixes #1628 

As soon as we get a `see-other-host` error from libstrophe, unpack the new URL and try a reconnect.

Maybe ToDo: This will do an unlimited number of reconnects whereas `rfc6120` recommends to stop at 5.

An alternative would be to implement this in libstrophe, but this was easier for now. Maybe it will be done in libstrophe at a later point of time.

### DIFF
--- a/src/xmpp/connection.c
+++ b/src/xmpp/connection.c
@@ -872,6 +872,70 @@ connection_set_priority(const int priority)
     conn.priority = priority;
 }
 
+#if defined(LIBXMPP_VERSION_MAJOR) && defined(LIBXMPP_VERSION_MINOR) \
+    && ((LIBXMPP_VERSION_MAJOR > 0) || (LIBXMPP_VERSION_MINOR >= 12))
+static xmpp_stanza_t*
+_get_soh_error(xmpp_stanza_t* error_stanza)
+{
+    return xmpp_stanza_get_child_by_path(error_stanza,
+                                         XMPP_STANZA_NAME_IN_NS("error", STANZA_NS_STREAMS),
+                                         XMPP_STANZA_NAME_IN_NS("see-other-host", STANZA_NS_XMPP_STREAMS),
+                                         NULL);
+}
+#else
+static xmpp_stanza_t*
+_get_soh_error(xmpp_stanza_t* error_stanza)
+{
+    const char* name = xmpp_stanza_get_name(error_stanza);
+    const char* ns = xmpp_stanza_get_ns(error_stanza);
+    if (!name || !ns || strcmp(name, "error") || strcmp(ns, STANZA_NS_STREAMS)) {
+        log_debug("_get_soh_error: could not find error stanza");
+        return NULL;
+    }
+    return xmpp_stanza_get_child_by_name_and_ns(error_stanza, "see-other-host", STANZA_NS_XMPP_STREAMS);
+}
+#endif
+
+static bool
+_get_other_host(xmpp_stanza_t* error_stanza, gchar** host, int* port)
+{
+    xmpp_stanza_t* soh_error = _get_soh_error(error_stanza);
+    if (!soh_error || !xmpp_stanza_get_children(soh_error)) {
+        log_debug("_get_other_host: stream-error contains no see-other-host");
+        return false;
+    }
+    const char* alturi = xmpp_stanza_get_text_ptr(xmpp_stanza_get_children(soh_error));
+    if (!alturi) {
+        log_debug("_get_other_host: see-other-host contains no text");
+        return false;
+    }
+    /* Construct a valid URI with `schema://` as `g_uri_split_network()`
+     * requires this to be there.
+     */
+    const char* xmpp = "xmpp://";
+    char* xmpp_uri = malloc(strlen(xmpp) + strlen(alturi) + 1);
+    if (!xmpp_uri) {
+        log_debug("_get_other_host: malloc failed \"%s\"", alturi);
+        return false;
+    }
+    memcpy(xmpp_uri, xmpp, strlen(xmpp));
+    memcpy(xmpp_uri + strlen(xmpp), alturi, strlen(alturi) + 1);
+
+    if (!g_uri_split_network(xmpp_uri, 0, NULL, host, port, NULL)) {
+        log_debug("_get_other_host: Could not split \"%s\"", xmpp_uri);
+        free(xmpp_uri);
+        return false;
+    }
+    free(xmpp_uri);
+    /* fix-up `port` as g_uri_split_network() sets port to `-1` if it's missing
+     * in the passed-in URI, but libstrophe expects a "missing port"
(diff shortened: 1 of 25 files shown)


src/xmpp/connection.c
```diff
@@ -872,6 +872,70 @@ connection_set_priority(const int priority)
     conn.priority = priority;
 }
 
+#if defined(LIBXMPP_VERSION_MAJOR) && defined(LIBXMPP_VERSION_MINOR) \
+    && ((LIBXMPP_VERSION_MAJOR > 0) || (LIBXMPP_VERSION_MINOR >= 12))
+static xmpp_stanza_t*
+_get_soh_error(xmpp_stanza_t* error_stanza)
+{
+    return xmpp_stanza_get_child_by_path(error_stanza,
+                                         XMPP_STANZA_NAME_IN_NS("error", "http://etherx.jabber.org/streams"),
```
jubalh, line 881: `http://etherx.jabber.org/streams` should go into `src/xmpp/stanza.h` as a define

src/xmpp/connection.c
```diff
@@ -872,6 +872,70 @@ connection_set_priority(const int priority)
...
+                                         XMPP_STANZA_NAME_IN_NS("see-other-host", "urn:ietf:params:xml:ns:xmpp-streams"),
```
jubalh, line 882: also define

src/xmpp/connection.c
```diff
@@ -872,6 +872,70 @@ connection_set_priority(const int priority)
...
+static xmpp_stanza_t*
+_get_soh_error(xmpp_stanza_t* error_stanza)
+{
+    const char* name = xmpp_stanza_get_name(error_stanza);
+    const char* ns = xmpp_stanza_get_ns(error_stanza);
+    if (!name || !ns || strcmp(name, "error") || strcmp(ns, "http://etherx.jabber.org/streams")) {
+        log_debug("_get_soh_error: could not find error stanza");
+        return NULL;
+    }
+    return xmpp_stanza_get_child_by_name_and_ns(error_stanza, "see-other-host", "urn:ietf:params:xml:ns:xmpp-streams");
```
jubalh, line 895: define

src/xmpp/connection.c
```diff
@@ -872,6 +872,70 @@ connection_set_priority(const int priority)
...
+                                         XMPP_STANZA_NAME_IN_NS("see-other-host", "urn:ietf:params:xml:ns:xmpp-streams"),
+                                         NULL);
+}
+#else
+static xmpp_stanza_t*
+_get_soh_error(xmpp_stanza_t* error_stanza)
+{
+    const char* name = xmpp_stanza_get_name(error_stanza);
+    const char* ns = xmpp_stanza_get_ns(error_stanza);
+    if (!name || !ns || strcmp(name, "error") || strcmp(ns, "http://etherx.jabber.org/streams")) {
```
jubalh, line 891: define

src/xmpp/connection.c
```diff
@@ -872,6 +872,70 @@ connection_set_priority(const int priority)
...
+     * requires this to be there.
+     */
+    const char* xmpp = "xmpp://";
+    char* xmpp_uri = malloc(strlen(alturi) + strlen(xmpp) + 1);
+    if (!xmpp_uri) {
+        log_debug("_get_other_host: malloc failed \"%s\"", alturi);
+        return false;
+    }
+    memcpy(xmpp_uri, xmpp, strlen(xmpp));
+    memcpy(xmpp_uri + strlen(xmpp), alturi, strlen(alturi));
```
jubalh, line 922: will this include `\0` at the end of the string?

jubalh (changes requested):

jubalh:
Seems like `g_uri_split_network` is only available since glib 2.66.
Debian 11, Fedora 33: 2.66.8
openSUSE Leap 15.4 (soon to be released): 2.70

Seems like all big distros have a latest stable version that would have a new enough glib.
We use `FROM ubuntu:latest` in the Dockerfile so maybe will need to disable Ubuntu CI for a while.

I think it would be okay to bump glib and disable Ubuntu CI for now. If it's too troublesome to rewrite this function.

sjaeckel:
> If it's too troublesome to rewrite this function.

I hope this fall-back won't break too badly ... :grimacing: 

If you prefer to not include this patch, I can also drop the commit again.

jubalh:
Thanks for the patch :)

I think few distros will be affected by this so probably only few users who will run this code path.
Maybe it can make things a little trickier when they create a bug report and don't mention their versions :D 

But surely good to have.
And maybe we even can drop it before the next release depending on when that is.

Good to have CI building!

jubalh:
Thanks for you work! :)

PR #1637, opened 2022-02-08
fix handling of connection errors

sjaeckel:
When a `see-other-host` stream-error is received we try to re-connect to
the other host. Erroneously this also started the `reconnect_timer`.
This lead to the behavior that in cases where e.g. the login failed
we try to reconnect instead of bailing out with an error.

This commit fixes the wrong behavior by not starting the `reconnect_timer`.

Fix 0e58509c161ae8409c9accabb9606e0c7006b880 

Related to #1628 #1631


jubalh:
Thanks!

PR #1640, opened 2022-02-17
Update the contributing guide

jubalh:
We need to define some rules to make sure people know what they need to
take care about.
To allow us to later still understand the changes in an optimal way and
make everyones work easier.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,38 @@
 # Contributing to Profanity
 
-`make doublecheck` will run the code formatter, spell checker and unit tests.
+## Submitting patches
+We recommend for people to always work on a dedicated git branch for each fix or feature.
+Don't work on master.
+So that they can easily pull master and rebase their work if needed.
+
+For fixing (reported) bugs we usually use `git checkout -b fix/issuenumber-somedescription`.
+When working on a new feature we usually use `git checkout -b feature/optionalissuenumber-somedescription`.
+
+However this is not a rule just a recommendation to keep an overview of things.
+If your change isn't a bugfix or new feature you can also just use any branch name.
+
+### GitHub
+We would like to encourage people to use GitHub to create pull requests.
+It makes it easy for us to review the patches, track WIP branches, organize branches with labels and milestones,
+and help others to see what's being worked on.
+
+Also see the blogpost [Contributing a Patch via GitHub](https://profanity-im.github.io/blog/post/contributing-a-patch-via-github/).
+
+### E-Mail
+In case GitHub is down or you can't use it for any other reason, you can send a patch to our [mailing list](https://lists.notraces.net/mailman/listinfo/profanity).
+
+We recommend that you follow the workflow mentioned above.
+And create your patch using the [`git-format-patch`](https://git-scm.com/docs/git-format-patch) tool: `git format-patch master --stdout > feature.patch`
+
+### Rules
+
+* When fixing a bug, describe it and how your patch fixes it.
+* When fixing a reported issue add an `Fixes https://github.com/profanity-im/profanity/issues/23` in the commit body.
+* When adding a new feature add a description of the feature and how it should be used (workflow).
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,6 +1,38 @@
...
+
+### GitHub
+We would like to encourage people to use GitHub to create pull requests.
+It makes it easy for us to review the patches, track WIP branches, organize branches with labels and milestones,
+and help others to see what's being worked on.
+
+Also see the blogpost [Contributing a Patch via GitHub](https://profanity-im.github.io/blog/post/contributing-a-patch-via-github/).
+
+### E-Mail
+In case GitHub is down or you can't use it for any other reason, you can send a patch to your [mailing list](https://lists.notraces.net/mailman/listinfo/profanity).
```
sjaeckel, line 22:
```suggestion
In case GitHub is down or you can't use it for any other reason, you can send a patch to our [mailing list](https://lists.notraces.net/mailman/listinfo/profanity).
```

CONTRIBUTING.md
```diff
@@ -1,6 +1,38 @@
...
+
+We recommend that you follow the workflow mentioned above.
+And create your patch using: `git format-patch master --stdout > feature.patch`.
+
+### Rules
+
+* When fixing a bug, describe it and how your patch fixes it.
+* When fixing a reported issue add an `Fix https://github.com/profanity-im/profanity/issues/23` in the commit body.
```
sjaeckel, line 30: I'm usually using `Fixes #issue` instead of the full URL. Would that be acceptable as well?

CONTRIBUTING.md
```diff
@@ -1,6 +1,38 @@
...
+* When adding a new feature add a description of the feature and how it should be used (workflow).
+* If your patch adds a new configuration option add this to the `profrc.example` file.
+* If your patch adds a new theming option add this to the `theme_template` file.
+* Each patch or pull request should only contain related modifications.
+* Run the tests and code formatters before submitting.
```
sjaeckel, line 35:
```suggestion
* Run the tests and code formatters before submitting (c.f. Chapter 'Check everything' of this README).
```

CONTRIBUTING.md
```diff
@@ -1,6 +1,38 @@
...
+It makes it easy for us to review the patches, track WIP branches, organize branches with labels and milestones,
+and help others to see what's being worked on.
+
+Also see the blogpost [Contributing a Patch via GitHub](https://profanity-im.github.io/blog/post/contributing-a-patch-via-github/).
+
+### E-Mail
+In case GitHub is down or you can't use it for any other reason, you can send a patch to your [mailing list](https://lists.notraces.net/mailman/listinfo/profanity).
+
+We recommend that you follow the workflow mentioned above.
+And create your patch using: `git format-patch master --stdout > feature.patch`.
```
sjaeckel, line 25:
```suggestion
And create your patch using the [`git-format-patch`](https://git-scm.com/docs/git-format-patch) tool.
```

CONTRIBUTING.md
jubalh: I used to use this too. And I like it better because typing less.
My only reason for changing was: what if we ever move away from GitHub indeed?
I think there are tools to help with migration, like copying issues over and so on. But I have never done this nor read about it. I just thought maybe it would be better to include the full URL to be more explicit.
What do you think?

sjaeckel: The only thing I can talk of is a migration to GitLab and they can pull all issues&PR's over and re-create them including all the communication. I'm not so sure anymore whether they are then also numbered the same...

Also we don't know what will be supported by service X when a migration once would happen.

Answering your question ... I'm not sure, I'm more the kind of 'easy way now, adapt later when it's necessary' guy so I wouldn't force the URL. But hey, that's also just my opinion :smile:

jubalh: I see!
Well I prefer the shorter version for readability, but I think I'll go with the URL since it feels more "future proof" to me.

Issue #1638, opened 2022-02-12
Support editor for editing subject


mdosch:
Some chat rooms have a multiline subject and therefore editing those with `/subject edit <tab>` is cumbersome to nearly impossible.
I think using the editor functionly (like in `/editor` or `/correct-editor`) for editing the subject (e.g. with `/subject editor`) would be very nice.

## Environment
* Debian unstable

```
profanity --version
Profanity, version 0.12.0dev.master.6d1c661e
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2021 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.16.0-unknown)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Disabled
GTK icons/clipboard: Disabled
```

PR #1643, opened 2022-02-23
Add `/subject editor` command

jubalh:
Allow editing the MUC subject using external editor.

Use `/subject editor`.

Fix https://github.com/profanity-im/profanity/issues/1638


jubalh:
@mdosch could you test this?

mdosch:
It works but there is room for improvements:

- If you type `/subject ed<tab>` it autocompletes to `/subject edit old 
   subject` and pressing <tab> again has no impact.
- If you manually type `/subject editor` you can create a subject in 
   your editor but the old subject is not shown in the editor so you have 
   to retype everything although you make only a minor edit.

jubalh:
@mdosch Thanks for the review! Here goes another round. Let me know if it works well now!

mdosch:
Looks good now. :)

Issue #1580, opened 2021-07-15
Make Alt+F forward a word again


qorg11:
## Expected Behavior

The input moves to the next word (Like in emacs and most shells)
## Current Behavior

It sets the attention flag
## Possible Solution

Configure keybindings
## Steps to Reproduce (for bugs)

1. Press Alt + F while writing a message with more than 2 words

jubalh:
I totally forgot about usual readline shortcuts since I never use them.
We should make sure not to overwrite the most default ones (that make sense) probably: https://readline.kablamo.org/emacs.html

jubalh:
So what would be good alternatives for the key combos?

alt-f flag window as marked
alt-m - switch between marked windows

alt-a marking as unread

I'm tempted to just let them stay as they are since I guess all the readline combos can also be done with ctrl+arrow keys. And we probably want short key combos for the marking/switching between windows.

qorg11:
Well, if it will be less pain for you, leave them like that. Altough i'd like it to be possible to use Alt+f again

jubalh:
It's not hard to change to code to use another binding.
We just need to have a good idea with letters that make sense and at the same time not use default readline letters :/

Do you have sugestions?

qorg11:
Hm, well I think Alt-n doesn't make sense but it is near m so you can quickly press Alt-n and Alt-m?

jubalh:
What about the others though?

4drift:
I agree with alt+n as it is convenient. I am greatly displeased by having alt+f gone as I use it to jump words ever constantly and ctrl+arrows doesn't even work for me.

Besides it is terribly inconvenient compared to alt+f/b -- I use it in the terminal constantly too. I'd think more people would have a problem with this binding being taken over. Having a memorable letter just isn't reason enough to overwrite a default terminal binding.

jubalh:
> I am greatly displeased by having alt+f gone as I use it to jump words ever constantly

We don't need to know about this. The original poster said this already.

> and ctrl+arrows doesn't even work for me.

Then fix your setup.

>  Having a memorable letter just isn't reason enough to overwrite a default terminal binding.

alt+f is not a "default terminal binding". Actually it has nothing to do with your terminal. It's the default binding for GNU Readline in emacs mode.
If we would have not chosen to use the readline library then this terminal wouldn't work out of the box at all. Like is the case for many applications out there.

[...]

Issue #1521, opened 2021-04-16
Feature: Spawn external editor to edit message


jubalh:
@StefanKropp had this idea and started it in https://github.com/profanity-im/profanity/pull/1485.

What needs to be done:
* [x] Decide what PREF_COMPOSE_EDITOR should return by default `/usr/bin/vim`, `xdg-open`, `$EDITOR`, `$VISUAL`? Right now we use `fork()`/`execl()` we need a whole path. So just `xdg-open` or `$EDITOR` won't work right now. #1546
* [x] Check whether it works with a graphical editor -> https://github.com/profanity-im/profanity/issues/1521#issuecomment-854110973
* [ ] Fix cleanup error when quitting Profanity

Last thing happens when:
Test1:

    start profanity
    /msg someone
    /quit

Test2:

    start profanity
    /msg someone
    write a message using /editor
    /quit

In Test2 the terminal will not be cleared properly. There is still part of profanity visible. Maybe something ncurses related.

* [x] Remove the "EDITOR PREVIEW" from the chatwindow. Or let it stay but replace it later when the actual message have been sent. -> a4deec8b8d26c7bd7312b972372b4ef3aab8eeda
* [ ] Since the inputline is empty it might confuse some people at the start. Let's indicate with something like `[Editor]` in the statusbar that we are in a special state.
* [x] Make it configurable which editor to use. Done in 19c5925c378edc12e383278561e1a2f5e2d1f597
* [x] Save in `.local/share/profanity/editor/barejid.md` instead of `/tmp/profanity-uuid.md`. Advantages: not a place where other users have access. By using barejid we can also reopen this later and only delete it once sent. -> 242381cfd012edbba3f502aaa287aa4a3806dddf
* [x] (Maybe) delete the tmp file only once we sent the message. So that one could run `/editor` several times and amend some text. -> not done

jubalh:
Some ideas where this could lead:

This can later be used to reply to messages too.
Maybe a workflow like his:
* /reply
* each line gets a number
* press the number (3)
* editor gets opened containing ">" + the content of line 3

@kaffeekanne also suggest to use it for correct via `/edit correct`

mdosch:
I'd prefer to have a number per message and quote the full message even if it's multiline.

jubalh:
> I'd prefer to have a number per message and quote the full message even if it's multiline.

That's what I said?

mdosch:
You said 'line' but a message could be several lines.

[...]

PR #1645, opened 2022-02-27
Add hotkey for sending readline text to editor

MarcoPolo-PasTonMolo:
Add a hotkey for sending text in readline to external text editor.
Should deprecate the need for /command-editor requests like /subject-editor, /correct-editor.

Example workflow:
alt+d to open editor or write something in readline first and then alt+d.

I have a feeling a put the declaration of _get_message_from_editor in a suboptimal header file.

### DIFF
--- a/src/tools/editor.c
+++ b/src/tools/editor.c
@@ -1,0 +1,132 @@
+/*
+ * editor.c
+ * vim: expandtab:ts=4:sts=4:sw=4
+ *
+ * Copyright (C) 2022 Michael Vetter <jubalh@iodoru.org>
+ * Copyright (C) 2022 MarcoPolo PasTonMolo  <marcopolopastonmolo@protonmail.com>
+ *
+ * This file is part of Profanity.
+ *
+ * Profanity is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Profanity is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Profanity.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * In addition, as a special exception, the copyright holders give permission to
+ * link the code of portions of this program with the OpenSSL library under
+ * certain conditions as described in each individual source file, and
+ * distribute linked combinations including the two.
+ *
+ * You must obey the GNU General Public License in all respects for all of the
+ * code used other than OpenSSL. If you modify file(s) with this exception, you
+ * may extend this exception to your version of the file(s), but you are not
+ * obligated to do so. If you do not wish to do so, delete this exception
+ * statement from your version. If you delete this exception statement from all
+ * source files in the program, then also delete it here.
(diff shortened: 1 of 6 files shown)


jubalh:
Thanks for this PR! Looks like a good idea!
I'll do a review soon. For now:
* can you tell us in this PR how/what you tested?
*  `_get_message_from_editor`  -> the `_` indicates that it's a local helper function and not exported.

MarcoPolo-PasTonMolo:
1. Just wrote random stuff in readline(including commands) and then pressed alt+d. Seems to work fine.
2. Ok I'll remove the _ in a minute

jubalh:
About testing I meant:
* Did you test with a command that requires no argument
* Requires one argument
* requires several
* commands that maybe not expect a newline? Can we think of something like this?

And I think we should move the `get_message_from_editor()` out of `cmd_funcs.c` maybe into `src/common.c`. I'm not sure what's the best location but probably that's a better one?
(@sjaeckel any opinion on this?)

sjaeckel (approved):
I tried it and it works well

I also think we should move `get_message_from_editor()` to another file.

jubalh:
We need to be careful with these hotkeys. In the past we already used some that are actually used by readline per default too.

Someone collected them here: https://github.com/profanity-im/profanity/issues/1580#issuecomment-908866258

sjaeckel (commented):
I'm not sure where this comes from but if I already have text written, then press `Alt+d` and then immediately close the editor, the last character is lost!

jubalh:
Please move `get_message_from_editor()` to `src/tools/editor.{c,h}`.

sjaeckel:
> I'm not sure where this comes from but if I already have text written, then press `Alt+d` and then immediately close the editor, the last character is lost!

I've a preliminary patch for this issue and will open a separate PR

MarcoPolo-PasTonMolo:
> So alt + d is on there. I was curious anyways why you wanted to use d for editor? :-)
It was a mistake, wanted to have e for editor but accidentaly typed d and it stuck. Should I change it?

> Please move get_message_from_editor() to src/tools/editor.{c,h}.
Will do!

> I've a preliminary patch for this issue and will open a separate PR
Ok

MarcoPolo-PasTonMolo:
Minor issue: what should I put after Copyright?

I saw that other files in that (tools) folder include a comment about the GNU license and someone's name above it
>  Copyright (C) 2012 - 2019 James Booth <boothj5@gmail.com>

MarcoPolo-PasTonMolo:
I noticed some memory gets lost when the get_message_from editor command gets called, as I understand it's the returned_message that gets generated from g_strdup, I tried calling gfree on it in inputwin.c but it still gets lost

src/tools/editor.c
```diff
@@ -110,13 +111,16 @@ get_message_from_editor(gchar* message, gchar** returned_message)
             GString* text = g_string_new(buf);
             *returned_message = g_strdup(text->str);
             g_string_free(text, TRUE);
+        } else {
+            *returned_message = g_strdup("");
```
jubalh, line 115: thats also a solution but i would prefer another one. but lets take it for now!

jubalh (approved):

jubalh:
@MarcoPolo-PasTonMolo quite a cool feature! Thanks for the idea and implementation!

I think this will be a game changer for me like alt+a :D

sjaeckel:
This relates to #1521
... and maybe that's what @hund wanted?

Hund:
> This relates to #1521 ... and maybe that's what @Hund wanted?

I'm humbled that you though of me. :) This looks like exactly what I was looking for!

Issue #1474, opened 2021-01-28
Cycle through Nicknames breaks on slashes


kaffeekanne:
When cycling through nicknames with tab in a MUC, it stops when there is a nickname containing slash `/`.

## Expected Behavior
Cycle through all nicknames.

## Current Behavior
Cycles until the nick starting with `/` is reached.

## Steps to Reproduce (for bugs)

1. Have a MUC with a participant named `/something`
2. Have an empty input window
3. cycle through nicknames to mention with tab
4. upon reaching nick `/something`, further cycling with tab stops working

## Context

Since i can't type cyrillic letters i tried to tab until that nick is reached to be highlighted.

## Environment
```
Profanity, version 0.10.0
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2021 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Disabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.14.0-unknown)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Disabled
GTK icons: Disabled
```
Raspberry Pi OS (Debian Testing/Bullseye)

jubalh:
@mwuttke97 something for you?

jubalh:
@mwuttke97 if I remember correctly you mentioned in the MUC that this bug should be trivial. Do you think you can tackle it soonish so it will be in 0.11? :)

jubalh:
@xenrox do you think you have time to tackle this one?

PR #1646, opened 2022-03-03
Fix not autocompleting nicks with '/'

MarcoPolo-PasTonMolo:
Remove check for '/' at the beginning of the line before autocompleting
and make it fallback to command autocompletion if no nicks found.

Fixes https://github.com/profanity-im/profanity/issues/1474

### DIFF
--- a/src/ui/inputwin.c
+++ b/src/ui/inputwin.c
@@ -583,14 +583,16 @@ _inp_rl_tab_handler(int count, int key)
     }
 
     ProfWin* current = wins_get_current();
-    if ((strncmp(rl_line_buffer, "/", 1) != 0) && (current->type == WIN_MUC)) {
+    if (current->type == WIN_MUC) {
         char* result = muc_autocomplete(current, rl_line_buffer, FALSE);
         if (result) {
             rl_replace_line(result, 1);
             rl_point = rl_end;
             free(result);
         }
-    } else if (strncmp(rl_line_buffer, "/", 1) == 0) {
+    }
+
+    if (strncmp(rl_line_buffer, "/", 1) == 0) {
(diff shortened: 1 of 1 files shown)


jubalh:
When testing this PR I discovered: https://github.com/profanity-im/profanity/issues/1647

Regarding this PR:
Looks good! And solves the mentioned issue indeed!

However there is one down side to the current implementation. Let's say a user joins with the nickname `/sendfile` then autocompletion will always autocomplete to this nick (`'/sendfile:`) and not do any file completion for `/sendfile ~/myfile` anymore.

So it's a question what behaviour do we want?

My opinion is that another user shouldn't be able to limit the functionality of the client. So the proper  autocompletion profanity commands is more important.

What to do? Not complete nicknames starting with `/` at all?

MarcoPolo-PasTonMolo:
> My opinion is that another user shouldn't be able to limit the functionality of the client. So the proper autocompletion profanity commands is more important.

I agree.

> What to do? Not complete nicknames starting with / at all?

Maybe?

jubalh:
> Maybe?

Yes, let's do it this way :)

MarcoPolo-PasTonMolo:
Do I have to do something? As I understand it just closing this pull request and related issue without merging it would be what we want

jubalh:
Oh? I thought we just turn it around. Like first checking for commands and autocompleting them and otherwise autocompleting the nick.

Which means we will autocomplete all nicks (even the ones starting with `/`) and only leave out the ones that start with `/` and are a profanity command as well.

Am I missing something?

MarcoPolo-PasTonMolo:
I see, I just misunderstood you. Ok I'll fix it

jubalh (approved):

PR #1648, opened 2022-03-13
Refactor editor & some other parts

sjaeckel:
This PR started by looking into the editor implementation and finding some erroneous behavior https://github.com/profanity-im/profanity/pull/1645#issuecomment-1057127993

Now you can have a look at the commit history what was all done ... :joy: 

Most important things:

* fixing the issue described in the linked comment above
* changed editor spawning to `Alt+c` as `Alt+e` is already bound to something else
* a bunch of refactorings and simplifications (I hope I didn't break anything :grimacing:)

### DIFF
--- a/src/config/files.c
+++ b/src/config/files.c
@@ -151,13 +151,8 @@ gchar*
 files_get_config_path(const char* const config_base)
 {
     gchar* xdg_config = _files_get_xdg_config_home();
-    GString* file_str = g_string_new(xdg_config);
-    g_string_append(file_str, "/profanity/");
-    g_string_append(file_str, config_base);
-    gchar* result = g_strdup(file_str->str);
+    gchar* result = g_strdup_printf("%s/profanity/%s", xdg_config, config_base);
     g_free(xdg_config);
-    g_string_free(file_str, TRUE);
-
     return result;
 }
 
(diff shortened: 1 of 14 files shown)


src/config/files.c
```diff
@@ -149,48 +151,53 @@ gchar*
...
+    gchar* result = g_strdup_printf("%s/%s", data_dir, account_dir);
+
+    g_free(account_dir);
+    g_free(data_dir);
 
     return result;
 }
 
+gchar*
+files_file_in_account_data_path(const char* const specific_dir, const char* const jid, const char* const file_name)
```
jubalh, line 183: Hmm is there `get` missing? `files_*get*_file_in_account_data_path` ?

jubalh (changes requested):
Cool! Thank you!

I made some comments about minor stuff!

src/config/files.c
sjaeckel: Well, I explicitly didn't add the `get` since it differs in its behavior to all the other functions as it creates the necessary path as well ... I would've liked to name it `file_in_account_data_path` but then the prefix wouldn't fit anymore ...

The name is bad, that's true ... but `files_get_file_in_account_data_path_and_create_path` is even worse :D ... and `files_get_file_in_account_data_path` would make me expect to have the same behavior than the other getters, to only construct the string ... 

We could also change the behavior of the other getters to create the path as well!?

If you want me to add

src/tools/editor.c
```diff
@@ -35,97 +36,92 @@
...
+            if (glib_error) {
+                g_error_free(glib_error);
+            }
+            g_free(filename);
+            g_free(editor);
             return TRUE;
         }
-        g_object_unref(file);
+        /* Remove all trailing new-line characters */
+        while (contents[length - 1] == '\r' || contents[length - 1] == '\n') {
```
jubalh, line 113: I think we could use `g_strchomp()` here.

src/config/files.c
jubalh: I see!
I'm undecided what's best :D 
Maybe just leave it as you have it right now :) I will make a note to check the naming of all the functions in that file on my todolist.

src/tools/editor.c
sjaeckel: nice, I forgot that this one exists

jubalh (approved):

Issue #1649, opened 2022-03-20
Add autocompletion for quotes


MarcoPolo-PasTonMolo:
Autocomplete previous messages when `>` is typed at the beginning of the line

I already have implemented most of this. Just posting this to be sure that it is something that's accepted by the main devs.
## Expected Behavior
When I type `>` and then tab or shift tab previous messages should autocomplete

## Current Behavior
No previous message autocompetion

jubalh:
So what would the workflow look like? I guess something like:

* type >
* tab until correct message is autocompleted
* press key to open editor
* enter for newline
* type reply

Or do you have another plan about the newline so that the formatting in other clients will be correct for the quote?

MarcoPolo-PasTonMolo:
That's the workflow I had in mind

jubalh:
Sounds good to me!

PR #1651, opened 2022-03-21
Add quote autocompletion for previous messages

MarcoPolo-PasTonMolo:
Fixes https://github.com/profanity-im/profanity/issues/1649

Type `>` then press tab or shift tab to autocomplete previous messages,
then type your reply and send message.
Newlines are replaced with newline followed by `> `.
A newline is added at the end so that the user can immediately type a
message without manually adding a new line.

I'm not sure if win_quote_autocomplete is in the correct file. I thought ui/window_list.c might have also been a good place to put it in since there are some other related functions there.

### DIFF
--- a/src/ui/window.c
+++ b/src/ui/window.c
@@ -2020,3 +2020,26 @@ win_insert_last_read_position_marker(ProfWin* window, char* id)
 
     g_date_time_unref(time);
 }
+
+char*
+win_quote_autocomplete(ProfWin* window, const char* const input, gboolean previous)
+{
+    if (window->type != WIN_CHAT && window->type != WIN_MUC && window->type != WIN_PRIVATE) {
+        return NULL;
+    }
+
+    char* result = autocomplete_complete(window->quotes_ac, input + 1, FALSE, previous);
+    if (result == NULL) {
+        return NULL;
+    }
+
+    GString* replace_with = g_string_new("> ");
+    g_string_append(replace_with, result);
+    g_string_replace(replace_with, "\n", "\n> ", 0);
+    g_string_append(replace_with, "\n");
+
+    g_free(result);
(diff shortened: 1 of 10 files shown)


jubalh:
`g_string_replace()` is only available since glib 2.68.
I'll have to check which distros ship which version.

jubalh:
Debian 12 has  2.66.8
Fedora 36 has 2.71.1
openSUSE Leap 15.4 will have 2.70.4

MarcoPolo-PasTonMolo:
Ok, I better change that then

jubalh:
You can also put some ifdef around it (I think we have that in other parts). And later when all distros have this version of the lib we will remove the alternative way.

MarcoPolo-PasTonMolo:
Sorry I don't know how to do that

src/ui/window.c
```diff
@@ -2020,3 +2020,31 @@ win_insert_last_read_position_marker(ProfWin* window, char* id)
...
+    if (result == NULL) {
+        return NULL;
+    }
+
+    gchar **parts = g_strsplit(result, "\n", -1);
+    gchar *quoted_result = g_strjoinv("\n> ", parts);
+
+    GString* replace_with = g_string_new("> ");
+    g_string_append(replace_with, quoted_result);
+    g_string_append(replace_with, "\n");
```
jubalh, line 2041: Is this working for you? When I open in the editor I'm still on the same line.

MarcoPolo-PasTonMolo: If I open it in vim no, if I open it in another editor yes. Also if I don't open it in editor and just write something, the newline is indeed present.

jubalh: Ok!

PR #1652, opened 2022-03-22
Fix #1624

### LINKED ISSUE #1624: /sendfile does not work with self signed cert
When using /sendfile to upload a file to an xmpp server using tls with a self signed certificate, the operation fails, stating "SSL peer certificate or SSH remote key was not OK". This fails even if the account is set to trust the server tls cert with tls trust in the account settings. 
As a sidenote, if the http upload module on the xmpp server is changed to use http for the upload, it is successful. However, when attempting to download the file from the server with /url the same "SSL peer certificate or SSH remote key was not OK" error occurs (because the server returns a https url).

## Expected Behavior
The file should upload properly without an error.

## Current Behavior
The file is not uploaded properly and returns an error instead.

1. Login to xmpp server using self signed cert
2. Attempt to upload file using /sendfile
3. Upload fails

## Context
Attempting to create an xmpp server using a self signed cert

## Environment
* glib version

Profanity, version 0.11.1
Build information:
XMPP library: libmesode
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.10.1)
GTK icons/clipboard: Disabled

Using Arch Linux 5.15.11-arch2-1
ldd (GNU libc) 2.33

sjaeckel:
This PR fixes #1624 

Please read the commit messages for further details.

I first started with only the CAfile, but for some reason that didn't work with the server I tried on. Then I realized that the TLS stack treats an expired certificate differently than a self-signed one. Therefor I've also added the option to completely ignore the validity of the certificates.

This behavior is a bit different than the usual behavior of profanity!
Profanity will allow you to connect to a server with an expired certificate if you add it to the whitelist.
If you also want to use the file-upload part with the same server/certificate this won't be possible. You'd have to completely disable certificate verification.

@elimin-8 can you please check if this solves your problem?


jubalh (approved):
Looks good! Thanks!

Issue #1654, opened 2022-03-24
Url autocompletion does not reset


MarcoPolo-PasTonMolo:
When I use /url open or /url save and get a url via autocompletion, when I try to use the same command autocompletion continues from last opened url instead of newest url in window.

## Expected Behavior
On using /url open a second time autocompletion should start from newest url in window

## Current Behavior
On using /url open a second time autocompletion starts from previously opened url.

## Possible Solution
Resetting url_ac when /url open gets used or when buffer changes or when readline is cleared

## Steps to Reproduce (for bugs)
1.  /url open <tab till random url in window>
2. /url open <tab>

PR #1655, opened 2022-03-24
Reset url autocompletion after open/save

jubalh:
I guess we should reset the position after we ran `/url open|save`.
So that next time `/url open <tab>` starts with the latest entry.

Fix https://github.com/profanity-im/profanity/issues/1654

@MarcoPolo-PasTonMolo I guess this is what we want?


MarcoPolo-PasTonMolo:
Just tried it, seems good to me

PR #1657, opened 2022-03-27
Autoconf fixes

paulfertser:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -348,7 +348,7 @@ AC_CHECK_LIB([expect], [exp_expectl], [AM_CONDITIONAL([HAVE_EXPECT], [true])],
     [AC_MSG_NOTICE([libexpect not found, will not be able to run functional tests])])
 
 ## Default parameters
-AM_CFLAGS="-Wall -Wno-deprecated-declarations -std=gnu99"
+AM_CFLAGS="$AM_CFLAGS -Wall -Wno-deprecated-declarations -std=gnu99"
 AM_LDFLAGS="$AM_LDFLAGS -export-dynamic"
 
 AS_IF([test "x$PACKAGE_STATUS" = xdevelopment],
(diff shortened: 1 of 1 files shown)


jubalh (approved):
Looks good to me!

Though I would prefer to have the second commit split into 2 commits. But not strictly necessary.

And I don't know about:
`While at it, don't miss to append CFLAGS for libstrope which is needed
if it's installed to a non-default location.`.

So would like to hear @sjaeckel opinion as well :-)

paulfertser:
On Sun, Mar 27, 2022 at 11:59:10AM -0700, Michael Vetter wrote:
> Though I would prefer to have the second commit split into 2 commits. But not
> strictly necessary.

So you prefer fully-granular commits, even if one-liners. I'll keep
that in mind, thanks.

> And I don't know about:
> While at it, don't miss to append CFLAGS for libstrope which is needed if it's
> installed to a non-default location..

That surfaced as I used custom PKG_CONFIG_PATH for self-compiled
libstrophe when running ./configure .

configure.ac
```diff
@@ -347,15 +348,15 @@ AC_CHECK_LIB([expect], [exp_expectl], [AM_CONDITIONAL([HAVE_EXPECT], [true])],
     [AC_MSG_NOTICE([libexpect not found, will not be able to run functional tests])])
 
 ## Default parameters
-AM_CFLAGS="-Wall -Wno-deprecated-declarations -std=gnu99"
+AM_CFLAGS="$AM_CFLAGS -Wall -Wno-deprecated-declarations -std=gnu99"
```
sjaeckel, line 351: Shouldn't this be appended? It's the user who defines it, so it should take precedence over our defaults.

paulfertser:
On Mon, Mar 28, 2022 at 03:00:24AM -0700, Steffen Jaeckel wrote:
> In [1]configure.ac:
> 
> > @@ -347,15 +348,15 @@ AC_CHECK_LIB([expect], [exp_expectl], [AM_CONDITIONAL([HAVE_EXPECT], [true])],
>      [AC_MSG_NOTICE([libexpect not found, will not be able to run functional tests])])
> 
>  ## Default parameters
> -AM_CFLAGS="-Wall -Wno-deprecated-declarations -std=gnu99"
> +AM_CFLAGS="$AM_CFLAGS -Wall -Wno-deprecated-declarations -std=gnu99"
> 
> Shouldn't this be appended? It's the user who defines it, so it should take
> precedence over our defaults.

CFLAGS is for the user-specified flags, AM_CFLAGS is for
Autoconf-specified flags. I'm changing it so that flags that Autoconf
considers necessary for the platform are not lost in this line.

sjaeckel (approved):

PR #1659, opened 2022-03-28
Add debug information to binary

paulfertser:
Option -ggdb3 (same as -g3) doesn't affect code-generation, just adds
to the binary size of an executable. Automake strips installed
binaries automatically when "make install-strip" is run so enabling
this option permanently doesn't affect the end result while allowing
for hassle-free interactive debugging.

If one is running a stripped binary it's still possible to attach GDB
by process PID and then use "file /path/to/unstripped/profanity" to
get all symbols and lines data.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -349,6 +349,7 @@ AC_CHECK_LIB([expect], [exp_expectl], [AM_CONDITIONAL([HAVE_EXPECT], [true])],
 
 ## Default parameters
 AM_CFLAGS="$AM_CFLAGS -Wall -Wno-deprecated-declarations -std=gnu99"
+AM_CFLAGS="$AM_CFLAGS -Wall -Wno-deprecated-declarations -std=gnu99 -ggdb3"
 AM_LDFLAGS="$AM_LDFLAGS -export-dynamic"
 
 AS_IF([test "x$PACKAGE_STATUS" = xdevelopment],


jubalh:
Thanks!
Looking forward to more improvements :)

paulfertser:
On Tue, Mar 29, 2022 at 02:16:10AM -0700, Michael Vetter wrote:
> Looking forward to more improvements :)

TBH, what I currently find annoying is the lack of online
documentation (so that I could get myself comfortable with actually
using profanity for any real conversations). I'm used to having Info
or manpages manuals which can be easily searched for keywords. But
this shortcoming can't be fixed with a one-liner it seems :(

sjaeckel:
Maybe this helps?

```
$ ./profanity mangen               

Processed 101 commands.

$ man docs/profanity.1
```

but I think I know what you're talking about ... I'd also like to see better examples, how stuff has to be done ... profanity is not really user friendly, it has a lot of commands and most of them are documented and if someone knows what they want to do, they will find their way around ... but from scratch, no f* way ...

sjaeckel:
> And the same content like there we also have in man pages in our release tarballs. It is auto generated.
> And then there is also: https://profanity-im.github.io/guide/latest/userguide.html

maybe I have to take back that last part ... I forgot about the userguide

paulfertser:
On Tue, Mar 29, 2022 at 02:44:58AM -0700, Michael Vetter wrote:
> This should be easily searchable:
> [1]https://profanity-im.github.io/guide/latest/reference.html

That's a documentation on some Internet server, it's not "online" in
the sense that it's not a documentation that comes installed locally
along with the binary.

> And the same content like there we also have in man pages in our release
> tarballs. It is auto generated.

That I missed, sorry. It somehow wasn't clear to me from reading
README files that man pages are generated which have at least the
commands reference.

> And then there is also:
> [2]https://profanity-im.github.io/guide/latest/userguide.html

Is it only available via the Internet, and not "online"?

> Additionally we try to explain some things on our blog:
> [3]https://profanity-im.github.io/blog/
> 
> Which documentation is lacking in your opinion?

Guess it looks like what I'm missing is the way to comfortably read
and search through user guide without relying on third-party machines
for that.

You know old habits die hard so probably I'm just too reluctant to
switch away from mcabber that I was using happily for years...

jubalh:
In our MUC we discussed several times that some advanced users could document their workflow and usage practises and put it into blog posts.
But everybody is busy and we have to decide what we can work on :)

jubalh:
> That's a documentation on some Internet server, it's not "online" in
the sense that it's not a documentation that comes installed locally
along with the binary.

This is exactly the same content like we install locally ;)

jubalh:
BTW on my personal TODO I have this:

Writing a `/wizard` command. It is automatically run when we see that there is nor `profrc` file yet (firstrun). Helps the user to setup their account and then guides them through the most basic Profanity commands.

But you know.. priorities and time :-)

jubalh:
> You know old habits die hard so probably I'm just too reluctant to
switch away from mcabber that I was using happily for years...

I understand that you mean that you are used to this.

Regarding newcomers thought:

I just checked the documentation of mcabber. And it seems to me like Profanity has a lot of more resources available for people to get started and learn how to use it.
We have the man pages, the user guide website, the blog, the in client help.

And you can even `git clone https://github.com/profanity-im/profanity-im.github.io` read this not on another machine ;)

PR #1660, opened 2022-03-29
Input window: handle invalid multibyte

paulfertser:
The current code enters an infinite loop if the input string happens
to get an invalid utf-8 sequence somehow. For me it was reproducible
by running profanity in a Screen session and pressing Alt-т (cyrillic
letter).

Fix it the way borrowed from 0501e49623f68aa39508e4e622924c1dd8147588
where mbrlen is used for the same purposes.


jubalh (approved):

jubalh:
Cool, thank you!

PR #1661, opened 2022-03-29
make OMEMO respect resource settings

akaWolf:
non-encrypted messages sending respecting «/resource set» setting
omemo-encrypted messages uses default resource (ignoring that setting)
this PR make behavior of both cases consistent

### DIFF
--- a/src/xmpp/message.c
+++ b/src/xmpp/message.c
@@ -619,12 +619,13 @@ message_send_chat_otr(const char* const barejid, const char* const msg, gboolean
 
 #ifdef HAVE_OMEMO
 char*
-message_send_chat_omemo(const char* const jid, uint32_t sid, GList* keys,
+message_send_chat_omemo(const char* const barejid, uint32_t sid, GList* keys,
                         const unsigned char* const iv, size_t iv_len,
                         const unsigned char* const ciphertext, size_t ciphertext_len,
                         gboolean request_receipt, gboolean muc, const char* const replace_id)
 {
-    char* state = chat_session_get_state(jid);
+    char* state = chat_session_get_state(barejid);
+    char* jid = chat_session_get_jid(barejid);
     xmpp_ctx_t* const ctx = connection_get_ctx();
     char* id;
     xmpp_stanza_t* message;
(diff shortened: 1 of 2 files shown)


sjaeckel:
IMO this shall be closed as [XEP-0384](https://xmpp.org/extensions/xep-0384.html) states in [Ch. 5.5.3](https://xmpp.org/extensions/xep-0384.html#message-structure-description) that the current behavior is correct and this patch would introduce non-standard behavior.

DebXWoody:
> which one sentence there declares the message should be sent only to resource with the highest priority?

Are you talking about RFC-3921 - 11.1.  Inbound Stanzas?

Anyway, I think we should not send messages to fulljid. I think
the barejid is fine. The server should route to the resources
which should be used. At least this is my understanding how it
should work.

akaWolf:
> Anyway, I think we should not send messages to fulljid. I think the barejid is fine. The server should route to the resources which should be used. At least this is my understanding how it should work.

why there is «/resource set» ever then?

jubalh:
> why there is «/resource set» ever then?

Because we want to be an advanced client.
And leave this option for special use cases and debugging purposes.

Many clients removed this option already. As per suggestion of [modernxmpp](https://docs.modernxmpp.org/).

In the modern xmpp world we don't want to send to a certain resource but let the server handle this.

jubalh:
The change makes it send to the resource always I think, not just in the special case when /set resoruce was used.
I think this change will break more than it will solve.

sjaeckel:
> > IMO this shall be closed as [XEP-0384](https://xmpp.org/extensions/xep-0384.html) states in [Ch. 5.5.3](https://xmpp.org/extensions/xep-0384.html#message-structure-description) that the current behavior is correct and this patch would introduce non-standard behavior.
> 
> which one sentence there declares the message should be sent only to resource with the highest priority?

Right, that's not specified in XEP-0384.

> The change makes it send to the resource always I think, not just in the special case when /set resoruce was used.

Indeed.

> I think this change will break more than it will solve.

I think so too

Why would you want to specify your resource manually?

jubalh:
I'll close this one as I think it's not right.
But please don't feel discouraged. I like your enthusiasm and clearly you can contribute on other parts :-)

PR #1662, opened 2022-03-29
editor: auto add slash

akaWolf:
if message starts with slash, add another one to send as is.

### DIFF
--- a/src/tools/editor.c
+++ b/src/tools/editor.c
@@ -112,6 +112,13 @@ get_message_from_editor(gchar* message, gchar** returned_message)
             g_free(editor);
             return TRUE;
         }
+        if (length && contents[0] == '/') {
+            gchar* new_content = g_malloc(length + 1);
+            new_content[0]='/';
+            g_stpcpy(new_content + 1, contents);
+            g_free(contents);
+            contents = new_content;
+        }
         /* Remove all trailing new-line characters */
         g_strchomp(contents);
         *returned_message = contents;


jubalh:
> if message starts with slash, add another one to send as is.

Unfortunately I don't understand what this means. Or why we would add another slash?

akaWolf:
> > if message starts with slash, add another one to send as is.
> 
> Unfortunately I don't understand what this means. Or why we would add another slash?

messages beginning with slash will be interpreted as commands
messages with two slashes -- not

MarcoPolo-PasTonMolo:
If we want this, there should be some other checks in place, for example using alt+c does not neccessarily mean that we don't want a command so in that case a slash shouldn't be added! Same for /correct-editor and /subject-editor.

akaWolf:
> If we want this, there should be some other checks in place, for example using alt+c does not neccessarily mean that we don't want a command so in that case a slash shouldn't be added! Same for /correct-editor and /subject-editor.

is there any reason to spawn editor for enter a command?
I guess command is always one-liner, so no need in editor.
and yeah there is no commands logging so you you can't edit a command.
so I think only one use case to spawn an editor is to edit a text message (which obvious can't be a command).

MarcoPolo-PasTonMolo:
Personally I do use it for commands sometimes. Also for subject and correct editor you shouldn't duplicate the slash because in those cases one slash does get recognized

jubalh:
Honestly I'm not understanding why we want this change or what's the idea behind it.

> is there any reason to spawn editor for enter a command?

No we usually don't do this. Just for some commands it makes sense. Like editing multiline. Like editing MUC subjects.

>  and yeah there is no commands logging so you you can't edit a command.

Do you mean with command logging something like history? Because we have readline history to retrieve earlier messages/commands.

> so I think only one use case to spawn an editor is to edit a text message (which obvious can't be a command).

Ahh maybe now I get what you mean. But I don't see a reason for this.
To me knowledge with our current implementation we cover all use cases.

akaWolf:
> Ahh maybe now I get what you mean. But I don't see a reason for this. To me knowledge with our current implementation we cover all use cases.
you don't cover a use case when someone wants to put a raw text starting with slash. that PR is aiming to close that gap.

jubalh:
> you don't cover a use case when someone wants to put a raw text starting with slash. that PR is aiming to close that gap.

Ok, finally I understood the purpose :-) Thanks for explaining.
Please try to summarize this a little more clear in the commit message (next time).

I'm unsure whether we want this. Let me think a little. Usuall we just add one space ` /` for this so far.

akaWolf:
> Personally I do use it for commands sometimes. Also for subject and correct editor you shouldn't duplicate the slash because in those cases one slash does get recognized

okay. if you want to have editor for enter a commands, then we can have an additional command to send raw text w/o any processing

akaWolf:
> I'm unsure whether we want this. Let me think a little. Usuall we just add one space ` /` for this so far.

yeah but that solution is require to keep in mind one more thing (rule) when you just want to send a multi-line text.

jubalh:
> yeah but that solution is require to keep in mind one more thing (rule) when you just want to send a multi-line text.

That's true :)

But some commands are multi line. Like `/subject`. And there we also have `/subject editor`.
So the question is: do we want to have an `editor` subcommand for all commands and disallow commands otherwise in multiline texts?

I have to say that I liked that we can just use multiline for everything. And it seems less of a hassle to me to add a space in the rare case of wanting to send a text starting with `/`.

akaWolf:
@jubalh then as I suggest before we can have separate command which spawn editor, get result, then send that message raw w/o processing.

sjaeckel:
@akaWolf please correct me if I'm wrong, but from what I understand your single target is to be able to send messages that start with a `/`, right?

IMO that's a valid use-case, ~~but from my feeling we should~~
1. ~~escape with something else than `/`. I'd propose to use `\` as that's a common escape-character (at least for programmers)~~
2. ~~should this really be limited to the editor mode? if we support it in one part, shouldn't we also support it directly!? (i.e. enter `\/foo` in any chat window will send `/foo`)~~

> have separate command which spawn editor, get result, then send that message raw w/o processing

I like this idea, should also be easier to implement than what I wrote above... even though once it's implemented the next user will ask "hey why can we only send raw messages like that and not in the regular chat windows" ... :smile: ... but maybe we should wait for that moment ;)

jubalh:
> we can have separate command which spawn editor, get result, then send that message raw w/o processing.

I think that's a better idea actually. But still not a big fan of this altogether :D I think adding a space is just the easiest way. Also for the users the most natural way.

> hey why can we only send raw messages like that and not in the regular chat windows

I wonder how irssi is doing this? Probably you can't do this there neither. Like in all other clients most likely.

akaWolf:
I guess you can close it PR in flavor of !1685

PR #1663, opened 2022-03-29
Add stamp settings

akaWolf:
command allow override standard stamps of incoming and outgoing messages

current format of the single line in chat window next: `<timestamp> <encryption sign> <stamp> <message>`
where <stamp> is "me:" for incoming messages or "username@server/resource" for outcoming messages.

this PR would allow customizing this stamp.


jubalh:
What is a `stamp` in your opinion?
Can you add more descriptive explanation, like what this change does. How did it look before and how will it look after your change applied? Maybe try to read the contributing.md file where we tried to explain some requirements.

akaWolf:
> How do I know which user send a message if the only thing I see is `<--`?

it should work only in chat window. and you see JID at the top of chat window.

akaWolf:
@jubalh I've update description and add options to show and unset stamp configuration

src/xmpp/roster_list.c
```diff
@@ -199,6 +199,12 @@ roster_get_msg_display_name(const char* const barejid, const char* const resourc
 {
     assert(roster != NULL);
 
+    char* incoming_str = prefs_get_string(PREF_INCOMING_STR);
```
jubalh, line 202: what is this part doing?

src/config/preferences.h
```diff
@@ -176,6 +176,8 @@ typedef enum {
     PREF_URL_SAVE_CMD,
     PREF_COMPOSE_EDITOR,
     PREF_SILENCE_NON_ROSTER,
+    PREF_OUTGOING_STR,
```
jubalh, line 179: I think you should somehow add `stamp` there.
Something like `PREF_OUTGOING_STAMP`

akaWolf: you can grep for PREF_INCOMING_STR and PREF_OUTGOING_STR through this PR
it's exactly the same what you mean by PREF_OUTGOING_STAMP

src/xmpp/roster_list.c
akaWolf: returns the value of "/stamp incoming" string.

src/config/preferences.h
jubalh: I think `PREF_OUTGOING_STR` is a too general term. Since this is about stamps I think it should be reflected in the name.

jubalh:
BTW Just in case this isn't clear. I was waiting for the changing of the PREF_OUTGOING_STR to something more descriptive.

src/config/preferences.h
jubalh: I did that.

jubalh (approved):

jubalh:
Thanks @akaWolf !
One step closer to make mcabber users happy I guess!

jubalh:
@akaWolf @mdosch found a mistake in your PR:
https://github.com/profanity-im/profanity/commit/4686a9ba2ce884396d9f1d27a8054ca26172f601#r72600186

Can you adress that?

PR #1665, opened 2022-03-30
More cleanup

sjaeckel:
Please check the commit messages for details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ profanity.workspace
 compile_commands.json
 .tern-port
 .tern-project
+.cproject
+.project
+.settings/
 
 # autotools
 .libs/
@@ -25,6 +28,7 @@ build-aux/
 config.log
 config.status
 configure
+configure*~
 libprofanity.la
 libtool
 m4/
@@ -82,6 +86,7 @@ python3/
 
 .DS_Store
 *.bak
+*.orig
 breaks
 
 *.tar.*
--- a/Makefile.am
+++ b/Makefile.am
@@ -221,7 +221,7 @@ icons_sources = $(top_srcdir)/icons/*
 
 script_sources = bootstrap.sh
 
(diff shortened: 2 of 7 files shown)


Makefile.am
```diff
@@ -221,7 +221,7 @@ icons_sources = $(top_srcdir)/icons/*
 
 script_sources = bootstrap.sh
 
-man_sources = $(wildcard docs/profanity*.1)
+man1_sources = $(wildcard $(top_srcdir)/docs/*.1)
```
jubalh, line 224: please mention in this commit message the commit that introduced the mistake. some explanation :)

.gitignore
```diff
@@ -60,6 +64,9 @@ gitpushall.sh
 main_fragment.html
 toc_fragment.html
 
+# man pages
+docs/*.1
```
jubalh, line 68: why? :) I like to see those in `git status`

jubalh (changes requested):

Makefile.am
sjaeckel: better now? :joy:

jubalh: yes :)

.gitignore
sjaeckel: because they are auto-generated!? :)

jubalh: Not all of them. Only the `profanity-*.1`.
And also I add them for each "release commit" and remove them later. And I like to see that with `git status` :)

sjaeckel: > Not all of them. Only the `profanity-*.1`.

OK, fair point, I'll update the ignore to `profanity-*.1`, OK?

or do I have to remove it completely?

jubalh: Since i add them in each release commit I would prefer to remove this line completey. Just so I see this and don't miss anything.

jubalh (approved):

jubalh:
Coding style CI is really cool :)

PR #1668, opened 2022-03-30
Document custom input configuration in man page

paulfertser:

### DIFF
--- a/docs/profanity.1
+++ b/docs/profanity.1
@@ -87,6 +87,18 @@ for current input line.
 .TP
 .BI CTRL+DOWN
 Store current input line in history but do not execute it.
+.PP
+.I Note:
+Profanity is using GNU Readline library to handle input so
+default configuration file
+.I ~/.inputrc
+affects operation. In addition to that
+.I $XDG_CONFIG_HOME/profanity/inputrc
+is read after all default keybindigs are set so one can override
+or add settings there. For reference, see Readline documentation:
+.I "info readline ""Command Line Editing"" ""Readline Init File"" ""Readline Init File Syntax"""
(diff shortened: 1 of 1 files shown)


jubalh:
Looks good!
I assume you saw https://profanity-im.github.io/guide/latest/keybindings.html as well?

paulfertser:
On Wed, Mar 30, 2022 at 12:53:03PM -0700, Michael Vetter wrote:
> Looks good!
> I assume you saw [1]https://profanity-im.github.io/guide/latest/keybindings.html
> as well?

Heh, in fact, no, as it's not part of the repository. I did "git grep
-i" in profanity sources to make sure I wasn't missing anything, I'm
used to having all documentation maintained along with the source code
in a single repo just because of it.

jubalh:
Just clone the website repo as well and grep there too :)

paulfertser:
On Wed, Mar 30, 2022 at 02:10:54PM -0700, Michael Vetter wrote:
> Just clone the website repo as well and grep there too :)

Guess I'd need to use "git subtree" trickery for "git grep" to work
with that. Sounds kinda like a plan but I still do not like the idea
that only part of the documentation is available online, and
everything else needs to be fetched from the Internet somehow.

jubalh:
Anyways I'm only linking you this website in case you want to use more information from there in the manpage. Or vice versa.

jubalh:
I remember when I started using Profanity I always was looking for all this information locally! I guess I just got used to it. But now it's cool when we improve this :)

jubalh:
So you want to leave this PR like this? Then I'm going to merge.

jubalh (approved):

Issue #1235, opened 2019-12-01
ncurses terminal resize improvement


jubalh:
I think somewhere ncurses redraw/refresh/update functions are not used properly.

If we change the terminals window size profanity is not redrawn correctly. Also some parts of the screen are not updated correctly but I don't know how to reproduce it yet and might create another issue later.

In this example I run profanity in gnome-terminal.

1. start profanity:
![prof-1](https://user-images.githubusercontent.com/1658215/69917480-e76b2580-1466-11ea-828f-76fd84fd1f69.png)

2. Press F11 to have fullscreen:
![prof-2](https://user-images.githubusercontent.com/1658215/69917491-f94cc880-1466-11ea-9db9-b1cf73da6f03.png)

Already now it looks weird.

3. Press F11 again to have the regular window size again:
![prof-3](https://user-images.githubusercontent.com/1658215/69917504-171a2d80-1467-11ea-8720-3344c01ad3e2.png)

Looks even weirder.

jubalh:
@xaizek sorry to bother you with this. But I know you know ncurses well. Do you have an idea what is happening here?

mdosch:
I don't have this problem with gnome-terminal on Debian Testing with i3/X.org:

```
#  GNOME Terminal 3.34.2 using VTE 0.58.2 +BIDI +GNUTLS
```
It works with F11 as well as with Super+F (i3 fullscreen).

xaizek:
In functions like `status_bar_resize()` and `title_bar_resize()` windows are moved before being resized. The operations should be done in the opposite order (resize and then move), because

    If the move would cause the window to be off the screen, it is an error and the window
    is not moved.

Resize on the other hand doesn't fail like this according to its documentation. So new size needs to be applied first.

Not sure if it will make all issues go away, but some might be fixed. I just glanced over the code and didn't actually build and run it. Fix order of operations and will see if anything else needs to be done.

jubalh:
@xaizek thanks for the feedback!

So my local code looks like this now:

```
diff --git a/src/ui/statusbar.c b/src/ui/statusbar.c
index 39b1ddbb..8f9900e1 100644
--- a/src/ui/statusbar.c
+++ b/src/ui/statusbar.c
@@ -137,8 +137,8 @@ status_bar_resize(void)
     int cols = getmaxx(stdscr);
     werase(statusbar_win);
     int row = screen_statusbar_row();
-    mvwin(statusbar_win, row, 0);
[...]

PR #1671, opened 2022-04-01
Fix race condition on resize

paulfertser:
The current code is inherently racy: if screen update takes
considerable time (e.g. when working over network) and a user
performed a series of resizes the final event might get ignored and
the display will be left in inconsistent state.

Fix the race by unsetting the flag first so if the next WINCH signal
is received while display is resizing it'll be processed on the next
iteration.

### DIFF
--- a/src/ui/core.c
+++ b/src/ui/core.c
@@ -144,10 +144,8 @@ ui_update(void)
     doupdate();
 
     if (perform_resize) {
-        signal(SIGWINCH, SIG_IGN);
-        ui_resize();
         perform_resize = FALSE;
-        signal(SIGWINCH, ui_sigwinch_handler);
+        ui_resize();
     }
 }
 


sjaeckel (approved):

jubalh:
Cool thank you!
I remember we had some issues with this!
Maybe this can even fix https://github.com/profanity-im/profanity/issues/1235 ?

paulfertser:
On Fri, Apr 01, 2022 at 03:14:40AM -0700, Michael Vetter wrote:
> I remember we had some issues with this!
> Maybe this can even fix [1]#1235 ?

If it's indeed due to "On Wayland with GNOME Shell / Mutter and
friends, CSD (client-side decorated) apps don't maximize/restore in a
single step, they resize through an intermediate incorrect size" then
it looks related, yes. Are you still able to reproduce it?

FWIW, I'm running urxvt on X11 with XMonad, connecting to a remote
with "mosh", attaching to a Screen session there, TERM=screen-256color
and it was reproducible consistently by doing Tall->Mirror
Tall->Full->Tall switching fast enough.

jubalh:
Issue really seems to be gone!

Thanks a lot! I suffered from this a long time. And many people couldn't even reproduce it.

paulfertser:
On Fri, Apr 01, 2022 at 03:41:02AM -0700, Michael Vetter wrote:
> Issue really seems to be gone!
> 
> Thanks a lot! I suffered from this a long time. And many people couldn't even
> reproduce it.

My pleasure :)

PR #1675, opened 2022-04-01
Use correct colour for their nick when showing history

paulfertser:


jubalh:
What does 'correct' mean in this case?

paulfertser:
On Sat, Apr 02, 2022 at 01:30:27PM -0700, Michael Vetter wrote:
> What does 'correct' mean in this case?

[colors].them rather than [colours].me should be used for messages not
from me.

PR #1681, opened 2022-04-02
fix theme load

akaWolf:
disable execution of colors-changing code when call for theme load and it doesn't exist


jubalh:
Thanks!

PR #1682, opened 2022-04-02
change connection error string

akaWolf:
change connection error string from "login failed" to "connection closed"

### DIFF
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -231,8 +231,8 @@ sv_ev_lost_connection(void)
 void
 sv_ev_failed_login(void)
 {
-    cons_show_error("Login failed.");
-    log_info("Login failed");
+    cons_show_error("Connection closed.");
+    log_info("Connection closed");
     tlscerts_clear_current();
 }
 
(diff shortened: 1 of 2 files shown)


jubalh:
Hmmm is that true? In all cases we want this text? I'll need to check more in detail.

akaWolf:
> Hmmm is that true? In all cases we want this text? I'll need to check more in detail.

yes because libstrophe returns that event in all cases, disconnect doesnt used.

sjaeckel:
It's true that libstrophe only gives a disconnect-event and profanity can't know whether the server isn't reachable or if the login really failed.

Having a look at the code it seems legit to keep the "login failed" message as it gives a better description of what most likely happened.

So this boils down to an issue when it comes to user-friendliness of libstrophe. You wanna open an issue in [libstrophe](https://github.com/strophe/libstrophe) referencing this one? Maybe I'll have the time to work on stuff like that sometime in the future.

jubalh:
Let's leave it as is and use more detailed errors once libstrophe reports them.

PR #1684, opened 2022-04-03
Add all the generated man pages to .gitignore

paulfertser:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,10 @@ apidocs/python/src/plugin.pyc
 apidocs/c/html/
 apidocs/c/doxygen_sqlite3.db
 
+# man pages
+docs/*.1
+!docs/profanity.1
+
 # Temp Vim files
 **/*.swp
 


jubalh:
This one I don't like :)

Usually people don't run the mangen command. Or they install them immediately afterwards, so can remove the generated ones again.

But when I create a release I add those generated man pages to the release commit (so it's the same like the tarball). For this reason I want to see these generated manpages and not have them ignored.

(BTW all the generated ones are `profaity-*.1` so you can just add this to your personal gitignore file).

paulfertser:
On Sun, Apr 03, 2022 at 02:42:32PM -0700, Michael Vetter wrote:
> But when I create a release I add those generated man pages to the release
> commit (so it's the same like the tarball). For this reason I want to see these
> generated manpages and not have them ignored.

I see, thank you for clarifying your slightly unusual workflow :)

FWIW, automake has "dist" target that's specifically meant for
creating source release tarballs (and "distcheck" to verify they're
buildable) but probably digging that does not worth the effort.

jubalh:
> FWIW, automake has "dist" target that's specifically meant for
creating source release tarballs

Thanks :) Yes I'm aware. We use them (mentioned here: https://github.com/profanity-im/profanity/blob/master/RELEASE_GUIDE.md).

PR #1685, opened 2022-04-03
add raw command

akaWolf:
see https://github.com/profanity-im/profanity/pull/1662#issuecomment-1082841202

### DIFF
--- a/src/command/cmd_defs.c
+++ b/src/command/cmd_defs.c
@@ -2624,6 +2624,24 @@ static struct cmd_t command_defs[] = {
       CMD_NOEXAMPLES
     },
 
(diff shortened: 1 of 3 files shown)


jubalh:
Please expand the description a little. To explain to a user what sending raw means.

akaWolf:
> Please expand the description a little. To explain to a user what sending raw means.

done

paulfertser:
Looks like this adds the third almost identical function for /editor (the other two being _inp_rl_send_to_editor() and cmd_editor()) :/

jubalh:
Then I guess we don't need this.

PR #1690, opened 2022-04-11
Show return symbol for embedded newlines

paulfertser:
When editing multi-line messages or comments everything past the first
newline becomes invisible.

This patch fixes it by substituting a Unicode symbol for "return"
instead of printing the newline as is. On locales where it's not
available single backslash is used instead.

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -9511,13 +9511,6 @@ cmd_change_password(ProfWin* window, const char* const command, gchar** args)
 gboolean
 cmd_editor(ProfWin* window, const char* const command, gchar** args)
 {
-    jabber_conn_status_t conn_status = connection_get_status();
-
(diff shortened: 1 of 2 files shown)


jubalh:
Good idea!

When using `/editor` I don't see this. Is it possible to add it there as well?

And just some minor point:
Do we also want this symbol in the following case?
Screenshot from 2022-04-11 13-29-25

jubalh (changes requested):
Please fix tests/style.
And would be great if we could fix `/editor` to work as well.

paulfertser:
Odd github doesn't let me see it. Am I supposed to be logged in or something?
I'll work on fixing the warning another way, thank you.

jubalh:
It works fine for me with a browser where I'm not logged in.

jubalh (approved):

jubalh:
Thanks!

PR #1696, opened 2022-04-21
Fix xscreensaver detection

### LINKED ISSUE #1695: Idle detection via xscreensaver does not work
## Expected Behavior
Status is set to 'away' when mouse movements are stopped longer than `autoaway.awaytime` seconds and back to 'online' once mouse is moved again.

## Current Behavior
Status is set to 'away' after configured timeout but not restored to 'online' if mouse is moved. It is restored to 'online' once commands are typed.

## Possible Solution
I have noticed that neither the binary nor the library is linked against `libXss.so`, which should be the case with `--with-xscreensaver` being supplied to `configure`.

Checking configure.log I found `xscrnsaver_LIBS='-lXss '` so the configure stage seems to have found the library via `pkgconf`, but somehow that library does not make it into `LIBS`.

Unfortunately I don't know enough about autoconf/automake to trace this further down. Could it be a side effect of 28a9605a?

1. build and install Profanity (with `--with-xscreensaver`, see below)
2. enable autoaway

## Environment
Output of `profanity -v`:

    Profanity, version 0.12.1
    
    Build information:
    XMPP library: libstrophe
    PGP support: Enabled (libgpgme 1.17.1)
    OMEMO support: Enabled
    C plugins: Enabled
    Python plugins: Enabled (3.10.4)
    GTK icons/clipboard: Enabled

Arguments to `configure` from Arch Linux package:

    ./configure \
          --prefix=/usr \
          --enable-icons-and-clipboard \
          --with-xscreensaver \
          --enable-notifications \
          --enable-python-plugins \
          --enable-c-plugins \
          --enable-plugins \
          --enable-otr \
          --enable-omemo \
          --enable-pgp

paulfertser:
Using pkg-config to find libraries requires explicit mention of the
relevant _CFLAGS and _LIBS variables.

Fixes #1695.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -270,7 +270,8 @@ AS_IF([test "x$PLATFORM" = xosx],
 dnl feature: xscreensaver
 AS_IF([test "x$enable_xscreensaver" != xno],
     [PKG_CHECK_MODULES([xscrnsaver], [xscrnsaver],
-        [AC_MSG_NOTICE([xscreensaver support is enabled])],
+        [AC_MSG_NOTICE([xscreensaver support is enabled]);
+         LIBS="$xscrnsaver_LIBS $LIBS" CFLAGS="$CFLAGS $xscrnsaver_CFLAGS"],
         [AS_IF([test "x$enable_xscreensaver" = xyes],
             [AC_MSG_ERROR([xscreensaver is required but does not exist])],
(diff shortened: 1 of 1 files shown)


jubalh (approved):

jubalh:
Thank you!

PR #1700, opened 2022-04-29
Use ISO8601 as date format in log files

jubalh:
Let's use ISO8601 in our chatlogs and log files.

Chatlogs are at:
`~/.local/share/profanity/chatlogs/me_at_domain.org/other_at_domain.org/YYYY_MM_DD.log`

Log files are at:
`~/.local/share/profanity/logs/profanity.log`


sjaeckel (approved):
:heart:

PR #1701, opened 2022-04-29
Log file improvements

jubalh:
Saner defaults and one cleanup.

### DIFF
--- a/src/config/preferences.h
+++ b/src/config/preferences.h
@@ -42,7 +42,7 @@
 #include <glib.h>
 
 #define PREFS_MIN_LOG_SIZE 64
-#define PREFS_MAX_LOG_SIZE 1048580
+#define PREFS_MAX_LOG_SIZE (10 * 1024 * 1024)
 
 // represents all settings in .profrc
 // each enum value is mapped to a group and key in .profrc (see preferences.c)
--- a/src/log.c
+++ b/src/log.c
@@ -59,9 +59,6 @@
 static FILE* logp;
 static gchar* mainlogfile = NULL;
 static gboolean user_provided_log = FALSE;
-
-static GTimeZone* tz;
-static GDateTime* dt;
 static log_level_t level_filter;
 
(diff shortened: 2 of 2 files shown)


src/config/preferences.h
```diff
@@ -42,7 +42,7 @@
 #include <glib.h>
 
 #define PREFS_MIN_LOG_SIZE 64
-#define PREFS_MAX_LOG_SIZE 1048580
+#define PREFS_MAX_LOG_SIZE 10485760
```
sjaeckel, line 45:
```suggestion
#define PREFS_MAX_LOG_SIZE (10 * 1024 * 1024)
```

but that's personal preference, you can also keep it as is.

src/log.c
```diff
@@ -237,15 +234,25 @@ _rotate_log_file(void)
     size_t len = strlen(log_file);
     gchar* log_file_new = malloc(len + 4);
 
-    // find an empty name. from .log -> log.01 -> log.99
-    for (int i = 1; i < 100; i++) {
-        g_sprintf(log_file_new, "%s.%02d", log_file, i);
+    // the mainlog file should always end in '.log', lets remove this last part
+    // so that we can have profanity001.log later
```
sjaeckel, line 238: let's use the "standard way"^TM when naming logfiles which is separated by a `.` - `profanity.NNN.log`

src/log.c
```diff
@@ -192,7 +189,7 @@ void
 log_msg(log_level_t level, const char* const area, const char* const msg)
 {
     if (level >= level_filter && logp) {
-        dt = g_date_time_new_now(tz);
+        dt = g_date_time_new_now_local();
```
sjaeckel, line 192: does `dt` really have to have global storage? should be sufficient to do 
```suggestion
        GDateTime *dt = g_date_time_new_now_local();
```
and after usage `g_date_time_unref(dt)`

sjaeckel (approved):

PR #1706, opened 2022-05-07
Improve grammar.

mdosch:
I am no native english speaker, but for me this sounds
a lot better. Comment of a native speaker is very
much appreciated.


jubalh:
I had the same urge some time ago :D :D 
Then contacted some UK native and they told me that both is equally fine.

However I think we can still take this to make us non native English speakers feel better :D :D

mdosch:
Just now MattJ told me both are fine and the first one sounds more natural. I am pretty surprised, but in this case I think we better don't touch it.

jubalh:
I was surprised as well :D

jubalh:
Closing as discussed with @mdosch

PR #1712, opened 2022-05-25
Update capabilities of muc on available presence

MarcoPolo-PasTonMolo:
Updates whether the muc is members only and its anonymity type.
Problem: the caps request is not always sent, it is always sent when name of muc is changed but not if the members only or anonymity type is changed. Any tips?

Should fix 1347 when completed

### DIFF
--- a/src/xmpp/iq.c
+++ b/src/xmpp/iq.c
@@ -892,13 +892,6 @@ _caps_response_id_handler(xmpp_stanza_t* const stanza, void* const userdata)
             log_debug("Capabilities not cached: %s, storing", given_sha1);
             EntityCapabilities* capabilities = stanza_create_caps_from_query_element(query);
 
-            // Update window name
-            ProfMucWin* win = wins_get_muc(from);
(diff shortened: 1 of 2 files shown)


jubalh:
Do I see it right that now it will work when *we* change the room config. But what if someone else changed the room config from another client?

MarcoPolo-PasTonMolo:
You are right, I didin't think of that. Hmm I'll need to think of another way to do this then

MarcoPolo-PasTonMolo:
Tried sending info when I get this stanza https://xmpp.org/extensions/xep-0045.html#roomconfig-notify

jubalh:
Hmm yes I think that might be a good solution. Did you test the 2 cases?

MarcoPolo-PasTonMolo:
Yes seems to work as intended. Configuration changes when I update it through profanity or through another client

jubalh (approved):

jubalh:
Thanks!

Issue #1687, opened 2022-04-04
[Question] How to set up your own avatar?


FoundOne:
I see you can see someone else's avatar, but is there a way to set your own avatar? I can't see how anywhere in the documentation.

If there is no way. Can I make it as a feature proposal?

jubalh:
Hi, that's right so far there is no way to set the avatar.

PR #1714, opened 2022-05-26
Add `/avatar set` command to publish avatar

MarcoPolo-PasTonMolo:
Use `/avatar set <path>` where <path> is an image file to upload a new
avatar for the current user. When the avatar is too big it gets scaled
down. Scaling code copied from dino.

Not so sure about using `gdk_pixbuf_save_to_buffer` because it's available after 2.4

Fixes https://github.com/profanity-im/profanity/issues/1687

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -9196,9 +9196,13 @@ cmd_avatar(ProfWin* window, const char* const command, gchar** args)
     }
 
     if (g_strcmp0(args[0], "set") == 0) {
+#ifdef HAVE_PIXBUF
         if (avatar_set(args[1])) {
             cons_show("Avatar updated successfully");
         }
+#else
+        cons_show("This version of Profanity has not been built with GDK Pixbuf support enabled");
+#endif
     } else if (g_strcmp0(args[0], "get") == 0) {
         avatar_get_by_nick(args[1], false);
     } else if (g_strcmp0(args[0], "open") == 0) {
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -9195,7 +9195,11 @@ cmd_avatar(ProfWin* window, const char* const command, gchar** args)
         return TRUE;
     }
 
-    if (g_strcmp0(args[0], "get") == 0) {
+    if (g_strcmp0(args[0], "set") == 0) {
+        if (avatar_set(args[1])) {
+            cons_show("Avatar updated successfully");
+        }
+    } else if (g_strcmp0(args[0], "get") == 0) {
         avatar_get_by_nick(args[1], false);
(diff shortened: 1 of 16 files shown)


src/command/cmd_funcs.c
```diff
@@ -9195,7 +9195,11 @@ cmd_avatar(ProfWin* window, const char* const command, gchar** args)
         return TRUE;
     }
 
-    if (g_strcmp0(args[0], "get") == 0) {
+    if (g_strcmp0(args[0], "set") == 0) {
+        if (avatar_set(args[1])) {
```
jubalh, line 9199: I would suggest to add the gdkpixbuf lib as an optional dependency.
At this point we should add an ifdef and in case Profanity is compiled without the pixbuf lib print some informative message (like done in OMEMO etc cases with ligsignal).

src/xmpp/avatar.c
```diff
@@ -92,6 +95,58 @@ avatar_pep_subscribe(void)
...
+        GdkPixbuf* new_pixbuf = gdk_pixbuf_scale_simple(pixbuf, dest_width, MAX_PIXEL, GDK_INTERP_BILINEAR);
+        g_object_unref(pixbuf);
+        pixbuf = new_pixbuf;
+    }
+
+    gchar* img_data;
+    gsize len = -1;
+
+    if (!gdk_pixbuf_save_to_buffer(pixbuf, &img_data, &len, "png", &err, NULL)) {
+        cons_show("Unable to scale and convert avatar.");
```
jubalh, line 132: cons_error (or was it cons_show_error?) might be more appropriate here.

jubalh (changes requested):
Excellent! Thanks for your contribution!
I skimmed over it and everything looks fine.

As you already noted in your comment the pixbuf library is currently not used AFAIK. If we add this new dependency I would keep it optional and make the `/avatar set` command also optional. Meaning if `/avatar set` is called and Profanity is not compiled with pixbuf then we should print some message like "Profanity needs to be compiled with nameoflib to upload an avatar".

MarcoPolo-PasTonMolo:
I have a problem. First of all I'm not at all confident for what I wrote in configure.ac, second checks don't compile. I can't fix it. I'll need some help

jubalh:
> I have a problem. First of all I'm not at all confident for what I wrote in configure.ac, second checks don't compile. I can't fix it. I'll need some help

Everything looks good. I can do the rest. I'll check everything later and try to fix the the remaining this.
Just a little short on time right now. Might take a few days :-)

Thanks for your work!

jubalh:
I don't see what's wrong with `expect`. @sjaeckel any idea?

jubalh:
Once this is merged we need to update https://profanity-im.github.io/guide/latest/build.html

Issue #1713, opened 2022-05-26
Segfault while using "/ox discover"


mdosch:
When doing "/ox discover certain_jid" profanity segfaults. Maybe the pubsub node for this certain jid is misconfigured. I will tell you the jid in private.

```
Program received signal SIGSEGV, Segmentation fault.
__strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:74
74	../sysdeps/x86_64/multiarch/strlen-avx2.S: Datei oder Verzeichnis nicht gefunden.
(gdb) bt full
#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:74
No locals.
#1  0x00005555555b1955 in _ox_metadata_result (stanza=0x55555ce959a0, userdata=0x0) at src/xmpp/ox.c:328
        fingerprint = 0x0
        pubsub = 0x55555c820d90
        items = 0x55555c48f9c0
        item = 0x55555c493d30
        publickeyslist = 0x55555cb62730
        pubkeymetadata = 0x55555ce954e0
#2  0x000055555559821e in _iq_handler (conn=0x555555b08380, stanza=0x55555ce959a0, userdata=0x555555b07540) at src/xmpp/iq.c:231
        keep = 21845
        handler = 0x55555ce921e0
        text = 0x555555af7650 "gG`\003PU"
        text_size = 469
        cont = 1
        type = 0x55555ce92170 "result"
        discoinfo = 0x0
        discoitems = 0x0
        lastactivity = 0x0
        version = 0x0
        ping = 0x0
        roster = 0x0
        blocking = 0x0
        id = 0x55555be09080 "8E5tofCn6OygNw5013b0c11be7398f3150c924067f0060d001a88f7"
#3  0x00007ffff7700379 in ?? () from target:/lib/x86_64-linux-gnu/libstrophe.so.0
No symbol table info available.
#4  0x00007ffff76fdb7c in ?? () from target:/lib/x86_64-linux-gnu/libstrophe.so.0
No symbol table info available.
#5  0x00007ffff770bdbf in ?? () from target:/lib/x86_64-linux-gnu/libstrophe.so.0
No symbol table info available.
[...]

PR #1715, opened 2022-05-26
Fix segfault on `/ox discover`

MarcoPolo-PasTonMolo:
`/ox discover` segfaults on some misconfigured? nodes because there are
newlines before and after some pubkey-metadata stanzas so the newlines
get treated as seperate stanzas. This commit just skips each stanza in
public-keys-list that doesn't have a fingerprint.

Fixes https://github.com/profanity-im/profanity/issues/1713


jubalh (approved):

jubalh:
Great! thanks!

Issue #757, opened 2016-03-12
Profanity ignores the nick sent to it


kousu:
If I join a MUC with profanity and the MUC gives me a different nick than I joined with, profanity says "you have joined the room as $requested_nick, role: $role, affiliation: member" in the room and possibly "Autojoined $room as $requested_nick".

The actual room roster doesn't list $requested_nick though, and messages come with the right nickname (because in MUC, all messages are echoed from the server, so the clients are not responible for choosing nicks there).

There are [one and a half ways](http://xmpp.org/extensions/xep-0045.html#enter-pres) a room can give you a different nick. First, in the `<status code=110>` stanza, it can give you a different nick than you requested. Second, if it gives you a different nick, it is supposed to attach `<status code=210>` as well, but that doesn't seem to be widely supported (I've only found Gajim paying attention to 210 so far) and I would be happy if you just declared the `<status code=110>` nick to be The Nick. You need to be careful about handling the case where $requested_nick matches a nick that is on the roster but is not the nick the room gives you.

Conversations had the same bug https://github.com/siacs/Conversations/issues/1732.

kousu:
I should emphasize: this happens when the nick is silently changed, not if you are rejected and asked to pick a different nick. The 210 feature doesn't make a lot of sense for plain MUCs, but it's key for transports.

MarcoPolo-PasTonMolo:
How do I reproduce this?

kousu:
Well, it's been a while, so I'm sorry for not having been more specific. But one way should be to use [biboumi](https://github.com/louiz/biboumi) to join an IRC room with a nickname with spaces -- biboumi should rewrite it to use underscores.

I can probably use aioxmpp to rig up a minimal transport that always corrupts your nickname.

MarcoPolo-PasTonMolo:
Ok I was able to replicate this. Let's hope I can also fix it

PR #1716, opened 2022-05-26
Respect silent nick change in mucs

MarcoPolo-PasTonMolo:
Profanity would ignore the silent nick change in some places. The roster
and history would show the correct nick, new messages from the current
user and the "Autojoined <jid> as <nick>" message would show the wrong
one. This commit fixes that problem.

Fixes https://github.com/profanity-im/profanity/issues/757


jubalh:
Thanks!

Issue #1595, opened 2021-09-02
OMEMO - Sending messages to myself, 'OMEMO not supported'


daeluk:
## Expected Behavior

Sending OMEMO encrypted messages to my own JID should only contain messages sent from any of my devices.

## Current Behavior

Currently, when sending messages to my own JID I get spurious "your Client doesn't support OMEMO" messages,
in addition to the original, encrypted (`~`)  message. This happens both with sent and received messages.
This doesn't happen in any other clients (tested with Conversations and Gajim). All of my device fingerprints are trusted.

## Steps to Reproduce (for bugs)

1. `/msg <own JID>`
2. `/omemo start`
3. `/omemo trust <known fingerprints>`
4. Write a message

Here is an example conversation:
```
23:20:00 ~ me: Hello from profanity
23:20:00 - me: You received a message encrypted with OMEMO but your client doesn't support OMEMO.
23:24:28 ~ me: Hello from Conversations
23:24:28 - me: I sent you an OMEMO encrypted message but your client doesn’t seem to support that. Find more information on https://conversations.im/omemo
```

## Context

Sometimes it's useful to be able to write a notes to yourself, e.g. to easily share text between phone and pc.

## Environment
* Give us the version and build information output generated by `profanity -v`
```
Profanity, version 0.11.0dev.master.c89e3126
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2021 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libmesode
Desktop notification support: Disabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.16.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.8.11)
GTK icons/clipboard: Enabled
```
* Operating System/Distribution
OpenBSD -current (7.0 beta)
* glib version
glib2-2.68.4

jubalh:
> All of my device fingerprints are trusted.

Can you please check your device key and see if it is indeed in the list of trusted keys? I mean: it could be that profanity lists a couple of keys but actually the key from the other client is not listed there because it couldn't find it.

daeluk:
[...]

PR #1717, opened 2022-05-29
Fix duplicate messages in chat with oneself.

MarcoPolo-PasTonMolo:
Messages would get duplicated when you chat with yourself, worse if you
had omemo enabled the duplicated message would say something along the
lines of "Your client doesn't support OMEMO". The cause was carbons
when the message was sent from another client, whilst it was a sent
and received message when profanity was the one to send it. This commit
ignores the carbon message in the 1st case and ignores the received one
in the 2nd.

Fixes https://github.com/profanity-im/profanity/issues/1595

### DIFF
--- a/src/ui/window.c
+++ b/src/ui/window.c
@@ -60,6 +60,7 @@
 #include "ui/screen.h"
 #include "xmpp/xmpp.h"
 #include "xmpp/roster_list.h"
+#include "xmpp/connection.h"
 
 #define CONS_WIN_TITLE "Profanity. Type /help for help information."
 #define XML_WIN_TITLE  "XML Console"
@@ -1192,7 +1193,10 @@ win_print_incoming(ProfWin* window, const char* const display_name_from, ProfMes
         if (prefs_get_boolean(PREF_CORRECTION_ALLOW) && message->replace_id) {
             _win_correct(window, message->plain, message->id, message->replace_id, message->from_jid->barejid);
         } else {
-            _win_printf(window, enc_char, 0, message->timestamp, flags, THEME_TEXT_THEM, display_name_from, message->from_jid->barejid, message->id, "%s", message->plain);
+            // Prevent duplicate messages when current client is sending a message
+            if (g_strcmp0(message->from_jid->fulljid, connection_get_fulljid()) != 0) {
+                _win_printf(window, enc_char, 0, message->timestamp, flags, THEME_TEXT_THEM, display_name_from, message->from_jid->barejid, message->id, "%s", message->plain);
+            }
         }
 
         free(enc_char);
--- a/src/xmpp/message.c
+++ b/src/xmpp/message.c
@@ -1314,6 +1314,14 @@ _handle_carbons(xmpp_stanza_t* const stanza)
         return NULL;
     }
 
+    // Eliminate duplicate messages in chat with oneself when another client is sending a message
+    char* bare_from = xmpp_jid_bare(connection_get_ctx(), xmpp_stanza_get_from(message_stanza));
+    if (g_strcmp0(bare_from, xmpp_stanza_get_to(message_stanza)) == 0) {
+        free(bare_from);
+        return NULL;
+    }
+    free(bare_from);
+
     return message_stanza;
 }
 


jubalh:
> Messages would get duplicated when you chat with yourself

What do you mean by that?

i write to myself in Profanity
I get the message once printed via outgoing
and once via incoming.

So that works.

I now tested to be online with Profanity and Dino. and writing to myself. Same as above happens.

I think that's what's expected?

MarcoPolo-PasTonMolo:
> I get the message once printed via outgoing
> and once via incoming.
we don't want that, each message should be printed once, if we use omemo the second message is always encrypted

> I tried your PR and I see no difference:
huh, weird. Could you send me the message stanzas you get in xml console when sending from profanity and another client?

jubalh:
> we don't want that

actually i think people expect that when they write to themsleves that they will see it 2 times (if writing from the same client). like some test about the connection. no?
Anybody knows how it looks like in other clients? Always when I did this I got the message twice?

MarcoPolo-PasTonMolo:
Yeah you are right. Some messaging app show the response some don't (for the same client). Most xmpp clients seem to not show it, for profanity I think we should also go with that approach since the second message is unencryptable if using OMEMO. Or we should at least find a way to unencrypt it first (I think someone said there's some limitation where clients can't encrypt for themselves?).
Does this pr work for you jubalh?

jubalh:
>  Does this pr work for you jubalh?

It does. But I feel like maybe I'm missing something.
Need to think a little about it (and didn't review the code yet), thinking about some sideeffects etc.
Or maybe other people (client devs?) can weigh in.

src/xmpp/message.c
```diff
@@ -258,6 +258,11 @@ _message_handler(xmpp_conn_t* const conn, xmpp_stanza_t* const stanza, void* con
             free(mybarejid);
         }
 
+        // Prevent duplicate messages when current client is sending a message
+        if (msg_stanza && g_strcmp0(xmpp_stanza_get_from(msg_stanza), xmpp_conn_get_jid(conn)) == 0) {
```
jubalh, line 262: This will prevent all further parsing of the stanza. Including `_handle_chat_states`

jubalh:
I did a quick test. Without OMEMO.

And when I sent a message to myself (from Profanity) and had receipts on I never got the indication (color change) that the message was received.

jubalh:
I'm really not sure if we want to throw away all stanzas that come from us. I'm not sure about all the possible use cases and XEPs. But it sounds a little risky to me.

Not displaying them might be another case. But I think we should probably process them anyways.
Still quite unsure about the overall effect of this change.

MarcoPolo-PasTonMolo:
I moved the check to where the messages get printed on screen.

jubalh:
That's much better I think.
I tested this with both unencrypted and OMEMO enabled. Works as expected. Just prints the messages one time. No OMEMO errors (like in above mentioned issue).
Only one nitpick left!

src/ui/window.c
```diff
@@ -1192,7 +1193,10 @@ win_print_incoming(ProfWin* window, const char* const display_name_from, ProfMes
         if (prefs_get_boolean(PREF_CORRECTION_ALLOW) && message->replace_id) {
             _win_correct(window, message->plain, message->id, message->replace_id, message->from_jid->barejid);
         } else {
-            _win_printf(window, enc_char, 0, message->timestamp, flags, THEME_TEXT_THEM, display_name_from, message->from_jid->barejid, message->id, "%s", message->plain);
+            // Prevent duplicate messages when current client is sending a message
+            if (g_strcmp0(message->from_jid->fulljid, xmpp_conn_get_jid(connection_get_conn())) != 0) {
```
jubalh, line 1197: Please use `const char* connection_get_fulljid();` instead

jubalh (changes requested):

src/ui/window.c
MarcoPolo-PasTonMolo: Done

sjaeckel:
IMO this patch should not be applied.

If it gets applied it should be hidden behind an option that is disabled by default.

jubalh:
Okay, let's do that for now.

Maybe over time we will get some feedback from users about this and then can still decide to remove the option or switch the default behaviour.

jubalh:
I have thought more about this and actually think the new behaviour is better. And that I just was too used to the old one.
I guess it is more modern/user friendly.

I then talked with @weiss about this and he agrees that it's a good behaviour.
He also mentioned that Conversations is doing the same.

So i think we should merge this as is (still not entirely sure if I miss something).
Or did you already start with implementing the preference @MarcoPolo-PasTonMolo ?

MarcoPolo-PasTonMolo:
> Or did you already start with implementing the preference @MarcoPolo-PasTonMolo ?
No

jubalh (approved):

jubalh:
Thanks for your contribution (and patience ;)) !

Issue #1320, opened 2020-04-20
Generate/show OMEMO verification barcode


mdosch:
It would be a handy feature to be able to show/export your [OMEMO verification QR code](https://github.com/omemo/gajim-omemo/issues/146). 

Maybe this could be tackled together with #1301.

jubalh:
> Maybe this could be tackled together with #1301.

Probably for black and white square ncurses is enough.

hashworks:
[libqrencode](https://github.com/fukuchi/libqrencode) could be used.

jubalh:
We could build this in profanity. But should it output a png with the QR code or display in terminal?

Also users could use xmppc: https://codeberg.org/Anoxinon_e.V./xmppc/wiki/Usage#omemo
```
xmppc -m omemo|while read line; do echo $line | qrencode -o $line.png ;done
```

jubalh:
General ability to display QR codes is merged.
Now we just need to find the format of the text to use in that barcode!
Anybody knows details?

sjaeckel:
I just checked with `zbarimg` and the qr code of conversations contains:

```
xmpp:<user@server>?omemo-sid-<numerical-sid>=<omemo-fingerprint-hex-string>
```

alexandre1985:
I have build profanity from source code.
My system is Artix Linux x86_64.

When I execute /omemo qrcode it says that this version has not been built with libqrencode, although I have that library installed.

How do I build profanity source code, with libqrencode support?

jubalh:
> How do I build profanity source code, with libqrencode support?

With `--enable-omemo-qrcode` passed to configure.

PR #1718, opened 2022-05-31
Use our omemo sid/fingerprint in qr code

jubalh:
Current clients sid/fingerprint will be shown in following format:
`xmpp:<user@server>?omemo-sid-<numerical-sid>=<omemo-fingerprint-hex-string>`

Fix https://github.com/profanity-im/profanity/issues/1320


sjaeckel (approved):
Generated QR code takes a while to scan, but it works.

PR #1720, opened 2022-06-15
remove VLAs in OMEMO QR Code function

binex-dsk:
Removes VLAs present in `cons_show_qrcode` in favor of calloc. Thanks @sjaeckel

### DIFF
--- a/src/ui/console.c
+++ b/src/ui/console.c
@@ -878,17 +878,21 @@ cons_show_qrcode(const char* const text)
 
     ProfWin* console = wins_get_console();
 
-    char buf[(width * 4) + 1];
-    memset(buf, 0, sizeof buf);
+    char *buf = calloc((width * 4) + 1, 1);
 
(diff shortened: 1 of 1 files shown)


jubalh:
```src/ui/console.c: In function ‘cons_show_qrcode’:
src/ui/console.c:906:31: error: argument to ‘sizeof’ in ‘memset’ call is the same expression as the destination; did you mean to provide an explicit length? [-Werror=sizeof-pointer-memaccess]
  906 |         memset(buf, 0, sizeof buf);
```

jubalh (changes requested):
src/ui/console.c:906:31: error: argument to ‘sizeof’ in ‘memset’ call is the same expression as the destination; did you mean to provide an explicit length? [-Werror=sizeof-pointer-memaccess]
  906 |         memset(buf, 0, sizeof buf);

jubalh (approved):

Issue #497, opened 2015-01-28
Empty argument "" becomes "


incertia:
This is an issue that came up with implementing the feature in #493. Supplying an empty argument with `'""'` somehow becomes `'"'` instead of `''` when passed to commands.

jubalh:
@mwuttke97 :)

PR #1721, opened 2022-06-16
Improve cmd argument parser

MarcoPolo-PasTonMolo:
"" used to become " now it just becomes an empty argument.
Also if quotes appeared after a token started then if the number of
quotes in the token is n the resulting one would be a token with the
n last characters cut off, now it's fixed.

Fixes https://github.com/profanity-im/profanity/issues/497


jubalh (approved):

jubalh:
Thanks a lot!

PR #1724, opened 2022-06-20
MAM improvements

MarcoPolo-PasTonMolo:
Started trying to finish MAM here's what I've done so far. Nowhere near production ready btw. Many temporary stuff in here.

* When we open new chat (`/msg someone`) if no msgs in db fetch using before: now. else fetch after most recent one.
* When scrolled to top, fetch older messages from db and display them above the previous ones.
* After all db history gets displayed when scrolling up, fetch older messages from mam (these ones get displayed after everything instead of before)

Todo:
* Only fetch once if no msgs for one chat (further fetching will be done when scrolling).
* MAM for mucs
* Have different pagination counters for each chat (currently one temporary global one)
* When scrolling up display new mam messages before other ones instead of after. (start of buffer instead of end)
* In case that the buffer fills up while scrolling upwards and messages get deleted from the end, handle scrolling down to the bottom.
* Check notifications cause I think older messages from db might get displayed in notification after scrolling instead of new incoming ones.
* Mam when we receive a message instead of `/msg someone`

### DIFF
--- a/src/database.c
+++ b/src/database.c
@@ -248,15 +248,15 @@ log_database_get_previous_chat(const gchar* const contact_barejid, char* start_t
     gchar* sort = !flip ? "ASC" : "DESC";
     GDateTime* now = g_date_time_new_now_local();
     gchar* end_date_fmt = end_time ? end_time : g_date_time_format_iso8601(now);
-    gchar* start_date_fmt = start_time ? start_time : NULL;
-    query = sqlite3_mprintf("SELECT * FROM (SELECT `message`, `timestamp`, `from_jid`, `type` from `ChatLogs` WHERE ((`from_jid` = '%q' AND `to_jid` = '%q') OR (`from_jid` = '%q' AND `to_jid` = '%q')) AND `timestamp` < '%q' AND (%Q IS NULL OR `timestamp` > %Q) ORDER BY `timestamp` DESC LIMIT %d) ORDER BY `timestamp` %s;", contact_barejid, myjid->barejid, myjid->barejid, contact_barejid, end_date_fmt, start_date_fmt, start_date_fmt, MESSAGES_TO_RETRIEVE, sort);
+    query = sqlite3_mprintf("SELECT * FROM (SELECT `message`, `timestamp`, `from_jid`, `type` from `ChatLogs` WHERE ((`from_jid` = '%q' AND `to_jid` = '%q') OR (`from_jid` = '%q' AND `to_jid` = '%q')) AND `timestamp` < '%q' AND (%Q IS NULL OR `timestamp` > %Q) ORDER BY `timestamp` %s LIMIT %d) ORDER BY `timestamp` %s;", contact_barejid, myjid->barejid, myjid->barejid, contact_barejid, end_date_fmt, start_time, start_time, sort, MESSAGES_TO_RETRIEVE, sort);
+
+    g_date_time_unref(now);
+    g_free(end_date_fmt);
+
     if (!query) {
         log_error("log_database_get_previous_chat(): SQL query. could not allocate memory");
         return NULL;
     }
-
-    g_date_time_unref(now);
-    g_free(end_date_fmt);
 
     jid_destroy(myjid);
 
--- a/src/database.c
+++ b/src/database.c
@@ -233,9 +233,11 @@ log_database_get_limits_info(const gchar* const contact_barejid, gboolean is_las
     return msg;
 }
 
-// Query previous chat
+// Query previous chats, constraints start_time and end_time. If end_time is
+// null the current time is used. from_start gets first few messages if true
+// otherwise the last ones. Flip flips the order of the results
 GSList*
-log_database_get_previous_chat(const gchar* const contact_barejid, char* start_time, char* end_time, gboolean flip)
+log_database_get_previous_chat(const gchar* const contact_barejid, char* start_time, char* end_time, gboolean from_start, gboolean flip)
 {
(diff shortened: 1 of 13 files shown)


MarcoPolo-PasTonMolo:
Yes I read those first before starting.
> load messages from DB when

I already imlemented this in the commit

jubalh:
> Yes I read those first before starting.

Awesome!

> I already imlemented this in the commit

I'm aware. What I meant to say was that it's great that you did that!

MarcoPolo-PasTonMolo:
Todo

* initial mam doesn't display(right after we /msg someone) until we scroll
* move the print at start or end logic away from _win_printf
* In case that the buffer fills up while scrolling upwards and messages get deleted from the end, handle scrolling down to the bottom.
* MAM for mucs
* Check notifications cause I think older messages from db might get displayed in notification after scrolling instead of new incoming ones.
* Mam when we receive a message instead of /msg someone

Done
* Only fetch once if no msgs for one chat (further fetching will be done when scrolling).
* Have different pagination counters for each chat (currently one temporary global one) -> no counters we check first messages timestamp in buffer
* Correctly use the before and after attributes
* Complete rsm paging
* When scrolling up display new mam messages before other ones instead of after. (start of buffer instead of end)

MarcoPolo-PasTonMolo:
Todo:
* Reenable regular _chawin_history when needed
* Add the loader when initial mam is fetching (to prevent user of loading older mam whilst initial latest mam still in progress)
* In case that the buffer fills up while scrolling upwards and messages get deleted from the end, handle scrolling down to the bottom.
* MAM for mucs
* Check notifications cause I think older messages from db might get displayed in notification after scrolling instead of new incoming ones.
* Mam when we receive a message instead of /msg someone
* remove cons_show debug messages
* Resolve conflicts

src/database.c
```diff
@@ -244,15 +245,17 @@ log_database_get_previous_chat(const gchar* const contact_barejid, GDateTime* en
         return NULL;
 
     // Flip order when querying older pages
-    char* sort = !flip ? "ASC" : "DESC";
-    gchar* date_fmt = g_date_time_format_iso8601(end_time ? end_time : g_date_time_new_now_local());
-    query = sqlite3_mprintf("SELECT * FROM (SELECT `message`, `timestamp`, `from_jid`, `type` from `ChatLogs` WHERE ((`from_jid` = '%q' AND `to_jid` = '%q') OR (`from_jid` = '%q' AND `to_jid` = '%q')) AND `timestamp` < '%q' ORDER BY `timestamp` DESC LIMIT 10) ORDER BY `timestamp` %s;", contact_barejid, myjid->barejid, myjid->barejid, contact_barejid, date_fmt, sort);
+    gchar* sort = !flip ? "ASC" : "DESC";
+    gchar* end_date_fmt = end_time ? end_time : g_date_time_format_iso8601(g_date_time_new_now_local());
```
jubalh, line 249: I think the return value of g_date_time_new_now_local() needs to be unref ed if I remember correctly. So I think you need a var for this.

MarcoPolo-PasTonMolo: Ok

src/database.c
```diff
@@ -246,14 +246,16 @@ log_database_get_previous_chat(const gchar* const contact_barejid, char* start_t
...
+    GDateTime* now = g_date_time_new_now_local();
+    gchar* end_date_fmt = end_time ? end_time : g_date_time_format_iso8601(now);
     gchar* start_date_fmt = start_time ? start_time : NULL;
     query = sqlite3_mprintf("SELECT * FROM (SELECT `message`, `timestamp`, `from_jid`, `type` from `ChatLogs` WHERE ((`from_jid` = '%q' AND `to_jid` = '%q') OR (`from_jid` = '%q' AND `to_jid` = '%q')) AND `timestamp` < '%q' AND (%Q IS NULL OR `timestamp` > %Q) ORDER BY `timestamp` DESC LIMIT %d) ORDER BY `timestamp` %s;", contact_barejid, myjid->barejid, myjid->barejid, contact_barejid, end_date_fmt, start_date_fmt, start_date_fmt, MESSAGES_TO_RETRIEVE, sort);
     if (!query) {
         log_error("log_database_get_previous_chat(): SQL query. could not allocate memory");
         return NULL;
     }
 
+    g_date_time_unref(now);
```
jubalh, line 258: I think you can free this directly after `g_date_time_format_iso8601()`. Or at least before the `if !(query)` otherwise we leak this in the error case. (Though that error case is about allocating memory and should almost never happen..)

MarcoPolo-PasTonMolo:
Also should I disable notifications for MAM messages?

MarcoPolo-PasTonMolo:
Todo:
* MAM for mucs
* MAM for private wins
* Correct way to log corrections in db
* See what to do about notifs
* Check if url and quotes autocompletion works fine
* Check if the api still works fine
* Resolve conflicts
Now only the boring stuff is left

jubalh:
> MAM for private wins

I'm not sure if we need this. I think in anonymous MUCs we can't be sure if the nick is from the same JID anyways.

jubalh:
> Also should I disable notifications for MAM messages?

I'd say yes.
In an optimal scenario we would only disable notifications for "read" messages. But I think that's too much for now. We still need to implement message markers afaik.

jubalh:
Thanks a lot for your work @MarcoPolo-PasTonMolo !!
I tested it a bit and it seems to work fine. Maybe we need some delay so the user cant hold PG UP and make the client super slow.

It's merged on master now, so people can give feedback.
MUC MAM can be done late.

PR #1730, opened 2022-06-29
Don't scroll if not needed

jubalh:
If we are in a window with a lot of text and press PAGE UP we scroll up
and write [scrolled] in the titlebar.

So far we also wrote [scrolled] in there even when actually nothing
happened. Like when opening a new window (/msg someone) and there is no
text inside.


MarcoPolo-PasTonMolo (approved):
Looks good

PR #1736, opened 2022-07-28
Fix /autoaway command logic

netboy3:
Two issues were fixed in the parser logic:
* A call to cons_bad_cmd_usage() was placed at the end of the "time"
  parser section that blocked reachability to both "message" and
  "check" parser sections. This caused "/autoaway message ..." and
  "/autoaway check ..." to always fail with "Invalid usage". This
  issue was introduced in commit 3c1e4ba.
* "/autoaway message xa" with no message argument returns message
  set to (null). This should be fixed the same way as
  "/autoaway message away" was fixed in commit 3c1e4ba.

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -6677,9 +6677,6 @@ cmd_autoaway(ProfWin* window, const char* const command, gchar** args)
             cons_bad_cmd_usage(command);
             return TRUE;
         }
-    } else {
-        cons_bad_cmd_usage(command);
-        return TRUE;
     }
 
     if (g_strcmp0(args[0], "message") == 0) {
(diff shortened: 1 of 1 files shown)


src/command/cmd_funcs.c
```diff
@@ -6712,6 +6709,9 @@ cmd_autoaway(ProfWin* window, const char* const command, gchar** args)
     if (g_strcmp0(args[0], "check") == 0) {
         _cmd_set_boolean_preference(args[1], command, "Online check", PREF_AUTOAWAY_CHECK);
         return TRUE;
+    } else {
```
jubalh, line 6712: I would remove this else block and just add `cons_bad_cmd_usage..` before the `return TRUE` on line 6717.

jubalh:
Looks good, thank you!
Made one small optional comment.

src/command/cmd_funcs.c
netboy3: Good idea. Just forced pushed your suggestion.

jubalh (approved):

Issue #1742, opened 2022-08-02
Setting avatar executable to gio open crashes Profanity


jubalh:
It works with url open but doesnt for avatar.

## Expected Behavior
`/executable avatar "gio open"` should set `gio open` as default opener for `/avatar open`.

## Current Behavior
Crash

## Possible Solution
Check GSubprocess instead of spawn.

## Steps to Reproduce (for bugs)
1. `/executable urlopen set "gio open %u"`
2. `/url open <tab> ..select one
3. it works

1. `/executable avatar "gio open"`"`
2. `/avatar open somejid`
3. crash

## Environment
https://github.com/profanity-im/profanity/commit/9d094f73e39b18367fab814be26efabe3c468d28

jubalh:
Thanks @nandesu-utils !

PR #1748, opened 2022-08-24
Fix for avatar opening executable

nandesu-utils:
In issue #1742 the reason behind crashing is supposedly undefined behavior because it was observed by me, `g_spawn_sync` may not modify `spawn_error` pointer. When I was debugging, it was pretty random and its value didn't fall into any of mapped memory regions.

`spawn_error` pointer was not null and code tried to read its field but this caused segfault instantly. I believe there was no message because for some reason glib did not set it to some value (there was an error!).

Though the basis for that is because mentioned `gio open` was considered to be one argument, it was failing to find such program and hence `g_spawn_sync` returned an error to that.

So this fixes undefined behaviour for that pointer and moreover made both pointers null by default just in case. Not sure why glib does not set the pointer value.

Also I made the `avatar open` use the preference string as a command template rather than an argv argument.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -445,9 +445,10 @@ get_mentions(gboolean whole_word, gboolean case_sensitive, const char* const mes
 gboolean
 call_external(gchar** argv, gchar** std_out, gchar** std_err)
 {
-    GError *spawn_error, *status_error;
-    gboolean spawn_result;
-    gint exit_status;
+    GError *spawn_error = NULL;
+    GError *exit_error = NULL;
+    gboolean is_successful;
+    gint wait_status;
 
     GSpawnFlags flags = G_SPAWN_SEARCH_PATH;
     if (std_out == NULL)
@@ -455,30 +456,34 @@ call_external(gchar** argv, gchar** std_out, gchar** std_err)
     if (std_err == NULL)
         flags |= G_SPAWN_STDERR_TO_DEV_NULL;
 
-    spawn_result = g_spawn_sync(NULL, // Inherit the parent PWD.
+    is_successful = g_spawn_sync(NULL, // Inherit the parent PWD.
                                 argv,
                                 NULL, // Inherit the parent environment.
                                 flags,
                                 NULL, NULL, // No func. before exec() in child.
                                 std_out, std_err,
-                                &exit_status, &spawn_error);
+                                &wait_status, &spawn_error);
 
-    if (!spawn_result || !g_spawn_check_exit_status(exit_status, &status_error)) {
+    if (!is_successful) {
         gchar* cmd = g_strjoinv(" ", argv);
-        if (spawn_error && spawn_error->message) {
-            log_error("Spawning '%s' failed with '%s'.", cmd, spawn_error->message);
-            g_error_free(spawn_error);
-        } else if (status_error && status_error->message) {
-            log_error("Spawning '%s' failed with '%s'.", cmd, status_error->message);
-            g_error_free(status_error);
-            spawn_result = FALSE;
-        } else {
-            log_error("Spawning '%s' failed with.", cmd);
-        }
+        log_error("could not spawn '%s' with error '%s'", cmd, spawn_error->message);
+
+        g_error_free(spawn_error);
         g_free(cmd);
     }
+    else {
+        is_successful = g_spawn_check_wait_status(wait_status, &exit_error);
 
-    return spawn_result;
+        if (!is_successful) {
+            gchar* cmd = g_strjoinv(" ", argv);
+            log_error("'%s' exited with error '%s'", cmd, exit_error->message);
+
+            g_error_free(exit_error);
+            g_free(cmd);
+        }
+    }
+
+    return is_successful;
 }
 
 gchar**
--- a/src/common.c
+++ b/src/common.c
@@ -445,8 +445,8 @@ get_mentions(gboolean whole_word, gboolean case_sensitive, const char* const mes
 gboolean
 call_external(gchar** argv, gchar** std_out, gchar** std_err)
 {
-    GError *spawn_error = NULL;
-    GError *exit_error = NULL;
+    GError* spawn_error = NULL;
+    GError* exit_error = NULL;
     gboolean is_successful;
     gint wait_status;
 
@@ -457,22 +457,22 @@ call_external(gchar** argv, gchar** std_out, gchar** std_err)
         flags |= G_SPAWN_STDERR_TO_DEV_NULL;
 
     is_successful = g_spawn_sync(NULL, // Inherit the parent PWD.
-                                argv,
-                                NULL, // Inherit the parent environment.
-                                flags,
-                                NULL, NULL, // No func. before exec() in child.
-                                std_out, std_err,
-                                &wait_status, &spawn_error);
+                                 argv,
+                                 NULL, // Inherit the parent environment.
+                                 flags,
+                                 NULL, NULL, // No func. before exec() in child.
+                                 std_out, std_err,
+                                 &wait_status, &spawn_error);
 
     if (!is_successful) {
         gchar* cmd = g_strjoinv(" ", argv);
-        log_error("could not spawn '%s' with error '%s'", cmd, spawn_error->message);
+        log_error("Spawning '%s' failed with with error '%s'", cmd, spawn_error ? spawn_error->message : "Unknown, spawn_error is NULL");
+        ;
 
         g_error_free(spawn_error);
         g_free(cmd);
-    }
-    else {
-        is_successful = g_spawn_check_wait_status(wait_status, &exit_error);
+    } else {
+        is_successful = g_spawn_check_exit_status(wait_status, &exit_error);
 
         if (!is_successful) {
             gchar* cmd = g_strjoinv(" ", argv);
--- a/src/xmpp/avatar.c
+++ b/src/xmpp/avatar.c
@@ -339,10 +339,20 @@ _avatar_request_item_result_handler(xmpp_stanza_t* const stanza, void* const use
 
     // if we shall open it
     if (g_hash_table_contains(shall_open, from_attr)) {
-        gchar* argv[] = { prefs_get_string(PREF_AVATAR_CMD), filename->str, NULL };
-        if (!call_external(argv, NULL, NULL)) {
-            cons_show_error("Unable to display avatar: check the logs for more information.");
+        gchar* cmdtemplate = prefs_get_string(PREF_AVATAR_CMD);
+
+        if (cmdtemplate == NULL) {
+            cons_show_error("No default `avatar open` command found in executables preferences.");
+        } else {
+            gchar** argv = format_call_external_argv(cmdtemplate, NULL, filename->str);
+
+            if (!call_external(argv, NULL, NULL)) {
+                cons_show_error("Unable to display avatar: check the logs for more information.");
+            }
+
+            g_strfreev(argv);
         }
+
         g_hash_table_remove(shall_open, from_attr);
     }
 


src/xmpp/avatar.c
```diff
@@ -339,10 +339,21 @@ _avatar_request_item_result_handler(xmpp_stanza_t* const stanza, void* const use
...
+            cons_show_error("No default `avatar open` command found in executables preferences.");
+        }
+        else {
+            gchar **argv = format_call_external_argv(cmdtemplate, NULL, filename->str);
+
+            if (!call_external(argv, NULL, NULL)) {
+                cons_show_error("Unable to display avatar: check the logs for more information.");
+            }
+
+            g_strfreev(argv);
```
sjaeckel, line 354:
```suggestion
        gchar* argv[] = { prefs_get_string(PREF_AVATAR_CMD), filename->str, NULL };
        if (argv[0] == NULL)
            cons_show_error("No default `avatar open` command found in executables preferences.");
        else if (!call_external(argv, NULL, NULL)) {
            cons_show_error("Unable to display avatar: check the logs for more information.");
```

I like the extended check but TBH I'd prefer to keep the old style and simply do the check as suggested

sjaeckel (commented):
:+1: 

good catch

sjaeckel:
> `spawn_error` pointer was not null and code tried to read its field but this caused segfault instantly. I believe there was no message because for some reason glib did not set it to some value (there was an error!).

That's strange and sounds like a bug inside glib, since [the documentation clearly states](https://docs.gtk.org/glib/func.spawn_sync.html):

> `error` [...]
> The argument will be set to NULL by the function if there are no errors.
> In case of error, the argument will be set and the caller will take ownership of the data, and be responsible for freeing it."

src/xmpp/avatar.c
nandesu-utils: I mean this one as will render it impossible to specify "gio open %p" because making it first argument in an array makes `gio open` an entire one argument. It doesn't push `open` as second one. `format_call_external_argv` on the other hand handles spaces.

I think that was the reason why the issue owner had the error initially which then coincidentally resulted in a crash due to another reason (undefined behavior). `gio open` was just not found as a program.

nandesu-utils:
> shows that glib simply returns an error w/o setting `error` if `argv[0]` is `NULL`, therefor the check before calling `call_external()` should be sufficient to fix this issue.

The exit status was not zero (or spawn result, sorry I don't remember). The code was like `255` or something.

Hence this is not about `argv[0]` being null. See, `gio open` was as the first argument and `g_spawn_sync` failed to find the executable file (because there was no executable in the `PATH` named `gio open`, it was not just `gio`, see the batch comment). I think this is a bug, the exit code (wait code) is not zero due to inability to find the executable file.

Just tested out with setting the executable to plain `chromium` and it went fine without crashing.

src/xmpp/avatar.c
nandesu-utils: Though the failed code style check made me cry a little bit 😢

nandesu-utils:
Okay I've spent some time with unfamiliar glib and read through many a page.

In ac96a2d8781fbe87a3ecbf28619825b877cf0316 I refactored the `call_external` inner body which now corresponds to what I found.

Also note: **this may indicate a bug in glib**. I am not sure, but in fact `g_spawn_sync` **does not** set the error pointer to null when there is no error accordingly to [this](https://docs.gtk.org/glib/func.spawn_sync.html). Moreover, it is completely messed up when there is an error but the error pointer is not null initially, before it is set somewhere in the glib code. Can somebody reproduce? I'll be thankful if I'm wrong. I've looked through glib code but didn't understand much of it right away.

So far, when there is a spawn error, we print it. Spawn error means that we could not just spawn a task for some reason. Otherwise we examine whether the process which was spawned has exited abnormally i.e. returned a non-zero exit code. If it has, we just print an exit error.

Call external function returns true only and only if process was spawned successfully and exited successfully as well, obviously.

In before there were unnecessary checks for `error->message` and basically it would not print an error if process exited with a non-zero code. It is very unlikely that a non-zero code may indicate success so it is now printed as well.

src/xmpp/avatar.c
sjaeckel: > I mean this one as will render it impossible to specify "gio open %p" because making it first argument in an array makes gio open an entire one argument. It doesn't push open as second one. format_call_external_argv on the other hand handles spaces.

Oh alright, that makes a lot of sense!

sjaeckel:
> Also note: **this may indicate a bug in glib**

From reading the sources this is a bug. glib makes a promise that it doesn't hold. Will you open a bug with glib or should I? Fixing it isn't that straight-forward, since one would have to touch several places, so it should IMO first be discussed with the glib team how it should be fixed... Whether to fix the code or the documentation ;)

Could you please squash those commits together into one commit with a good commit message? ... and then force-push :)

src/common.c
```diff
@@ -445,40 +445,44 @@ get_mentions(gboolean whole_word, gboolean case_sensitive, const char* const mes
...
-            log_error("Spawning '%s' failed with '%s'.", cmd, spawn_error->message);
-            g_error_free(spawn_error);
-        } else if (status_error && status_error->message) {
-            log_error("Spawning '%s' failed with '%s'.", cmd, status_error->message);
-            g_error_free(status_error);
-            spawn_result = FALSE;
-        } else {
-            log_error("Spawning '%s' failed with.", cmd);
-        }
+        log_error("could not spawn '%s' with error '%s'", cmd, spawn_error->message);
```
sjaeckel, line 469: you're potentially de-referencing a `NULL`-pointer here
```suggestion
        log_error("could not spawn '%s' with error '%s'", cmd, spawn_error ? spawn_error->message : "Unknown, spawn_error is NULL");
```

src/common.c
```diff
@@ -445,40 +445,44 @@ get_mentions(gboolean whole_word, gboolean case_sensitive, const char* const mes
...
+        is_successful = g_spawn_check_wait_status(wait_status, &exit_error);
+
+        if (!is_successful) {
+            gchar* cmd = g_strjoinv(" ", argv);
+            log_error("'%s' exited with error '%s'", cmd, exit_error->message);
+
+            g_error_free(exit_error);
+            g_free(cmd);
+        }
     }
```
sjaeckel, line 483: `g_error_free()` has to be called in every case, `exit_error` will always be set when calling `g_spawn_check_wait_status()`

```suggestion
            g_free(cmd);
        }
        g_error_free(exit_error);
    }
```

sjaeckel (changes requested):

src/common.c
```diff
@@ -445,40 +445,44 @@ get_mentions(gboolean whole_word, gboolean case_sensitive, const char* const mes
...
+        g_error_free(spawn_error);
         g_free(cmd);
+    } else {
+        is_successful = g_spawn_check_wait_status(wait_status, &exit_error);
+
+        if (!is_successful) {
+            gchar* cmd = g_strjoinv(" ", argv);
+            log_error("'%s' exited with error '%s'", cmd, exit_error->message);
+
+            g_error_free(exit_error);
```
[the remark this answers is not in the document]
nandesu-utils: Is that just a clarity thing or?

According to [this](https://docs.gtk.org/glib/func.spawn_check_wait_status.html#return-value) the error won't be set when process exited successfully thus moving `g_error_free(exit_error)` out of `if (!is_successful) { ... }` block is unnecessary (there is a null check within g_error_free anyway so that is okay).

src/common.c
nandesu-utils: If that has to deal with that `g_spawn_sync` has weird behavior about setting errors then OK. Otherwise `is_successful == FALSE` leads to `spawn_error` to be set to some value hence `spawn_error` is not null.

src/common.c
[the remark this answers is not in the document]
sjaeckel: You're right (and the documentation as well :) ), I missed the happy path when looking at the implementation and thought it will always set the error

It's fine as is.

src/common.c
sjaeckel: Yes, that's to deal with the weird behavior of `g_spawn_sync`.

nandesu-utils: I believe we better submit a bug report for glib because the problem is that `spawn_error` stays uninitialized no matter what, or is set to a valid error otherwise. I haven't observed anything else except with a case when `spawn_error` is uninitialized in the first place.

Maybe we'll better add a custom error log message indicating there's something weird with glib instead of just doing `spawn_error ? spawn_error->message : "Unknown, spawn_error is NULL"`?

nandesu-utils:
> From reading the sources this is a bug. glib makes a promise that it doesn't hold. Will you open a bug with glib or should I? Fixing it isn't that straight-forward, since one would have to touch several places, so it should IMO first be discussed with the glib team how it should be fixed... Whether to fix the code or the documentation ;)

I'd ask you to file a report because I'm sure I'm going to miss some information.

> Could you please squash those commits together into one commit with a good commit message? ... and then force-push :)

Yep, of course, but beforehand I'd like to resolve whether we should actually not use a ternary operator there (see comments).

src/common.c
sjaeckel: IMO that'd be already a "custom error log message" since that will only pop up once this exact issue in glib is hit.

@jubalh you're the official maintainer, what do you think?

jubalh: I would prefer `"Unknown, spawn_error is NULL"` compared to a `its glibs fault!`. I think this error is clear enough, telling us whats going on and easy to grep for in code.

sjaeckel:
> I'd ask you to file a report [...]

https://gitlab.gnome.org/GNOME/glib/-/issues/2732

> [...] because I'm sure I'm going to miss some information.

I hope I didn't miss any ;)

src/common.c
```diff
@@ -445,40 +445,44 @@ get_mentions(gboolean whole_word, gboolean case_sensitive, const char* const mes
 gboolean
 call_external(gchar** argv, gchar** std_out, gchar** std_err)
 {
-    GError *spawn_error, *status_error;
-    gboolean spawn_result;
-    gint exit_status;
+    GError* spawn_error = NULL;
+    GError* exit_error = NULL;
+    gboolean is_successful;
```
jubalh, line 450: this is a little too generic for my taste. If you don't like `spawn_result` what do you think about something like `spawn_succesful`?

jubalh:
@nandesu-utils when you squash/clean up the commits. Could you please add a note about the glib thing to the corresponding commit? Including the link to https://gitlab.gnome.org/GNOME/glib/-/issues/2732

sjaeckel:
GLib issues is closed! Lesson learned: 

> if you are passing a GError as an out argument, you’re supposed to set it to NULL: https://docs.gtk.org/glib/error-reporting.html

src/common.c
nandesu-utils: > this is a little too generic for my taste. If you don't like `spawn_result` what do you think about something like `spawn_succesful`?

I reused the same variable for whether the spawn was successful and also whether the exit status was OK or not.

Otherwise we'd introduce two boolean variables `spawn_successful` and `exited_successfully` respectively with an `&&` at the return statement for clarity.

Or just rework the code further making it explicitly return in places like:

```c
if (spawn_successful) {
    if (g_check_spawn_wait_status(...)) {
        return TRUE;
    } else {
        ...;
        return FALSE;
    }
} else {
    ...;
    return FALSE;
}
```

I don't know honestly. This seems like over-thinking to me as well.

nandesu-utils:
> GLib issue is closed! Lesson learned:
> 
> > if you are passing a GError as an out argument, you’re supposed to set it to NULL: https://docs.gtk.org/glib/error-reporting.html

When-you-read-the-docs-but-actually-they-are-quite-sparse-and-misleading-because-they-are-inconsistent-and-barely-contain-some-information-on-details™️

I wonder if there is something that neglects setting error pointers to null in the codebase at the moment? I am quite unfamiliar with it.

src/common.c
jubalh: yeah true its not a big deal. you can also leave it.

jubalh:
> Also I wonder if there is something that neglects setting error pointers to null in the codebase at the moment? I am quite unfamiliar with it.

After reading the comments here I also thought about this.
But a quick `ack --cc GError` brought:
```
src/command/cmd_funcs.c
7228:        GError* error = NULL;
7237:        GError* error = NULL;

src/config/cafile.c
74:    GError* glib_error = NULL;

src/config/preferences.c
88:    GError* err = NULL;

src/omemo/omemo.c
148:    GError* error = NULL;
606:    GError* error = NULL;
622:    GError* error = NULL;
638:    GError* error = NULL;
648:    GError* error = NULL;
1498:    GError* error = NULL;

src/pgp/gpg.c
190:        GError* gerr = NULL;

src/pgp/ox.c
456:    GError* error = NULL;

src/plugins/plugins.c
181:    GError* error = NULL;

src/tools/editor.c
58:    GError* glib_error = NULL;

src/ui/notifier.c
218:        GError* error = NULL;

src/ui/tray.c
85:    GError* err = NULL;

src/xmpp/avatar.c
105:    GError* err = NULL;
331:    GError* err = NULL;

src/main.c
92:    GError* error = NULL;

src/common.c
88:    GError* error = NULL;
448:    GError *spawn_error, *status_error;
```

So funny enough only this one time we don't do this :)

src/common.c
sjaeckel: This is missing

sjaeckel (changes requested):
LGTM after the re-opened discussion has been taken care of

nandesu-utils:
I think that change request is broken.

jubalh:
> also now code uses `g_spawn_check_wait_status` which
`g_spawn_check_exit_status` has been aliased to.

g_spawn_check_wait_status is only available since glib 2.70. So far we require only 2.62.
Last I checked most distros still had a version < 2.70 but not sure about the current situation.

nandesu-utils:
> > also now code uses `g_spawn_check_wait_status` which
> > `g_spawn_check_exit_status` has been aliased to.
> 
> g_spawn_check_wait_status is only available since glib 2.70. So far we require only 2.62. Last I checked most distros still had a version < 2.70 but not sure about the current situation.

OK! Will change back.

jubalh:
@nandesu-utils ping :)

src/common.c
jubalh: I'll do this now

jubalh (approved):
i squashed the last two commits together and also changed the error message according to our previous  review. all good now!

jubalh:
@nandesu-utils thanks for your contribution!

PR #1752, opened 2022-09-15
typo in configure: enable_gdk_pixbuf not enable_pixbuf

omar-polo:
Otherwise gdk-pixbuf is always added as dependency even if disabled, when found.


jubalh (approved):

jubalh:
Thanks!

PR #1753, opened 2022-09-15
Fix typo in configure.ac

wahjava:
This is similar to #1752. If `xscreensaver` is installed then port can't be built without xscreensaver support due to the typo.

Thanks!


jubalh (approved):

jubalh:
Thanks!

PR #1755, opened 2022-09-18
Fix plugins managment

vinegret:
* The `/plugins uninstall` command  <__without plugin name__>  behave same as `/quit`. I'm not sure that it was so conceived.  dd1a9a1e in this case  displays a hint, without exit.
 * For the `/plugins` command  <__without args__> 45fa60e7 add to output the list plugins which have already been installed, but  unloaded  (If there are such).

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -7247,14 +7247,16 @@ cmd_plugins(ProfWin* window, const char* const command, gchar** args)
         const gchar* filename;
         cons_show("The following Python plugins are available globally and can be installed:");
         while ((filename = g_dir_read_name(global_pyp_dir))) {
-            cons_show("  %s", filename);
+            if (g_str_has_suffix(filename, ".py"))
+                cons_show("  %s", filename);
         }
     }
     if (global_cp_dir) {
         const gchar* filename;
         cons_show("The following C plugins are available globally and can be installed:");
         while ((filename = g_dir_read_name(global_cp_dir))) {
-            cons_show("  %s", filename);
+            if (g_str_has_suffix(filename, ".so"))
+                cons_show("  %s", filename);
         }
     }
 
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -7258,14 +7258,25 @@ cmd_plugins(ProfWin* window, const char* const command, gchar** args)
         }
     }
 
+    GSList* unloaded_plugins = plugins_unloaded_list();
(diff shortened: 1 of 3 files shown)


src/command/cmd_funcs.c
```diff
@@ -7117,7 +7117,8 @@ gboolean
 cmd_plugins_uninstall(ProfWin* window, const char* const command, gchar** args)
 {
     if (args[1] == NULL) {
-        return FALSE;
+        cons_show("Please specify plugin name, see /help plugins");
```
jubalh, line 7120: We usually use `cons_bad_cmd_usage(command);` for this. See other functions as an example.

src/command/cmd_funcs.c
```diff
@@ -7258,14 +7258,25 @@ cmd_plugins(ProfWin* window, const char* const command, gchar** args)
         }
     }
 
+    GSList* unloaded_plugins = plugins_unloaded_list();
```
jubalh, line 7261: `/plugins` should list all loaded plugins and also all the plugins that are available on the global installation path.
When did you get into the situation that you need to print the plugins like you did here?
When they are in `.local/share/profanity/plugins` but unloaded by hand?

jubalh (changes requested):

src/command/cmd_funcs.c
vinegret: fixed

src/command/cmd_funcs.c
vinegret: Yes. All plugins can be loaded only locally (from `$XDG_DATA_HOME/profanity/plugins/`) , as far as I understood. From other paths they must be installed to `$XDG_DATA_HOME/profanity/plugins/`. The only difference between determined global path and custom path - the ability to skip the full path when installing:
```shell
/plugins install /tmp/profanity-plugins/stable/say.py  # install  and load locally
/plugins install say.py   # same as /plugins install GLOBAL_PYTHON_PLUGINS_PATH/say.py
/plugins install pid.so   # same as  /plugins install GLOBAL_C_PLUGINS_PATH/pid.so
```

Current message is misleading:
https://github.com/profanity-im/profanity/blob/fcc20628d5a01c3d1c7a26fca48d79631eaac93d/src/command/cmd_funcs.c#L7267
Prints only loaded plugins. Getting information about unloaded (but still installed) plugins could also be useful.

jubalh (approved):

jubalh:
Thank you!

PR #1756, opened 2022-09-22
plugin: python: fix build for Python 3.11

sgn:
From Python 3.11, PyFrameObject has been changed into opaque struct. We need to access those fields via API.

Signed-off-by: Đoàn Trần Công Danh <congdanhqx@gmail.com>


jubalh (approved):
I'll trust you on that one :)

jubalh:
Thanks!

PR #1760, opened 2022-09-26
Replace `call_external`'s implementation with an async one

### LINKED ISSUE #1759: Opening a URL or avatar freezes Profanity for as long as the opened executable is currently running
## Expected Behavior

Being able to use profanity while the opened executable is running

## Current Behavior

Profanity freezes until I kill the opened executable

## Possible Solution

Replace `g_spawn_sync` with `g_spawn_async` in `call_external`

1. Have `/url open` set to open a browser, for example like Firefox or Chromium
2. Close any existing browsers
3. Open a URL in profanity using `/url open`

## Context

This is noticable, if for example, `/url open` opens a new browser

## Environment
* Artix Linux (OpenRC)
* glib:  2.72.3
```
Profanity, version 0.13.0dev.feature/1158-vcard.2e84acce
Copyright (C) 2019 - 2022 Michael Vetter <jubalh@iodoru.org>.

Build information:
XMPP library: libstrophe
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.10.7)
GTK icons/clipboard: Enabled
GDK Pixbuf: Enabled
```

techmetx11:
The original `call_external` is kept as `call_external_sync`

Fixes #1759

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -443,44 +443,25 @@ get_mentions(gboolean whole_word, gboolean case_sensitive, const char* const mes
 }
 
 gboolean
-call_external(gchar** argv, gchar** std_out, gchar** std_err)
+call_external(gchar** argv)
 {
-    GError* spawn_error = NULL;
-    GError* exit_error = NULL;
+    GError* spawn_error;
     gboolean is_successful;
-    gint wait_status;
 
-    GSpawnFlags flags = G_SPAWN_SEARCH_PATH;
-    if (std_out == NULL)
-        flags |= G_SPAWN_STDOUT_TO_DEV_NULL;
-    if (std_err == NULL)
-        flags |= G_SPAWN_STDERR_TO_DEV_NULL;
+    GSpawnFlags flags = G_SPAWN_SEARCH_PATH | G_SPAWN_STDOUT_TO_DEV_NULL | G_SPAWN_STDERR_TO_DEV_NULL;
 
-    is_successful = g_spawn_sync(NULL, // Inherit the parent PWD.
-                                 argv,
-                                 NULL, // Inherit the parent environment.
-                                 flags,
-                                 NULL, NULL, // No func. before exec() in child.
-                                 std_out, std_err,
-                                 &wait_status, &spawn_error);
-
+    is_successful = g_spawn_async(NULL, // Inherit the parent PWD
+                                  argv,
+                                  NULL, // Inherit the parent environment
+                                  flags,
+                                  NULL, NULL, NULL,
+                                  &spawn_error);
     if (!is_successful) {
         gchar* cmd = g_strjoinv(" ", argv);
-        log_error("Spawning '%s' failed with with error '%s'", cmd, spawn_error ? spawn_error->message : "Unknown, spawn_error is NULL");
-        ;
+        log_error("Spawning '%s' failed with error '%s'", cmd, spawn_error ? spawn_error->message : "Unknown, spawn_error is NULL");
 
         g_error_free(spawn_error);
         g_free(cmd);
-    } else {
-        is_successful = g_spawn_check_exit_status(wait_status, &exit_error);
-
-        if (!is_successful) {
-            gchar* cmd = g_strjoinv(" ", argv);
-            log_error("'%s' exited with error '%s'", cmd, exit_error->message);
-
-            g_error_free(exit_error);
-            g_free(cmd);
-        }
     }
 
     return is_successful;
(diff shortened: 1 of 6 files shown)


jubalh:
I don't quite see where `call_external_sync()` is used/needed now?

techmetx11:
Me neither, I left it there for compatibility, but maybe it'll be better if I remove it

jubalh:
We are no library :)
So don't need to worry about compatibility in the regard of not renaming/removing functions.

I'll have to check the code later. Did you test if all the situations that use this function now still work?

techmetx11:
> We are no library :) So don't need to worry about compatibility in the regard of not renaming/removing functions.
> 
> I'll have to check the code later. Did you test if all the situations that use this function now still work?

I've only tested it with `/url open` right now

jubalh (changes requested):
If all the functions that use `call_external` work you can actually remove `call_external_sync`, no need to keep it.

techmetx11:
> If all the functions that use `call_external` work you can actually remove `call_external_sync`, no need to keep it.

I removed it now :3

jubalh:
Did you also test the functionality which calls this function?

And since it's now two commits about one change you can squash it into one.

techmetx11:
> Did you also test the functionality which calls this function?
> 
> And since it's now two commits about one change you can squash it into one.

I've tried it out with `/avatar open`, `/url open` (both HTTP and AESGCM) and `/url save`
I'll squash it

src/common.c
```diff
@@ -443,44 +443,24 @@ get_mentions(gboolean whole_word, gboolean case_sensitive, const char* const mes
...
     if (!is_successful) {
         gchar* cmd = g_strjoinv(" ", argv);
-        log_error("Spawning '%s' failed with with error '%s'", cmd, spawn_error ? spawn_error->message : "Unknown, spawn_error is NULL");
-        ;
+        log_error("Spawning '%s' failed with error '%s'", cmd, spawn_error ? spawn_error->message : "Unknown, spawn_error is NULL");
 
         g_error_free(spawn_error);
         g_free(cmd);
-    } else {
-        is_successful = g_spawn_check_exit_status(wait_status, &exit_error);
```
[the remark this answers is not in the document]
techmetx11: You could setup some child watch with the child PID

src/common.c
```diff
@@ -443,44 +443,35 @@ get_mentions(gboolean whole_word, gboolean case_sensitive, const char* const mes
 }
 
 gboolean
-call_external(gchar** argv, gchar** std_out, gchar** std_err)
+call_external(gchar** argv, gint* std_out, gint* std_in, gint* std_error, GPid* pid)
```
sjaeckel, line 446: What's the point of adding those two parameters and then not using them?

IMO the better change here would be to remove those two parameters that are unused.

src/common.c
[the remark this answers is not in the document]
sjaeckel: Do we really care what happened after the process was spawned? I think it can stay like this.

sjaeckel (changes requested):
Looks acceptable to me.

Please also squash the commits into one.

jubalh (changes requested):
Please see @sjaeckel s comments.
After this we can merge :)

src/common.c
techmetx11: Should I be removing the 4 parameters in the new `call_external`?

jubalh: Yes, it looks like they are not used anywhere.

techmetx11: Sure :3

src/common.c
```diff
@@ -443,44 +443,26 @@ get_mentions(gboolean whole_word, gboolean case_sensitive, const char* const mes
...
-                                 argv,
-                                 NULL, // Inherit the parent environment.
-                                 flags,
-                                 NULL, NULL, // No func. before exec() in child.
-                                 std_out, std_err,
-                                 &wait_status, &spawn_error);
 
+    GSpawnFlags flags = G_SPAWN_SEARCH_PATH | G_SPAWN_STDOUT_TO_DEV_NULL | G_SPAWN_STDERR_TO_DEV_NULL;
+
+    is_successful = g_spawn_async_with_pipes(NULL, // Inherit the parent PWD
```
jubalh, line 453: Why `g_spawn_async_with_pipes()` and not `g_spawn_async()` ?

sjaeckel: Because before there were all those arguments, but now they're not used anymore, so the slimmer API `g_spawn_async()` could be used ... which under the hood still calls `g_spawn_async_with_pipes()` https://docs.gtk.org/glib/func.spawn_async.html ... :D

jubalh: i'll change it.

jubalh (approved):

jubalh:
Thanks for your contribution @techmetx11 !

PR #1765, opened 2022-10-09
Add some necessary changes 

ragu-1997:
rm -rf is the command unfortunately missed the -rf in previous commit so, sometime file is not removed. safer side rm -rf is necessary  
Shell file final exit code is necessary (exit 0). we miss the exit code some time it is running the background

### DIFF
--- a/.builds/openbsd.yml
+++ b/.builds/openbsd.yml
@@ -22,6 +22,9 @@ packages:
   - libgcrypt
   - libsignal-protocol-c
   - sqlite3
+  - wget
+  - nohup
+  
 
 sources:
   - https://github.com/strophe/libstrophe.git#0.11.0
--- a/apidocs/c/gen.sh
+++ b/apidocs/c/gen.sh
@@ -1,1 +1,2 @@
 rm -rf html && doxygen c-prof.conf
+exit 0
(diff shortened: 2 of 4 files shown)


jubalh:
I saw now that you edited the text here on the GitHub comment.

Please squash all commits together into one and add the explanation there as well. Like mentioned in the contributing.md. 

About your comments:
i don't see what you mean about `rm -rf` you don't change anything in this regard.

Also i think you didn't explain the nohup and wget changes.

.builds/openbsd.yml
```diff
@@ -22,6 +22,9 @@ packages:
   - libgcrypt
   - libsignal-protocol-c
   - sqlite3
+  - wget
+  - nohup
```
sjaeckel, line 26: Where is this file even used?

If you change it, make it used in CI, otherwise simply remove it IMO

sjaeckel (changes requested):
The three `exit 0` are borderline nonsense, I'd say they are even harmful.

The default shell behavior is that the exit code of a shell script is equal to the exit code of the last executed process/command. Those changes make these scripts simply always succeed, whereas before their exit code was appropriately the exit code of the thing that should be done.

ragu-1997:
Sorry for the misunderstanding Thanks for your explanation sir

PR #1766, opened 2022-10-09
package

ragu-1997:
Net-tools is the networking tools.
That is most important why because verify what are the ports are running inside the container (use this command netstat -ntlp)
so, we easily avoid some networking vulnerability.
**Use case:
monitor the networking process and ports**

### DIFF
--- a/Dockerfile.ubuntu
+++ b/Dockerfile.ubuntu
@@ -30,7 +30,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
   python-dev-is-python3 \
   libsqlite3-dev \
   libgdk-pixbuf-2.0-dev \
-  libqrencode-dev
+  libqrencode-dev \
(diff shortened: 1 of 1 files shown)


jubalh:
I saw now that you edited here on GitHub.

But the final commit later also needs to contain the information and also needs to follow the guidelines mentioned in CONTRIBUTING.md.

The Dockerfile you are editing is for running our CI. So I'm slightly puzzled what your change should do.

jubalh:
Still waiting for your explanation/answer.

ragu-1997:
Sorry for the late reply Micheael

Unfortunately i misunderstood the code and scenario you are wright thanks
for your time and effort

jubalh:
No problem. Keep up investigating, maybe you find something to improve!

ragu-1997:
Sure

PR #1772, opened 2022-10-22
Fix segfault when loading from MAM

MarcoPolo-PasTonMolo:
When loading messages from MAM profanity would segfault. Reason was that we were freeing the timestamp of messages when displaying them and we needed it for loading MAM.


jubalh (approved):

jubalh:
Thanks!

PR #1773, opened 2022-10-22
Improve MAM loading string.

mdosch:
Replaced three dots `...` by a proper ellipsis `…`.
Removed plenking (space between `messages` and ellipsis.


jubalh:
Thanks!

PR #1780, opened 2022-12-05
Minor improvements

sjaeckel:
This PR adds a new `/strophe` command that allows the user to modify strophe-internal settings.

Currently there's support for manipulation of the Stream-Management feature and the verbosity of Debug output.

This also 

* modifies the log command to be less strict regarding capitalization when accepting user input.
* fixes some memory leaks
* introduces the usage of auto-free'ing of strings via a `auto_char` resp. `auto_gchar` macro
* improves the `command_defs[]` struct
* adds a `now` option to the reconnect command
* changes the API of `account_new()` to do less memory allocations

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -8436,18 +8436,19 @@ _cmd_execute(ProfWin* window, const char* const command, const char* const inp)
             int i = 0;
             while (cmd->sub_funcs[i].cmd) {
                 if (g_strcmp0(args[0], (char*)cmd->sub_funcs[i].cmd) == 0) {
-                    gboolean result = cmd->sub_funcs[i].func(window, command, args);
-                    g_strfreev(args);
-                    return result;
+                    result = cmd->sub_funcs[i].func(window, command, args);
+                    goto out;
                 }
                 i++;
             }
         }
         if (!cmd->func) {
             ui_invalid_command_usage(cmd->cmd, cmd->setting_func);
-            return TRUE;
+            result = TRUE;
+            goto out;
         }
-        gboolean result = cmd->func(window, command, args);
+        result = cmd->func(window, command, args);
+out:
         g_strfreev(args);
         return result;
     } else if (plugins_run_command(inp)) {
@@ -8887,6 +8888,7 @@ cmd_omemo_fingerprint(ProfWin* window, const char* const command, gchar** args)
         free(formatted_fingerprint);
     }
 
+    jid_destroy(jid);
     g_list_free(fingerprints);
 
     win_println(window, THEME_DEFAULT, "-", "You can trust it with '/omemo trust <fingerprint>'");
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -9857,6 +9857,41 @@ cmd_mood(ProfWin* window, const char* const command, gchar** args)
 }
 
 gboolean
+cmd_strophe(ProfWin* window, const char* const command, gchar** args)
+{
+    if (g_strcmp0(args[0], "verbosity") == 0) {
+        int verbosity;
+        auto_gchar gchar* err_msg = NULL;
+        if (string_to_verbosity(args[1], &verbosity, &err_msg)) {
+            xmpp_ctx_set_verbosity(connection_get_ctx(), verbosity);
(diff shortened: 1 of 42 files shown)


src/command/cmd_funcs.c
```diff
@@ -8580,31 +8579,24 @@ _update_presence(const resource_presence_t resource_presence,
...
 {
     if (arg == NULL) {
         cons_bad_cmd_usage(command);
-    } else if (strcmp(arg, "on") == 0) {
-        GString* enabled = g_string_new(display);
-        g_string_append(enabled, " enabled.");
-
-        cons_show(enabled->str);
+        return FALSE;
+    } else if (g_strcasecmp(arg, "on") == 0) {
```
jubalh, line 8589: Usually we use case sensitive comparison. This one is fine for me as well. Just thinking whether it might be strange if it works in some cases case sensitive and in others not.

src/ui/console.c
```diff
@@ -2311,6 +2311,7 @@ cons_show_connection_prefs(void)
     cons_autoping_setting();
     cons_autoconnect_setting();
     cons_rooms_cache_setting();
+    cons_strophe_setting();
```
jubalh, line 2314: Hehe Just wanted to write "I think this should go into `cons_show_connection_prefs()` or `cons_prefs()`" but after a refresh it was done already :)

jubalh (commented):
Nice feature! And some good catches in your changes as well! Thanks!
Only thing I see is that autocompletion for `/stophe` command is missing.

src/command/cmd_funcs.c
sjaeckel: > Usually we use case sensitive comparison

why? I can understand case sensitivity for some things, but ON vs. On vs. on vs. oN shouldn't make a difference ... this is als true for other settings, but that's the first one where I've changed it.

IMO Postel's law is very important when it comes to usability of software.

jubalh: I'm not saying we SHOULD use case sensitive comparison.
I'm only saying that so far we do. And I would find it weird to now change this in only *one* occasion :)
I think it should be uniform.

From the top of my head I don't remember whether it actually is important/needed that we do that. So we could change it everywhere to be case insensitive. But I don't actually see any advantage it that. It might be that we miss something where it would make sense to be case sensitive. And then we have to change everything back. Plus it would just create work.

Those to reasons are why I think it it should be changed. The main reason is to be uniform though.

And I also see:
```
g_strcasecmp has been deprecated since version 2.2 and should not be used in newly-written code.
See g_strncasecmp() for a discussion of why this function is deprecated and how to replace it.
```

```
The problem with g_strncasecmp() is that it does the comparison by calling toupper()/tolower(). These functions are locale-specific and operate on single bytes...
````

File: src/command/cmd_funcs.c [hunk 3]
```diff
@@ -9856,6 +9851,41 @@ cmd_mood(ProfWin* window, const char* const command, gchar** args)
     return TRUE;
 }
 
+gboolean
+cmd_strophe(ProfWin* window, const char* const command, gchar** args)
+{
+    if (g_strcasecmp(args[0], "verbosity") == 0) {
```
jubalh, line 9857: same as above (strcasecmp)

sjaeckel:
@thexhr as you're on OpenBSD, could you please check whether this branch is OK for you? I've introduced usage of `__attribute__((__cleanup__( )))` and I'm not 100% sure whether that's available for you!? It should be since it's a compiler extension, but I'd prefer to have it verified before merging and it breaks... Thanks!

thexhr:
@sjaeckel The branch builds fine on OpenBSD 7.2-current (amd64, clang 13.0.0) and I spot no regression so far.

jubalh:
`src/xmpp/message.c:532:19: error: unused variable 'jidp' [-Werror,-Wunused-variable]`

sjaeckel:
> `src/xmpp/message.c:532:19: error: unused variable 'jidp' [-Werror,-Wunused-variable]`

awesome! by removing/replacing code we find out that even more code can be removed ;)

jubalh (approved):

jubalh:
Looks great! Lots of improvements! Thanks!

Issue #1783, opened 2022-12-18
`/url save` Autocompletion doesn't autocomplete filenames


MarcoPolo-PasTonMolo:
`/url save` doesn't autocomplete filenames it only autocompletes the url. If I press tab while writing some filename the filename gets deleted and the link gets replaced by the next link in ac.

## Expected Behavior
`/url save` should autocomplete the url as it does now but after that is typed and a space is inserted it should autocomplete filenames/paths

## Current Behavior
`/url save` autocompletes only the url, anything else typed after gets deleted and the next url is shown

## Steps to Reproduce (for bugs)
1. Get url in your chat window
2. Type `/url save` press tab to get the url now start typing a location to save to, press tab
3. Observe

## Context
Makes `/url save` inconvenient to use.

## Environment
Profanity, version 0.13.1dev.fix/segfault-mam.135a4cd9
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2022 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.18.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.10.8)
GTK icons/clipboard: Enabled
GDK Pixbuf: Enabled

PR #1784, opened 2022-12-24
Make `/url save` autocomplete filenames after a url

MarcoPolo-PasTonMolo:
Previously after the url if you pressed tab, even if you typed out a filepath, profanity would erase that and cycle through url autocomplete results.

This patch solves that and autocompletes filepaths after the url.

Fixes https://github.com/profanity-im/profanity/issues/1783

### DIFF
--- a/src/command/cmd_ac.c
+++ b/src/command/cmd_ac.c
@@ -4352,7 +4352,21 @@ _url_autocomplete(ProfWin* window, const char* const input, gboolean previous)
             return result;
         }
 
-        result = autocomplete_param_with_func(input, "/url save", wins_get_url, previous, window);
+        gboolean arg_result;
+        gchar** args = parse_args(input, 1, 8, &arg_result);
+        gboolean space_at_end = g_str_has_suffix(input, " ");
+        int num_args = g_strv_length(args);
+
+        if (arg_result) {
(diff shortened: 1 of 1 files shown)


src/command/cmd_ac.c
```diff
@@ -4352,7 +4352,22 @@ _url_autocomplete(ProfWin* window, const char* const input, gboolean previous)
...
+        int num_args = g_strv_length(args);
+
+        if (arg_result) {
+            if ((num_args == 1 && space_at_end) || (num_args == 2 && !space_at_end)) {
+                result = autocomplete_param_with_func(input, "/url save", wins_get_url, previous, window);
+            } else if ((num_args == 2 && space_at_end) || (num_args == 3 && !space_at_end)) {
+                GString* beginning = g_string_new("/url save");
+                g_string_append_printf(beginning, " %s", args[1]);
+                result = cmd_ac_complete_filepath(input, beginning->str, previous);
+                g_string_free(beginning, TRUE);
```
sjaeckel, line 4367:
```suggestion
                gchar* cmd = g_strdup_printf("/url save %s", args[1]);
                result = cmd_ac_complete_filepath(input, cmd, previous);
                g_free(cmd);
```

`GString` is too fancy for that use case, also using `g_strdup_printf()` is IMO easier to read and a huge runtime overhead that will be avoided.

sjaeckel (approved):

PR #1787, opened 2023-01-11
Fix xscreensaver detection

jubalh:
In 28a9605a1 we migrated from AC_CHECK_LIB which defines HAVE_LIBXSS automatically. With pkg-config way you need it explicit. And also x11 is needed or else linking will fail missing XFree().

Patch provided by Paul Fertser and comitted by jubalh. Thanks Paul!

Fix https://github.com/profanity-im/profanity/issues/1695

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -273,9 +273,10 @@ AS_IF([test "x$PLATFORM" = xosx],
 
 dnl feature: xscreensaver
 AS_IF([test "x$with_xscreensaver" != xno],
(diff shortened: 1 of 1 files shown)


jubalh:
@lynix would you mind testing this?

lynix:
Tested and confirmed to be fixing my issue, awesome!

Linkage confirms as well:
```
lynix@pc1209:profanity$ ldd $(which profanity) | grep -i xss
	libXss.so.1 => /usr/lib/libXss.so.1 (0x00007fa80afe0000)
```

jubalh:
Great!

Issue #1770, opened 2022-10-21
Fix quoting after MAM merge


jubalh:
`wins_add_quotes_ac()` is called in `chatwin_incoming_msg()` and thus run only upon incoming new messages.
After merging https://github.com/profanity-im/profanity/pull/1724 we now send a MAM request and load new messages out of the DB and can scroll back.
Same should be done for `wins_add_urls_ac()` for url completion.

Maybe its worth it just to iterate backwards in the UI buffer?

PR #1789, opened 2023-02-13
Fix quote and url autocompletion for MAM and history

MarcoPolo-PasTonMolo:
After adding MAM quote and url autocompletion wouldn't pick up messages from MAM or history(from DB) or would have them in the wrong order. This commit fixes that.

Fixes https://github.com/profanity-im/profanity/issues/1770


jubalh:
Cool! Thanks :)

Issue #1790, opened 2023-02-13
MAM doesn't load some recent messages


MarcoPolo-PasTonMolo:
MAM doesn't load some recent messages. Something is wrong with the formatting of the date when requesting data from the db after MAM was fetched. After git bisecting I found out that the culprit is  e886ba0c6fed433b561c4d0efb4f91b0681aa788

## Current Behavior
Some recent messages are missing after opening a chat. Even though they exist in the db

## Possible Solution
I believe this might be an issue with wrongly formatted date when searching in the db. Dates in db are stored as gmt+02 but the searching date is gmt+00.

## Steps to Reproduce (for bugs)
1. Have some recent messages(within 2h for my case) in chat and MAM enabled
2. close and open a chat
3. Those recent messages don't appear

## Environment
* Latest master

PR #1791, opened 2023-02-13
Fix MAM not loading some recent messages

MarcoPolo-PasTonMolo:
Reverse a small change from commit e886ba0 that caused this issue.

Fixes https://github.com/profanity-im/profanity/issues/1790


sjaeckel (approved):
Sorry, thanks for bisecting this. I somehow missed the logic change I introduced ... :-/

Issue #1777, opened 2022-10-27
Usability: Make the message for how to trust own OMEMO fingerprint clearer


devhell:
I tried trusting my own fingerprint, following the instructions displayed to me after generating materials or checking my own fingerprint, the instructions, however, are not clear.

## Expected Behavior

The message correctly tells the user how to trust their own OMEMO fingerprint.

## Current Behavior

```
<... my own OMEMO fingerprint ...>
2022-10-27 19:49 - You can trust it with '/omemo trust <fingerprint>'
2022-10-27 19:49 - You can untrust it with '/omemo untrust <fingerprint>'
```
After trying the suggestion from above, I get the following error:

```
2022-10-27 19:49 - You must be in a regular chat window to trust a device without providing the contact.
```

Turns out, the correct way to trust one's own fingerprint is by including one's own jid, like so:

```
/omemo trust myjid@myserver.tld <fingerprint>
```

## Possible Solution

I realize that this is a simple fix, but right now I have zero time to implement it. If someone wants to take this on, be my guest, otherwise it'll be a while until I'm able to work on this. :/

## Context

I spent much longer than expected on trying to figure out why I can't trust my own fingerprint.

## Environment

```
Profanity, version 0.13.1
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2022 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.18.0-unknown)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.10.7)
GTK icons/clipboard: Enabled
GDK Pixbuf: Enabled
```

```
NixOS Unstable
glib: 2.74.0
```

miikkuu:
Hey! @devhell Would like to work on this... And also would like to know your suggestion on it as its my first PR.

devhell:
[...]

PR #1792, opened 2023-02-28
Fixed omemo trust prompt

JohnnyPoblano:
Fixes https://github.com/profanity-im/profanity/issues/1777


MarcoPolo-PasTonMolo:
This could make users think that to trust another person's fingerprint they would need to type their own jid. Maybe something like `You can untrust it with '/omemo trust <fingerprint>' or '/omemo trust <jid> <fingerprint>'` would be better?

jubalh:
I agree with @MarcoPolo-PasTonMolo .

And did you check whether your message is true in all cases? There are different window types and you need to compare the function.

Issue #1794, opened 2023-03-08
Download progress does not update when saving the same file twice


IsaacM88:
## Expected Behavior

When using "/url save" to download the same file twice within the same window, download progress should appear for both downloads.

## Current Behavior

When using "/url save" to download the same file twice within the same window, download progress appears for the first download, but subsequent downloads print progress to the first download. Subsequent downloads show 0% progress even when downloading is complete. See the attached image.

![GitHubIssue](https://user-images.githubusercontent.com/45052253/223593880-832270cc-e68b-4e3c-93e1-5f4afb87afca.png)

## Possible Solution

Assign a unique ID to each download. I have a pull request on the way.

## Steps to Reproduce (for bugs)

1. Use "/url save" to download a file.
2. Use "/url save" to download the same file again.

## Context

Sometimes I download the same file twice for reasons unknown to me.

## Environment
* Give us the version and build information output generated by `profanity -v`

Profanity, version 0.13.1
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2022 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.17.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.11.2)
GTK icons/clipboard: Disabled
GDK Pixbuf: Enabled

* If you could not yet build profanity, mention the revision you try to build from

Able to build from source.

* Operating System/Distribution

Fedora 37 Workstation.

* glib version

$ sudo dnf list --installed glib*
glib-networking.x86_64, 2.74.0-1.fc37
glib2.x86_64, 2.74.6-1.fc37
glib2-devel.x86_64, 2.74.6-1.fc37
glibc.x86_64, 2.36-9.fc37
glibc-all-langpacks.x86_64, 2.36-9.fc37
glibc-common.x86_64, 2.36-9.fc37
glibc-devel.x86_64, 2.36-9.fc37
glibc-gconv-extra.x86_64, 2.36-9.fc37
glibc-headers-x86.noarch, 2.36-9.fc37
[...]

PR #1795, opened 2023-03-08
Fix duplicate download IDs when using "/url save"

IsaacM88:
Fixes https://github.com/profanity-im/profanity/issues/1794

Explanation
The problem is the download's identifier. Downloads are given an ID so they can be referenced later when their progress changes. Currently, the download's ID is the download's URL. When you download the same file twice, you have two downloads with the same ID. Download progress updates are shown on the first of both downloads with the same ID.

Solution
Change the download's ID from its URL to a random number. There is a new function in cmd_funcs.c named "_url_make_id()" that generates a random number for a download's ID. Several other functions are updated to cope with the new ID format.

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -9399,12 +9399,13 @@ cmd_slashguard(ProfWin* window, const char* const command, gchar** args)
 
 #ifdef HAVE_OMEMO
 void
-_url_aesgcm_method(ProfWin* window, const char* cmd_template, const char* url, const char* filename)
+_url_aesgcm_method(ProfWin* window, const char* cmd_template, const char* url, const char* filename, const char* id)
 {
     AESGCMDownload* download = malloc(sizeof(AESGCMDownload));
     download->window = window;
     download->url = strdup(url);
     download->filename = strdup(filename);
+    download->id = strdup(id);
     if (cmd_template != NULL) {
         download->cmd_template = strdup(cmd_template);
     } else {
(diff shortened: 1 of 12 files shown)


src/command/cmd_funcs.c
```diff
@@ -9397,14 +9397,23 @@ cmd_slashguard(ProfWin* window, const char* const command, gchar** args)
     return TRUE;
 }
 
+gchar*
+_url_make_id()
+{
+    gchar* id = malloc(6 * sizeof(char));
```
jubalh, line 9403: If you need to allocate memory and call sprintf you should use g_strdup_printf() instead :)

src/command/cmd_funcs.c
```diff
@@ -9397,14 +9397,23 @@ cmd_slashguard(ProfWin* window, const char* const command, gchar** args)
...
+    sprintf(id, "%i", rand() % 9999);
```
jubalh, line 9404: BTW we have `get_random_string()` and maybe another helper function.

jubalh (changes requested):

src/config/color.c
```diff
@@ -59,7 +59,7 @@ static struct color_pair_cache
     struct
     {
         int16_t fg, bg;
-    } * pairs;
```
jubalh, line 62: you changed the coding style here for no reason :)
maybe some editor config?

jubalh:
A couple of git things:
You have merge commits in there. They shouldn't be there. You want to rebase on master not merge it into this branch.
Also on some commits you are `Your Name`. And on others you set your email properly and GH recognizes your login.

So I would suggest to remove the merge commits, squash your two commits into one and set your name/email as you want it to be displayed.

If you don't want to do that I can also cherry pick/squash myself.

Also you changed the coding style somewhere. Which makes our CI complain.
As mentioned in https://github.com/profanity-im/profanity/blob/master/CONTRIBUTING.md `clang-format` can help with that.

jubalh (changes requested):
See comments

jubalh:
Looks good now. Would you mind to expand the commit message?
Like use what you already have as the title. then have a new line. then paste what you wrote here on github already, your first comment. Which mentions the issue it fixes and explains how/why.
That would be great.

jubalh (approved):

PR #1796, opened 2023-03-10
Minor improvements

sjaeckel:
Check commit messages for details


jubalh (approved):
Everything looks great.
What I'm not sure about is `change OMEMO QR code 0 bits to 'space'`, but let's merge it and let a wider audience test it in different terminals/fonts/themes.

PR #1801, opened 2023-03-30
Editor executable support flags

alexandre1985:
* Make editor executable into a string to be able to support (multiple) flags.
 * Change /help executable to suit this new feature

It was not working when setting /executable editor to "emacsclient -t", because of the "-f" flag.
Now it does work with this setting.

### DIFF
--- a/src/command/cmd_defs.c
+++ b/src/command/cmd_defs.c
@@ -2545,7 +2545,7 @@ static const struct cmd_t command_defs[] = {
               "/executable urlsave set \"curl %u -o %p\"",
               "/executable urlsave default",
               "/executable vcard_photo set \"feh %p\"",
-              "/executable editor set vim")
+              "/executable editor set \"emacsclient -t\"")
     },
 
     { CMD_PREAMBLE("/url",
--- a/src/command/cmd_defs.c
+++ b/src/command/cmd_defs.c
@@ -2523,6 +2523,7 @@ static const struct cmd_t command_defs[] = {
               "/executable urlopen default",
               "/executable urlsave set <cmdtemplate>",
               "/executable urlsave default",
+              "/executable editor set <cmdtemplate>",
               "/executable vcard_photo set <cmdtemplate>",
               "/executable vcard_photo default")
       CMD_DESC(
--- a/src/tools/editor.c
+++ b/src/tools/editor.c
@@ -88,11 +88,16 @@ get_message_from_editor(gchar* message, gchar** returned_message)
     }
 
     char* editor = prefs_get_string(PREF_COMPOSE_EDITOR);
+    gchar* editor_with_filename = g_strdup_printf("%s %s", editor, filename);
+    gchar** editor_argv = g_strsplit(editor_with_filename, " ", 0);
+
+    g_free(editor_with_filename);
 
     // Fork / exec
     pid_t pid = fork();
     if (pid == 0) {
-        int x = execlp(editor, editor, filename, (char*)NULL);
+        int x = execvp(editor_argv[0], editor_argv);
+        g_strfreev(editor_argv);
         if (x == -1) {
             log_error("[Editor] Failed to exec %s", editor);
         }


alexandre1985:
this can be considered a fix. Sorry for the branch name, not having fix/ . Will do it next pull request.

jubalh:
Thanks for your PR! Will review soon.

> this can be considered a fix. Sorry for the branch name, not having fix/ . Will do it next pull request.

No worries, that's not mandatory.

jubalh (approved):

src/command/cmd_defs.c
```diff
@@ -2545,7 +2545,7 @@ static const struct cmd_t command_defs[] = {
               "/executable urlsave set \"curl %u -o %p\"",
               "/executable urlsave default",
               "/executable vcard_photo set \"feh %p\"",
-              "/executable editor set vim")
+              "/executable editor set \"emacsclient -t\"")
```
sjaeckel, line 2548:
```suggestion
              "/executable editor set <editor>",
              "/executable editor set \"<editor> -args\"")
```

let's maybe change this to a less opinionated/more generic example?

alexandre1985: sounds good.
I don't know if it will work an editor without being inside double quotes (such as `/executable editor set vim` instead of `/executable editor set "vim"`).

jubalh: I actually think as an example this is fine. I think all our (most?) examples are practical. See feh, firefox etc. We could add a generic part into `CMD_SYN` though.

jubalh: I actually do believe the `emacsclient -t` is what some emacs user will want. So I think it's a good example.
Sure just to show how to pass a flag via "" would be enough. But I don't see any harm in pointing them to the thing that they might actually want to be doing anyways.

But actually (not related to your changes) the `/executables editor` is missing anyways from `CMD_SYN`. Would you mind to add the generic way there?

alexandre1985:
sorry for closing.
I have renamed the branch locally and I did some git command to rename upstream. I though that it would just rename the branch upstream. Here is the continuation -> https://github.com/profanity-im/profanity/pull/1803

alexandre1985:
can do that.

jubalh:
> the reopen pull request button is grey

That is indeed weird.
Might be that you should have reopened first an then pushed to the branch name. Might be a GitHub thing here.
I wonder what would happen if you: close the new PR. Open this one. Push to this branch.

alexandre1985:
> > the reopen pull request button is grey
> 
> That is indeed weird. Might be that you should have reopened first an then pushed to the branch name. Might be a GitHub thing here. I wonder what would happen if you: close the new PR. Open this one. Push to this branch.

The thing is that it doesn't let me Reopen this pull request. As you can see by the amount of pull requests that I have made.

alexandre1985:
it says there is an already open pull request from alexandre1985:editor-with-arguments to profanity-im:master.

Only the comment button gets ungreyed.

jubalh:
> it says there is an already open pull request from alexandre1985:editor-with-arguments to profanity-im:master.

And it did this before you created 1804? (That's why I mentioned above about closing)

sjaeckel:
The problem is that IIRC regular users aren't allowed to re-open PR's. I've just closed the new one and could re-open this one.

alexandre1985:
> The problem is that IIRC regular users aren't allowed to re-open PR's. I've just closed the new one and could re-open this one.

ok. I have made 2 commits because I believe they address 2 different things. That is why I did not amend/squash.

sjaeckel (approved):

alexandre1985:
can I close https://github.com/profanity-im/profanity/pull/1803 ?

jubalh:
Thanks for your contribution @alexandre1985 !

jubalh:
> for me too, until I closed the other PR that referenced the same source branch

ahh! That makes sense :)

PR #1802, opened 2023-03-31
Security fix of OTR, PGP and other

H3rnand3zzz:
Fix of OTR in message.c: carbon in stanza breaks OTR and prevents it from starting and working properly

Fix of other encryption methods by calling correct function that would initiate them in case if someone writes.

PGP Bug Description
If someone writes using PGP and chat window was closed even if PGP was enabled before, PGP won't initiate, even though it should. The problem is in calling the wrong function to initiate window. It can make a user send a message which he expect to be encrypted in plain text, thus making a big security problem.

OTR Bug Description
If user tries to initiate OTR, it does not work. E.g. PSI+ will just print out the messages and other profanity client won't start the OTR session either. The problem is in the carbons inside of the message.

Minor formatting.

### DIFF
--- a/src/xmpp/message.c
+++ b/src/xmpp/message.c
@@ -598,10 +598,6 @@ message_send_chat_otr(const char* const barejid, const char* const msg, gboolean
         stanza_attach_state(ctx, message, state);
     }
 
-    stanza_attach_carbons_private(ctx, message);
-    stanza_attach_hints_no_copy(ctx, message);
-    stanza_attach_hints_no_store(ctx, message);
-
     if (request_receipt) {
         stanza_attach_receipt_request(ctx, message);
     }
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -645,8 +645,8 @@ sv_ev_incoming_message(ProfMessage* message)
     chatwin = wins_get_chat(looking_for_jid);
 
     if (!chatwin) {
-        ProfWin* window = wins_new_chat(looking_for_jid);
-        chatwin = (ProfChatWin*)window;
+        chatwin = chatwin_new(looking_for_jid);
+        ProfWin* window = (ProfWin*)chatwin;
         new_win = TRUE;
 
         if (prefs_get_boolean(PREF_MAM)) {
@@ -689,8 +689,7 @@ sv_ev_incoming_carbon(ProfMessage* message)
     gboolean new_win = FALSE;
     ProfChatWin* chatwin = wins_get_chat(message->from_jid->barejid);
     if (!chatwin) {
-        ProfWin* window = wins_new_chat(message->from_jid->barejid);
-        chatwin = (ProfChatWin*)window;
+        chatwin = chatwin_new(message->from_jid->barejid);
         new_win = TRUE;
 
 #ifdef HAVE_OMEMO
--- a/src/otr/otr.c
+++ b/src/otr/otr.c
@@ -151,7 +151,7 @@ cb_gone_secure(void* opdata, ConnContext* context)
 {
     ProfChatWin* chatwin = wins_get_chat(context->username);
     if (!chatwin) {
-        chatwin = (ProfChatWin*)wins_new_chat(context->username);
+        chatwin = chatwin_new(context->username);
     }
 
     chatwin_otr_secured(chatwin, otr_is_trusted(context->username));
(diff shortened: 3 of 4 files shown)


src/config/color.c
```diff
@@ -59,7 +59,7 @@ static struct color_pair_cache
     struct
     {
         int16_t fg, bg;
-    } * pairs;
```
jubalh, line 62: For this please see https://github.com/profanity-im/profanity/pull/1774

src/xmpp/message.c
```diff
@@ -598,10 +598,6 @@ message_send_chat_otr(const char* const barejid, const char* const msg, gboolean
         stanza_attach_state(ctx, message, state);
     }
 
-    stanza_attach_carbons_private(ctx, message);
```
jubalh, line 601: Please move the OTR change in a separate commit. So that you have one commit per change that belongs together.

jubalh (changes requested):
Thanks for you patch! I didn't review it in detail yet. Only superficial. Please see the comments.

src/xmpp/message.c
H3rnand3zzz: Done.

src/config/color.c
H3rnand3zzz: Done.

src/xmpp/message.c
jubalh: Ah yes :) But now we have 1) the not optimal commit 2) the revert commit 3) one good commit 4) another good commit. And the last 2 we actually don't need in the history. You can drop the first two commits or ammend them. Using something like `git rease -i HEAD~5`. Maybe google the details. It's quite handy to know about it. Afterwards you will have to do a `git push -f` to force push it on this branch.

jubalh (changes requested):
Ah yes :) But now we have 1) the not optimal commit 2) the revert commit 3) one good commit 4) another good commit. And the last 2 we actually don't need in the history. You can drop the first two commits or ammend them. Using something like `git rease -i HEAD~5`. Maybe google the details. It's quite handy to know about it. Afterwards you will have to do a `git push -f` to force push it on this branch.

sjaeckel (changes requested):
Thanks for those patches!

Could you please rebase your changes and remove the first two commits (initial commit and its revert)?
How did you find those problems (if I may ask)?

After checking for other references of `wins_new_chat()` I've found that it's still used from `src/otr/otr.c:150 cb_gone_secure()` ... should this maybe also be changed to calling `chatwin_new()` instead?

H3rnand3zzz:
> Thanks for those patches!
> 
> Could you please rebase your changes and remove the first two commits (initial commit and its revert)? How did you find those problems (if I may ask)?
> 
> After checking for other references of `wins_new_chat()` I've found that it's still used from `src/otr/otr.c:150 cb_gone_secure()` ... should this maybe also be changed to calling `chatwin_new()` instead?

Thanks for the feedback. I found these problems when I was trying to communicate with other people using profanity. Fixed it.

H3rnand3zzz:
> Ah yes :) But now we have 1) the not optimal commit 2) the revert commit 3) one good commit 4) another good commit. And the last 2 we actually don't need in the history. You can drop the first two commits or ammend them. Using something like `git rebase -i HEAD~5`. Maybe google the details. It's quite handy to know about it. Afterwards you will have to do a `git push -f` to force push it on this branch.

Thank you for the feedback. I knew about this feature, but wasn't sure that "push -f" is a good practice. Now I learned that under certain circumstances, it can be used for the greater good.

sjaeckel (approved):

jubalh:
> but wasn't sure that "push -f" is a good practice

That's true if several people work on the same branch :)

jubalh (approved):

jubalh:
Thanks for your contribution @H3rnand3zzz !

sjaeckel:
> > but wasn't sure that "push -f" is a good practice
> 
> That's true if several people work on the same branch :)

FYI: Since some time there are the [--force-with-lease](https://git-scm.com/docs/git-push#Documentation/git-push.txt---no-force-with-lease) and [--force-if-includes](https://git-scm.com/docs/git-push#Documentation/git-push.txt---no-force-if-includes) options for `git push` which are supposed to prevent the most common mishaps when working together on the same branch.

Issue #1556, opened 2021-06-08
Message window is empty after reconnect


mdosch:
If a reconnect happens while a chat window (MUC or 1-1) is open the whole window (top bar, bottom bar, participant list, chat win etc.) is empty (only background color) after the reconnect. Then it gets redrawn but the chat win remains empty.

https://user-images.githubusercontent.com/5103003/121163534-2b569c00-c84f-11eb-9839-1e930bc71f74.mp4

## Expected Behavior
Show the messages in chat window after the reconnect.

## Current Behavior
See above.

## Steps to Reproduce (for bugs)

1. Have a chat window open
2. Force a reconnect (restart network manager or similar)
3. Look at the chat window
4. The message window disappears after the reconnect.

## Environment
* Debian Unstable
* i3 / Xorg
* GNOME Terminal 3.38.3 using VTE 0.62.3 +BIDI +GNUTLS +ICU +SYSTEMD
* libncursesw6/testing,unstable,now 6.2+20201114-2 amd64
* libreadline8/unstable,now 8.1-2 amd64

```
cat ~/.config/profanity/profrc
[connection]
defaccount=martin

account=martin

carbons=true

receipts.request=true
receipts.send=true

reveal.os=true

correction.allow=true

autoping=20
autoping.timeout=10

mam=true

[ui]
theme=solarized-dark
titlebar.position=1
mainwin.position=2
statusbar.position=3
inputwin.position=4

statuses.chat=none
history=true

adv.notify.discoversion=true

statusbar.show.number=true
statusbar.show.name=false
statuses.muc=none
statuses.console=none

beep=true
flash=false
splash=true
wrap=true
resource.title=false
resource.message=false
occupants=true
roster=true
roster.offline=true
roster.resource=false
roster.presence=false
roster.status=false
roster.empty=false
roster.wrap=true
roster.count.zero=false
roster.priority=false
roster.contacts=true
roster.unsubscribed=true
roster.rooms=true
privileges=true
presence=true
intype=true
enc.warn=true
tls.show=true
time.console=%d.%m.%y %H:%M:%S
time.chat=%d.%m.%y %H:%M:%S
time.muc=%d.%m.%y %H:%M:%S
time.config=%d.%m.%y %H:%M:%S
time.private=%d.%m.%y %H:%M:%S
time.xmlconsole=%d.%m.%y %H:%M:%S
time.statusbar=%H:%M:%S
console.muc=mention
console.private=first
console.chat=first
roster.by=group
[...]

PR #1808, opened 2023-04-02
Fix empty window after reconnect

MarcoPolo-PasTonMolo:
Sometimes after a reconnect the current window would get cleared. This was a deliberate change to fix the profanity window looking all garbled up after providing the passphrase for a gpg key using pinentry-curses.
See https://github.com/profanity-im/profanity/commit/6034b833be41909982b0d2bdac0a1d8499ad8e76

Fixes https://github.com/profanity-im/profanity/issues/1556

### DIFF
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -160,7 +160,7 @@ sv_ev_roster_received(void)
         // Redraw the screen after entry of the PGP secret key, but not init
         ProfWin* win = wins_get_current();
         char* theme = prefs_get_string(PREF_THEME);
-        win_clear(win);
+        win_redraw(win);
         theme_init(theme);
         g_free(theme);
         ui_resize();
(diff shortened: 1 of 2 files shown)


jubalh:
@mdosch

MarcoPolo-PasTonMolo:
Btw here's how I reproduced the bug:
1) Added a pgp.keyid=SOME PGP KEY  to my account in the profanity accounts file
2) open profanity you'll notice now that you have the pgp key at some point the welcome screen clears (you can't see the profanity logo)
3) /msg a contact
4) /xmlconsole
5) type `<` in xmlconsole and send it, now we have lost connectivity
6) got to the window of the contact, type `/reconnect now`
7) the window should clear.

Doing this after the fix shouldn't clear the window. And yes you do need to have a pgp.keyid to your account or this probably shouldn't happen

Issue #1810, opened 2023-04-03
Log file is flooded with "bad color name"


shahab-vahedi:
My log files are flooded with:

```
...
prof: ERR: Color: bad color name black_bold_default
prof: ERR: Unable to load colour theme
...
```

However, I'm not able to find a reference to `black_bold_default` anywhere:
```
$ grep -irl 'black_bold_default' /usr/share/profanity
$ grep -irl 'black_bold_default' ~/.config/profanity
$ grep -irl 'black_bold_default' ~/.local/share/profanity
  [ only the log files printing the errors are listed ]

$ for e in s S b B l L; do strings -e $e /usr/bin/profanity; done | grep -i 'black_bold'
```

This is my `profrc`
```
[ui]
theme=batman

roster.private=room
roster.rooms.server=true
roster.rooms=true

history=true

[connection]

defaccount=xxxxxx

[a@b.c]
notify.mention=true

[notifications]
message=true
message.text=true
room=false
message.current=true
room.current=true
room.text=true
room.mention=true
tray=true

tray.timer=10

[logging]
chlog=true
grlog=true

[omemo]
log=on
```

And finally, this is my profanity build:
```
Profanity, version 0.13.1
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2022 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.18.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.11.2)
GTK icons/clipboard: Enabled
GDK Pixbuf: Enabled
```

**EDIT** I can confirm that this happens if and only if the `batman` theme is chosen.

jubalh:
> I will submit a PR.

Great, thank you!

PR #1811, opened 2023-04-03
themes/batman: Use "bold_black" iso "black_bold"

shahab-vahedi:
This makes all the related errors in the log files go away:
  ...
  prof: ERR: Color: bad color name black_bold_default
  prof: ERR: Unable to load colour theme
  ...

Fixes https://github.com/profanity-im/profanity/issues/1810


jubalh (changes requested):
Same should be done for `unsubscribed`, right?

shahab-vahedi:
> Same should be done for `unsubscribed`, right?

Good catch! Now that you brought it up, I checked other theme files as well for the `*_bold` pattern. None was found.

jubalh (approved):

jubalh:
Thanks!

PR #1815, opened 2023-04-03
Feature: Allow setting client identification name/version manually

H3rnand3zzz:
This privacy enhancing feature allows to change client name and version for each account that user have.

### DIFF
--- a/src/command/cmd_ac.c
+++ b/src/command/cmd_ac.c
@@ -478,6 +478,7 @@ cmd_ac_init(void)
     autocomplete_add(account_set_ac, "otr");
     autocomplete_add(account_set_ac, "pgpkeyid");
     autocomplete_add(account_set_ac, "startscript");
+    autocomplete_add(account_set_ac, "clientid");
     autocomplete_add(account_set_ac, "tls");
     autocomplete_add(account_set_ac, "auth");
     autocomplete_add(account_set_ac, "theme");
@@ -490,6 +491,7 @@ cmd_ac_init(void)
     autocomplete_add(account_clear_ac, "otr");
     autocomplete_add(account_clear_ac, "pgpkeyid");
     autocomplete_add(account_clear_ac, "startscript");
+    autocomplete_add(account_clear_ac, "clientid");
     autocomplete_add(account_clear_ac, "theme");
     autocomplete_add(account_clear_ac, "muc");
     autocomplete_add(account_clear_ac, "resource");
--- a/src/command/cmd_defs.c
+++ b/src/command/cmd_defs.c
@@ -2063,6 +2063,7 @@ static const struct cmd_t command_defs[] = {
               "/account set <account> otr <policy>",
               "/account set <account> pgpkeyid <pgpkeyid>",
               "/account set <account> startscript <script>",
+              "/account set <account> clientid \"<name> <version>\"",
               "/account set <account> tls force|allow|trust|legacy|disable",
               "/account set <account> auth default|legacy",
               "/account set <account> theme <theme>",
@@ -2073,6 +2074,7 @@ static const struct cmd_t command_defs[] = {
               "/account clear <account> otr",
               "/account clear <account> pgpkeyid",
               "/account clear <account> startscript",
+              "/account clear <account> clientid",
               "/account clear <account> muc",
               "/account clear <account> resource")
       CMD_DESC(
@@ -2102,6 +2104,7 @@ static const struct cmd_t command_defs[] = {
               { "set <account> otr <policy>", "Override global OTR policy for this account, see /otr." },
               { "set <account> pgpkeyid <pgpkeyid>", "Set the ID of the PGP key for this account, see /pgp." },
               { "set <account> startscript <script>", "Set the script to execute after connecting." },
+              { "set <account> clientid \"<name> <version>\"", "[EXPERIMENTAL] Set XMPP client name according for discovery according to XEP-0092. Use with caution." },
               { "set <account> tls force", "Force TLS connection, and fail if one cannot be established, this is default behaviour." },
               { "set <account> tls allow", "Use TLS for the connection if it is available." },
               { "set <account> tls trust", "Force TLS connection and trust server's certificate." },
@@ -2117,6 +2120,7 @@ static const struct cmd_t command_defs[] = {
               { "clear <account> otr", "Remove the OTR policy setting for this account." },
               { "clear <account> pgpkeyid", "Remove pgpkeyid associated with this account." },
               { "clear <account> startscript", "Remove startscript associated with this account." },
+              { "clear <account> clientid", "Reset client's name to default." },
               { "clear <account> theme", "Clear the theme setting for the account, the global theme will be used." },
               { "clear <account> resource", "Remove the resource setting for this account." },
               { "clear <account> muc", "Remove the default MUC service setting." })
@@ -2129,6 +2133,7 @@ static const struct cmd_t command_defs[] = {
               "/account set me nick dennis",
               "/account set me status dnd",
               "/account set me dnd -1",
+              "/account set me clientid \"Profanity 0.42 (Dev)\"",
               "/account rename me chattyme",
               "/account clear me pgpkeyid")
     },
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -841,6 +841,14 @@ _account_set_startscript(char* account_name, char* script)
 }
 
 gboolean
+_account_set_client(char* account_name, char* new_client)
+{
+    accounts_set_client(account_name, new_client);
+    cons_show("Client name for account %s has been set to %s", account_name, new_client);
+    return TRUE;
+}
+
+gboolean
 _account_set_theme(char* account_name, char* theme)
 {
     if (!theme_exists(theme)) {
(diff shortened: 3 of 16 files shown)


src/xmpp/iq.c
```diff
@@ -1612,35 +1612,54 @@ _version_get_handler(xmpp_stanza_t* const stanza)
     xmpp_ctx_t* const ctx = connection_get_ctx();
     const char* id = xmpp_stanza_get_id(stanza);
     const char* from = xmpp_stanza_get_from(stanza);
+    ProfAccount* account = accounts_get_account(session_get_account_name());
+    GString* client = account->client != NULL ? g_string_new(account->client) : NULL;
```
sjaeckel, line 1616: Maybe I missed that part in the code, but if you don't need the `GString` functionality, there's no need to create a `GString`. Instead you could simply use `account->client`.

sjaeckel (changes requested):
Basically LGTM

Not sure whether this should really be in `/account` or whether it'd make more sense in `/software`!?

src/xmpp/iq.c
H3rnand3zzz: I just needed to copy this part in memory because I didn't want to make any changes in the original of `account->client`.

H3rnand3zzz:
> Basically LGTM
> 
> Not sure whether this should really be in `/account` or whether it'd make more sense in `/software`!?

It makes it easier to configure multiple accounts. Each account will have its own client, thus preserving privacy.

src/xmpp/iq.c
```diff
@@ -1612,35 +1612,54 @@ _version_get_handler(xmpp_stanza_t* const stanza)
...
+    bool is_custom_client = client != NULL;
+    gchar* custom_version_str = NULL;
+    if (is_custom_client) {
+        custom_version_str = strstr(client->str, " ");
+        if (custom_version_str != NULL) {
+            *custom_version_str = '\0'; // Split string on name and version
```
H3rnand3zzz, line 1622: This part is changing the string. If it will change account->client, it can create issues. Potentially this part can later be moved in account class, resulting in less computations, but it's a really minor issue as this function is being O(n) thus does not have big impact and rarely called (on request by specific users and on login).

sjaeckel: That's the part I missed.

src/xmpp/iq.c
sjaeckel:
```suggestion
    auto_char char* client = account->client != NULL ? strdup(account->client) : NULL;
```

We're moving away from using `GString` since its overhead is really not required in most cases. This is one of those were a simply `strdup()` is sufficient. If you use `auto_char` the compiler will take care of `free()`'ing that newly allocated string.

src/xmpp/iq.c
```diff
@@ -1612,35 +1612,54 @@ _version_get_handler(xmpp_stanza_t* const stanza)
...
+            custom_version_str++;
+        }
+        cons_show("Received IQ version request (XEP-0092) from %s. Client: %s; ver: %s", from, client->str, custom_version_str);
+    }
```
sjaeckel, line 1626:
```suggestion
    auto_char char* client = account->client != NULL ? strdup(account->client) : NULL;
    bool is_custom_client = client != NULL;
    gchar* custom_version_str = NULL;
    if (is_custom_client) {
        custom_version_str = strstr(client, " ");
        if (custom_version_str != NULL) {
            *custom_version_str = '\0'; // Split string on name and version
            custom_version_str++;
        }
        cons_show("Received IQ version request (XEP-0092) from %s. Client: %s; ver: %s", from, client, custom_version_str);
    }
```

We're moving away from using `GString` since its overhead is really not required in most cases. This is one of those were a simply `strdup()` is sufficient. If you use `auto_char` the compiler will take care of `free()`'ing that newly allocated string.

src/xmpp/iq.c
```diff
@@ -1612,35 +1612,54 @@ _version_get_handler(xmpp_stanza_t* const stanza)
...
+    xmpp_stanza_set_to(response, from);
+
+    xmpp_stanza_t* query = xmpp_stanza_new(ctx);
+    xmpp_stanza_set_name(query, STANZA_NAME_QUERY);
+    xmpp_stanza_set_ns(query, STANZA_NS_VERSION);
+
+    xmpp_stanza_t* name = xmpp_stanza_new(ctx);
+    xmpp_stanza_set_name(name, "name");
+    xmpp_stanza_t* name_txt = xmpp_stanza_new(ctx);
+    xmpp_stanza_set_text(name_txt, is_custom_client ? client->str : "Profanity");
```
sjaeckel, line 1651:
```suggestion
    xmpp_stanza_set_text(name_txt, is_custom_client ? client : "Profanity");
```

src/xmpp/iq.c
```diff
@@ -1680,31 +1696,31 @@ _version_get_handler(xmpp_stanza_t* const stanza)
...
-        xmpp_stanza_release(name);
-        xmpp_stanza_release(version);
-        if (include_os) {
-            xmpp_stanza_release(os);
-        }
-        xmpp_stanza_release(query);
-        xmpp_stanza_release(response);
+    // Cleanup
+    g_string_free(version_str, TRUE);
+    g_string_free(client, TRUE);
```
sjaeckel, line 1713:
```suggestion
```

H3rnand3zzz:
Builds failed by unrelated reasons

src/command/cmd_defs.c
```diff
@@ -2063,6 +2063,7 @@ static const struct cmd_t command_defs[] = {
               "/account set <account> otr <policy>",
               "/account set <account> pgpkeyid <pgpkeyid>",
               "/account set <account> startscript <script>",
+              "/account set <account> client <client>",
```
jubalh, line 2066: I think the latter <client> should actually be `<name>`.

src/command/cmd_ac.c
```diff
@@ -478,6 +478,7 @@ cmd_ac_init(void)
     autocomplete_add(account_set_ac, "otr");
     autocomplete_add(account_set_ac, "pgpkeyid");
     autocomplete_add(account_set_ac, "startscript");
+    autocomplete_add(account_set_ac, "client");
```
jubalh, line 481: Should probably be clientname or client identifier or something like that.

src/command/cmd_defs.c
```diff
@@ -2102,6 +2104,7 @@ static const struct cmd_t command_defs[] = {
               { "set <account> otr <policy>", "Override global OTR policy for this account, see /otr." },
               { "set <account> pgpkeyid <pgpkeyid>", "Set the ID of the PGP key for this account, see /pgp." },
               { "set <account> startscript <script>", "Set the script to execute after connecting." },
+              { "set <account> client <client>", "Replace Jabber client's name to another one. For example, you can set it to PSI." },
```
jubalh, line 2107: I think should be "Set XMPP client name according for discovery according to XEP-0092. Which only contacts that have presence subscription can query.

jubalh: Also you don't mention here that the version can be set as well and how. A example later on in `CMD_EXAMPLES` would be useful for this. This also then gets rid of the `For example, you can set it to PSI`.

jubalh (changes requested):
See comments

jubalh:
I reviewed the code. But apart from this I have some points:
* I'm not sure whether setting this can cause problems. I'm unsure whether some servers or clients use the client info to decide which features to use. But probably they will only use service discovery?
* Need to check who actually can query this. If its only users where you allowed this anyways, then it might not be needed.

https://xmpp.org/extensions/xep-0092.html sais:
```
he jabber:iq:version protocol SHOULD NOT be used to determine the identity of entities from which an application receives presence (e.g., contacts in a user's roster and certain kinds of gateways); [Entity Capabilities (XEP-0115)](https://xmpp.org/extensions/xep-0115.html) [[3](https://xmpp.org/extensions/xep-0092.html#nt-idm45786692116480)] SHOULD be used instead. However, the jabber:iq:version protocol MAY be used to determine the identity of entities from which an application does not receive presence (e.g., servers and many kinds of components). The jabber:iq:version protocol MAY also be used to determine information available only via jabber:iq:version (e.g., operating system information) for contacts from which a user receives presence, but only if the user specifically requests such information for a particular contact.
```

H3rnand3zzz:
The issues are addressed. As for potential problems - it might theoretically break OMEMO and so warns about ClientSwitcher for Psi+. But the feature is advanced and optional, so users should use it cautiously, thoughtfully and ad hoc.

src/command/cmd_funcs.c
```diff
@@ -840,6 +840,14 @@ _account_set_startscript(char* account_name, char* script)
     return TRUE;
 }
 
+gboolean
+_account_set_client(char* account_name, char* new_client)
+{
+    accounts_set_client(account_name, new_client);
+    cons_show("Updated client name for account %s: %s", account_name, new_client);
```
jubalh, line 847: Please use `Set client name for account X: x`

jubalh (changes requested):
See comments.
Also you need to squash the fixup commit into the first one.
And change the commit title, you still have `Add ClientSwitcher Feature` there.

jubalh (approved):

jubalh:
Thanks!
BTW now I saw that PSI also has a plugin with this functionality.

jubalh:
Just for the record: I was not sure if this will lead to problems along the way. I'm not certain whether some clients/servers use this field to cache something or change behaviour.
But I guess not many users will use this anyways. It is hard to cater to all the people. And the privacy enthusiasts will probably like this feature. A note was added to the command description.

sjaeckel:
@H3rnand3zzz while looking into another thing I just noticed this:

https://github.com/profanity-im/profanity/blob/766dc76e337c96e71edc698a8ee292014293c44f/src/xmpp/stanza.c#L2064-L2067

not sure if that's another thing you'd want to anonymize!?

PR #1813, opened 2023-04-03
add `/statusbar tabmode actlist`

sjaeckel:
The existing way how active tabs are displayed didn't allow showing more than 10 tabs. This patch adds a mode where the statusbar shows a comma-separated list of tabs which were active since the last time viewed.
This view is inspired by how `irssi` shows the active tabs, therefore it is also called `actlist`.

PR #1817, opened 2023-04-04
JID Display in Titlebar and Fix

### LINKED ISSUE #1816: Statusbar nickname
Is it intended that if you assign nickname to user, it will overwrite setting and show nickname instead of jid even if "/statusbar chat jid" setting is chosen?

https://github.com/profanity-im/profanity/blob/6b597f6608c454e48000847bb3c0b5c2fdc4f292/src/ui/statusbar.c#L200

H3rnand3zzz:
Fixes #1816 and more
See commit messages.
**Note:** on all of the screenshots setting `/statusbar self` set to `off` to preserve privacy of testing account.
### `/statusbar chat jid`
image
### `/statusbar chat user`
image

### `/statusbar chat jid`
image

### `/statusbar chat user`
image

### DIFF
--- a/src/command/cmd_defs.c
+++ b/src/command/cmd_defs.c
@@ -1260,8 +1260,8 @@ static const struct cmd_t command_defs[] = {
               { "show tls", "Show or hide TLS indicator in the titlebar." },
               { "show encwarn", "Enable or disable the unencrypted warning message in the titlebar." },
               { "show resource", "Show or hide the current resource in the titlebar." },
-              { "show name", "In case of a MUC. Show the MUC name in the titlebar." },
-              { "show jid", "In case of a MUC. Show the JID in the titlebar." })
+              { "show name", "Show the nickname or MUC name in the titlebar." },
+              { "show jid", "Show the JID in the titlebar." })
       CMD_EXAMPLES(
               "/titlebar up",
               "/titlebar show tls",
--- a/src/ui/statusbar.c
+++ b/src/ui/statusbar.c
@@ -182,10 +182,7 @@ status_bar_inactive(const int win)
 void
 _create_tab(const int win, win_type_t wintype, char* identifier, gboolean highlight)
 {
-    int true_win = win;
-    if (true_win == 0) {
-        true_win = 10;
-    }
+    int true_win = win == 0 ? 10 : win;
 
     StatusBarTab* tab = malloc(sizeof(StatusBarTab));
     tab->identifier = strdup(identifier);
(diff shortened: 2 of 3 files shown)


jubalh (changes requested):
The code looks good.
But somehow I'm not too happy with the commit messages.


For example if I go through code later, and do git blame on a certain line and want to see why/what was changed here, it would be more helpful to have a good description that I can quickly read instead of trying to analyze the code in the commit. Because when doing this for many commits it can be quite time intensive. So it would be nice to have more explanatory commit messages.

For example `- Fixes statusbar tabs by respecting "/statusbar chat" setting.` I would like to see what is actually wrong with the current way. what did it look like/behaved and why did you change this.

Sometimes also we want to change something, have some idea in our head but introduce a mistake in the code. And if we describe our actual idea better in the commit message then a reviewer can also better see if the code actually does what the coder wanted it to do.

H3rnand3zzz:
> The code looks good. But somehow I'm not too happy with the commit messages.
> 
> For example if I go through code later, and do git blame on a certain line and want to see why/what was changed here, it would be more helpful to have a good description that I can quickly read instead of trying to analyze the code in the commit. Because when doing this for many commits it can be quite time intensive. So it would be nice to have more explanatory commit messages.
> 
> For example `- Fixes statusbar tabs by respecting "/statusbar chat" setting.` I would like to see what is actually wrong with the current way. what did it look like/behaved and why did you change this.
> 
> Sometimes also we want to change something, have some idea in our head but introduce a mistake in the code. And if we describe our actual idea better in the commit message then a reviewer can also better see if the code actually does what the coder wanted it to do.

Changed it. I hope that this time it's better

H3rnand3zzz:
@jubalh please, review. Changes in the description are made as requested, no problems with the code were mentioned.

jubalh (changes requested):
First commit is wrong.
We have `/titlebar show name|jid`, which so far seems to apply only for MUCs. You need to query this setting and use it to let the user enabling/disabling the nickname and/or jid. It should be possible to turn on BOTH.

jubalh (approved):

jubalh:
Looks good! And I also noticed that you improved the commit messages A LOT! Thanks for this. Now the commit messages are much more helpful in quickly understanding the change!

jubalh:
Now we have a conflict because in the meantime we merged https://github.com/profanity-im/profanity/pull/1813 (sorry about that) which touched the same parts of the code.
You will have to use `git rebase master` on your branch and then need to return `pos`. Do you know how to do that or should we do this for you?

jubalh:
I noticed something:

What I tried:
* `/statusbar show name` (I have this off by default
* `/msg someone` where I chose someone in my roster that has a nick assigned
* `/statusbar chat jid`
I do this while I'm on the console window. And nothing changes. Only when I go to the window where I wrote to the user it updates it from 'nickname' to his jid.
So it seems I need to cycle all windows where this applies, to make the change actually visible.

H3rnand3zzz:
I will submit fixed version with the update issue mentioned in a couple of hours.

jubalh:
Also see e3820d7037a63ff3e09ed84f518dfb7b67f3982f for the info message.

jubalh:
Thanks @H3rnand3zzz !

PR #1820, opened 2023-04-05
Use ASCII chars for MAM loading string

alexandre1985:
* Fix non-ASCII chars, for correct display in TTY
 * UTF-8 personal names are excluded from this fix

### DIFF
--- a/src/ui/window.h
+++ b/src/ui/window.h
@@ -57,7 +57,7 @@
 #include "xmpp/muc.h"
 
 #define PAD_SIZE        1000
-#define LOADING_MESSAGE "Loading older messages…"
+#define LOADING_MESSAGE "Loading older messages..."
 
 void win_move_to_end(ProfWin* window);
 void win_show_status_string(ProfWin* window, const char* const from,


H3rnand3zzz (approved):

alexandre1985:
@jubalh I use GNU Screen in my TTY and it doesn't show this chars correctly.
I have tried quitting GNU Screen, run `unicode_start` and still my TTY doesn't show those "illegal chars" correctly.

My system is Artix Linux and my TTY have `lat9w-16` for the console font and `8859-1_to_uni` as the console translation. My unicodemap is `iso01`. Is something missing to make the TTY more modern?

Maybe I should rename the commit. Ask suggestion for another name in place of "illegal chars".

jubalh:
In this kind of commit description how you got to this end result doesn't matter much. What you do change in this commit matters. And UTF-8 is nothing illegal ;)

alexandre1985:
@jubalh Ok, UTF-8 is not illegal. My phrasing could have been better. Please check now if it is acceptable.

alexandre1985:
> And again I'm confused about the description. It's not a "fix".

@jubalh It would be better if you tell me what it is (instead of telling what is not). So that I may write a better commit message.

jubalh:
>  It would be better if you tell me what it is (instead of telling what is not). So that I may write a better commit message.

You mean like in my first comment where I wrote `I guess what you want to say is **Use only ASCII characters in MAM loading string**` ?

alexandre1985:
@jubalh is the new force push better?

mdosch:
As I already said in the MUC: `…` is not an illegal character and it works fine in my TTY. I guess you have either some misconfiguration or you use a font which can't show the character `…`. I think you should rather fix your TTY to also show non-ASCII characters than patching this out.
To check/change your setup do the following in Debian (I don't know how to do this on other distros):

```
# dpkg-reconfigure console-setup
```

alexandre1985:
Terminus shows the ellipsis char correctly in my terminal.

@mdosch So, you prefer to enforce a terminal font in all profanity users, than making this little adjustment?

I actually dislike terminus font look.

mdosch:
You don't have to settle with Terminus, there are many more fonts which 
should support UTF8, so it's not like Terminus is enforced on all 
users.

Also I think limiting characters to ASCII because people might use a 
font which doesn't support UTF8 is not a good solution.

Anyway, I am not the one to decide…

alexandre1985:
I want my font to be `lat9w-16`. And profanity does not display well that font.

Such a simple fix, there are many other '...' (3 dots) in the code. Why the resistance?

alexandre1985:
> You don't have to settle with Terminus, there are many more fonts which should support UTF8, so it's not like Terminus is enforced on all users. Also I think limiting characters to ASCII because people might use a font which doesn't support UTF8 is not a good solution. Anyway, I am not the one to decide…

When typing "in am not the one to decide..." why didn't you use the ellipsis char, instead of the 3 dots?

You be awesome to know, and this might lead somewhere...

mdosch:
On 05.04.2023 12:08, Daniel Santos wrote:
>When typing "in am not the one to decide..." why didn't you use the 
>ellipsis char, instead of the 3 dots?

That's exactly what I did… 🤣

alexandre1985:
> On 05.04.2023 12:08, Daniel Santos wrote: When typing "in am not the one to decide..." why didn't you use the ellipsis char, instead of the 3 dots?
> That's exactly what I did… rofl

Yes, I did a mistake.

alexandre1985:
I just counted, the 3 dots are printed 26 times by profanity. The ellipsis char is printed 1 time by profanity. Choose. @mdosch

alexandre1985:
Here is my case:

In order for this project be consistant to the user, I am considering that for strings printed to the user, the project will either use 3dots or the ellipsis char for all cases.

=== 3dots ===

** Advantages **

1. Works in all TTY displays, independent of TTY font.
2. User is free to choose whatever font they wish for their TTY.
3. Needs to change 1 entry of commited work from 1 developer, instead of 26 entries from numerous developers.

** Disavantages **

None.


=== Ellipsis char ===

** Advantages **

1. Use a new char.

** Disavantages **

1. Doesn't work in all TTY displays, independent of TTY font.
2. Forcing user to use a small group of today's TTY fonts.
3. Needs to change 26 entries of commited work from numerous developers, instead of 1 entry from one developer.
4. The new ellipsis char has always been replaced by 3dots chars, before the ellipsis char existance. It is redundant.

=====

Choose wisely.

jubalh:
I updated the PR title for you, please do this the next time as well.

I checked on both my Gentoo and openSUSE machines. On both Unicode on TTY works without problems. All with default settings, I didn't change/configure anything in this regard manually.

alexandre1985:
Such a simple to approve PR. A no-brainer.

You have definitely not tried on Arch Linux. It does *not* work in Arch. Arch Linux has a much bigger number of users than either Gentoo or OpenSUSE.

And I guess you don't want *consistency* in on Profanity. What a poor decision.

> I updated the PR title for you, please do this the next time as well.

There is no need, there won't be a next PR from me.

sjaeckel:
btw. I'm using Arch as well, and it works for me

alexandre1985:
> btw. I'm using Arch as well, and it works for me

It does not work on the TTY with the default font (lat1-16 or lat9w-16).

mdosch:
On 09.04.2023 03:00, Daniel Santos wrote:
>> btw. I'm using Arch as well, and it works for me
>
>It does not work on the TTY with the default font (lat1-16 or 
>lat9w-16).

What do you mean by _it does not work_? Is profanity crashing? I guess 
you should just see an replacement char if your font does not support 
UTF8 which should be no big deal. Without an UTF8 font you will see 
those already often in every chat where people use UTF8 chars or emoji…

PR #1821, opened 2023-04-09
User proper ellipsis char

mdosch:
As stated in
https://github.com/profanity-im/profanity/pull/1820#issuecomment-1498083383
profanity uses "..." (three dots) in a lot of places instead the proper
ellipsis char "…".


sjaeckel:
m(

PR #1823, opened 2023-04-12
Fix double free crash

H3rnand3zzz:
Read commit messages. For crash payload, contact me.


sjaeckel (changes requested):
Thanks for the PR!

I agree with the changes, but I don't see a potential double free here.

The issue we're seeing is an out-of-bounds access of the data returned by `gpgme_data_release_and_get_mem()`. The call only allocates `len` bytes, but profanity is accessing `len+1` bytes.

Didn't we already have a similar issue in the same part of the code not long ago? Maybe those parts should be reviewed!?

Please update the commit message, PR title&description accordingly.

H3rnand3zzz:
> Thanks for the PR!
> 
> I agree with the changes, but I don't see a potential double free here.
> 
> The issue we're seeing is an out-of-bounds access of the data returned by `gpgme_data_release_and_get_mem()`. The call only allocates `len` bytes, but profanity is accessing `len+1` bytes.
> 
> Didn't we already have a similar issue in the same part of the code not long ago? Maybe those parts should be reviewed!?
> 
> Please update the commit message, PR title&description accordingly.

In fact, there was a crash caused by some payload, which glib claimed to be "double free". I don't see double free here as well, so probably it's a false positive, which breaks the execution anyway. It can be renamed to "memory corruption", I think.

Line from `where` of GDB:
#5  0x00007f48a05e8657 in malloc_printerr (str=str@entry=0x7f48a06f45f8 "double free or corruption (!prev)") at malloc.c:5651

H3rnand3zzz:
> > I renamed to "memory corruption".
> 
> please also reword the commit message. GH is ephemeral, the Git history will most likely survive this service.

I thought that I did, probably messed it up on some stage :) I will redo in a minute.

sjaeckel (approved):

jubalh:
Mmhh now we have a merge commit in there. Can you just rebase on master instead?
Also I would prefer if the 2nd commit (re cleanup) could be a bit more descriptive in the heading. Something like `Cleanup p_ox_gpg_decrypt` or something like that would already suffice. Then when skimming through the list of changes when looking for something specific I don't need to click on the entry to see which part is actually cleaned up/touched.

jubalh:
Thanks!

PR #1824, opened 2023-04-13
Smooth scrolling support (+mouse wheel scrolling in some terminals)

H3rnand3zzz:
Added mouse scroll support with shift+scroll using readline.

Points before release:

- [x] Mouse compatibility
- [x] @sjaeckel's suggestions
- [x] Infinite scroll support

Known Issues: Sometimes it gets incredibly laggy (probably have something to do with profanity core)

### DIFF
--- a/src/ui/inputwin.c
+++ b/src/ui/inputwin.c
@@ -136,6 +136,7 @@ static int _inp_rl_subwin_pageup_handler(int count, int key);
 static int _inp_rl_subwin_pagedown_handler(int count, int key);
 static int _inp_rl_startup_hook(void);
 static int _inp_rl_down_arrow_handler(int count, int key);
+static int _inp_rl_scroll_handler(int count, int key);
 static int _inp_rl_send_to_editor(int count, int key);
 static int _inp_rl_print_newline_symbol(int count, int key);
 
@@ -554,6 +555,8 @@ _inp_rl_startup_hook(void)
     rl_bind_key('\t', _inp_rl_tab_handler);
     rl_bind_keyseq("\\e[Z", _inp_rl_shift_tab_handler);
 
+    rl_bind_keyseq("\\e[1;3A", _inp_rl_scroll_handler);     // alt + scroll/arrow up
+    rl_bind_keyseq("\\e[1;3B", _inp_rl_scroll_handler);     // alt + scroll/arrow down
     rl_bind_keyseq("\\e[1;5B", _inp_rl_down_arrow_handler); // ctrl+arrow down
     rl_bind_keyseq("\\eOb", _inp_rl_down_arrow_handler);
 
@@ -906,7 +909,7 @@ static int
 _inp_rl_win_pageup_handler(int count, int key)
 {
     ProfWin* current = wins_get_current();
-    win_page_up(current);
+    win_page_up(current, 0);
     return 0;
 }
 
@@ -914,7 +917,7 @@ static int
 _inp_rl_win_pagedown_handler(int count, int key)
 {
     ProfWin* current = wins_get_current();
-    win_page_down(current);
+    win_page_down(current, 0);
     return 0;
 }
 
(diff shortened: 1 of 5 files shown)


src/ui/inputwin.c
```diff
@@ -391,7 +392,7 @@ _inp_edited(const wint_t ch)
     bytes[utf_len] = '\0';
     gunichar unichar = g_utf8_get_char(bytes);
 
-    return g_unichar_isprint(unichar);
+    return g_unichar_isprint(unichar) && (ch != KEY_MOUSE);
```
H3rnand3zzz, line 395: should be removed

src/ui/inputwin.c
```diff
@@ -924,6 +927,45 @@ _inp_rl_subwin_pagedown_handler(int count, int key)
     return 0;
 }
 
+static int
+_inp_rl_scroll(int count, int key)
+{
+    if (!prefs_get_boolean(PREF_MOUSE))
+        return 0;
+    cons_show(g_strdup_printf("scroll. count: %d, key: %d, %c", count, key, key));
```
sjaeckel, line 935: I guess this in only a debug statement, which will go soonish.

FYI `cons_show()` behaves like `printf()`, i.e. you can drop the call to `g_strdup_printf()`. (also you're leaking the value returned :D)

H3rnand3zzz:
It's ready for review, but not ready for release. I would appreciate advises on lag issue. Probably core-related. Even without binding, mouse scrolling makes considerable lag.

src/ui/inputwin.c
```diff
@@ -927,6 +935,44 @@ _inp_rl_subwin_pagedown_handler(int count, int key)
...
+}
+
+static int
+_inp_rl_scroll_impl(int count, int key)
+{
+    if (!prefs_get_boolean(PREF_MOUSE))
+        return 0;
+    ProfWin* window = wins_get_current();
+
+    if (key == 'B') { // mouse wheel down
```
jubalh, line 968: here i agree with clang that the comments should be on top or inside ;)

src/command/cmd_defs.c
```diff
@@ -1283,6 +1283,22 @@ static const struct cmd_t command_defs[] = {
...
+      CMD_SYN(
+              "/mouse on",
+              "/mouse off")
+      CMD_DESC(
+              "Mouse scroll controls.")
+      CMD_ARGS(
+              { "on", "Activate mouse scroll." },
+              { "off", "Deactivate mouse scroll." })
+    },
+    //
```
jubalh, line 1300: ?

jubalh (changes requested):

src/command/cmd_defs.c
sjaeckel, line 1300:
```suggestion
```

src/ui/inputwin.c
```diff
@@ -927,6 +935,46 @@ _inp_rl_subwin_pagedown_handler(int count, int key)
...
+_inp_rl_scroll_trampoline(int count, int key)
+{
+    return _inp_rl_scroll_fp(count, key);
+}
+
+static int
+_inp_rl_scroll_impl(int count, int key)
+{
+    if (!prefs_get_boolean(PREF_MOUSE))
+        return 0;
```
sjaeckel, line 965:
```suggestion
```
This can go, since this function will only be called if this flag is set to true.

sjaeckel (changes requested):
Again clean and easy to follow.

I'd like to have a second look tomorrow after those proposed changes are included.

H3rnand3zzz:
To make it clear

At the current stage, the commit is incompatible with mouse and works only with touch pad. I hope to resolve it as soon as possible.

H3rnand3zzz:
Rebased on master
Followed by changing for alt+up/down instead of shift+up/down. Alt it more intuitive (since it presumes ALTernative behaviour for the same button) and minor changes to merge with previous updates.

jubalh (approved):

H3rnand3zzz:
@sjaeckel, is PR ok now?

sjaeckel:
I'll have a look soon, currently only on mobile

H3rnand3zzz:
### New revision
 - Changes requested by @sjaeckel are implemented:
   - Removed preference (since alt+arrows support for navigation does not need to be optional)
   - Details are added in command description regarding other terminals' support

sjaeckel (approved):

PR #1826, opened 2023-04-13
Add nickname support for `/roster remove`

H3rnand3zzz:
See commit messages.

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -2922,14 +2922,17 @@ cmd_roster(ProfWin* window, const char* const command, gchar** args)
             cons_show("You are not currently connected.");
             return TRUE;
         }
-        char* jid = args[1];
-        if (jid == NULL) {
+        char* usr = args[1];
+        if (usr == NULL) {
             cons_bad_cmd_usage(command);
-        } else {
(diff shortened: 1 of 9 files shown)


jubalh:
Do I see it right that this removed the possibility to remove a user based on JID?

H3rnand3zzz:
> Do I see it right that this removed the possibility to remove a user based on JID?

It in fact did not remove this opportunity. It's covered by tests and I just tested it out manually. It only adds ability to remove user by nickname as well.

src/command/cmd_funcs.c
```diff
@@ -2922,14 +2922,17 @@ cmd_roster(ProfWin* window, const char* const command, gchar** args)
...
+        char* usr = args[1];
+        if (usr == NULL) {
             cons_bad_cmd_usage(command);
-        } else {
-            roster_send_remove(jid);
+            return TRUE;
         }
+        char* barejid = roster_barejid_from_name(usr);
+        if (barejid == NULL) {
+            barejid = usr;
```
H3rnand3zzz, line 2932: this lines secures the deletion if name was not found (hence jid was used)

jubalh (approved):

jubalh:
Thank you!

Issue #1825, opened 2023-04-13
Session alarm feature


H3rnand3zzz:
I plan to implement session alarm security feature and I want to hear an opinion about it. It will show alert if too many sessions are active at the moment. For example, you have 1 laptop and 1 phone. It means that normally up to 2 devices might use the same account. If 3  devices are using the same account at the same time, you will see notification and can spot suspicious activity.

## Usage
`/account set <account> session_alarm <max_sessions>`
will activate an alarm with max_sessions set. 

For example, you use account me from only 1 device and you use:
`/account set me session_alarm 1`
after this if another connection will be spotted, you will get an alarm. Possibly in this way: system notification, cons_alert, console message, current chat message.

## Current Behavior
Currently no security features with similar functionality are in place.

## Stage of completion
- [x] Session_alarm
    - [X] Discussion
    - [x] Add command
        - [x] Add description
        - [x] Add autocomplete
        - [x] Add function that will change account preference
    - [x] Implement logic
        - [x] Enable/disable in account preferences
        - [x] Track sessions count
        - [x] Use `connection_get_available_resources` to get current resources
        - [x] On new event make alert if needed and enabled
        - [x] Get jabber client, OS and other info. Display it in alert.
    - [x] Tests
        - [x] Alter old tests
    - [x] Push the commit #1827

PR #1827, opened 2023-04-14
New Feature: Session Alarm

H3rnand3zzz:
Implements #1825 
image

On the screenshot I connected with Psi+ using Client Switcher plugin. The plugin couldn't change its name everywhere, so you can see original name (Psi+) and also changed name (Pidgin).


jubalh:
I'm actually starting to think that it might be a good idea to have a new `/privacy` command.

Where this session alarm is a subcommand of.
And also the command that set's the client name introduced in https://github.com/profanity-im/profanity/pull/1815 should be a subcommand of this.

I however would first review and merge this PR.

And then you could create a new PR that introduces the new subcommand and moves the client name and session alarm into it. What do you think?

Also I'm thinking if we are bloating the `account` struct too much. And whether this could then be saves somewhere else. But about this latter idea I need to check more code first.

sjaeckel:
> I however would first review and merge this PR.
> 
> And then you could create a new PR that introduces the new subcommand and moves the client name and session alarm into it. What do you think?

I'd propose to introduce `/privacy` now and in that next PR only move the other commands.

> Also I'm thinking if we are bloating the `account` struct too much. And whether this could then be saves somewhere else. But about this latter idea I need to check more code first.

That struct is one of the best examples of BBOM I've seen in the wild ... I also looked into changing it, but that's going to be a huge refactoring ...

jubalh:
> I'd propose to introduce /privacy now and in that next PR only move the other commands.

I believe that PR review and discussion can get hard to overlook/confusing in this case.
Also I think the `/privacy` change needs more discussion to get it done right.

src/xmpp/presence.c
```diff
@@ -656,6 +658,69 @@ _available_handler(xmpp_stanza_t* const stanza)
...
+        char* cur_resource = accounts_get_resource(account_name);
+        if (max_sessions > 0) {
+            int res_count = connection_count_available_resources();
+            if (res_count > max_sessions && g_strcmp0(cur_resource, resource->name)) {
+                ProfWin* console = wins_get_console();
+                ProfWin* current_window = wins_get_current();
+                auto_gchar gchar* message = g_strdup_printf("Max sessions alarm! (%d/%d devices in use)", res_count, max_sessions);
+                win_println(console, THEME_RED, "|", message);
+                if (console != current_window) {
+                    win_println(current_window, THEME_RED, "|", message);
```
sjaeckel, line 672: That's the only place where we print in the `current_window`.
Would it maybe make sense to either add some explanation that the console window has more information? Or print everything in both windows? IMO adding an explanation here should be sufficient.
```suggestion
                    win_println(current_window, THEME_RED, "|", "%s - check the console for more details!", message);
```

src/xmpp/presence.c
```diff
@@ -656,6 +658,69 @@ _available_handler(xmpp_stanza_t* const stanza)
...
+        char* account_name = session_get_account_name();
+        int max_sessions = accounts_get_max_sessions(account_name);
+        char* cur_resource = accounts_get_resource(account_name);
+        if (max_sessions > 0) {
+            int res_count = connection_count_available_resources();
+            if (res_count > max_sessions && g_strcmp0(cur_resource, resource->name)) {
+                ProfWin* console = wins_get_console();
+                ProfWin* current_window = wins_get_current();
+                auto_gchar gchar* message = g_strdup_printf("Max sessions alarm! (%d/%d devices in use)", res_count, max_sessions);
+                win_println(console, THEME_RED, "|", message);
```
sjaeckel, line 670:
```suggestion
                win_println(console, THEME_RED, "|", "%s", message);
```

This string is statically created and has no chance of format string injection, but good practice is to only use const format strings.

@jubalh one easy improvement of the code base security-wise could be to use [`__attribute__ ((format (printf, x, y)))`](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-format-function-attri) in our `printf` class of internal API. What do you think?

src/xmpp/presence.c
```diff
@@ -656,6 +658,69 @@ _available_handler(xmpp_stanza_t* const stanza)
 
     if (g_strcmp0(xmpp_presence->jid->barejid, my_jid->barejid) == 0) {
         connection_add_available_resource(resource);
+        char* account_name = session_get_account_name();
+        int max_sessions = accounts_get_max_sessions(account_name);
+        char* cur_resource = accounts_get_resource(account_name);
+        if (max_sessions > 0) {
+            int res_count = connection_count_available_resources();
```
sjaeckel, line 665:
```suggestion
        const char* account_name = session_get_account_name();
        int max_sessions = accounts_get_max_sessions(account_name);
        if (max_sessions > 0) {
            const char* cur_resource = accounts_get_resource(account_name);
            int res_count = connection_count_available_resources();
```

src/xmpp/presence.c
```diff
@@ -656,6 +658,69 @@ _available_handler(xmpp_stanza_t* const stanza)
...
+            if (res_count > max_sessions && g_strcmp0(cur_resource, resource->name)) {
+                ProfWin* console = wins_get_console();
+                ProfWin* current_window = wins_get_current();
+                auto_gchar gchar* message = g_strdup_printf("Max sessions alarm! (%d/%d devices in use)", res_count, max_sessions);
+                win_println(console, THEME_RED, "|", message);
+                if (console != current_window) {
+                    win_println(current_window, THEME_RED, "|", message);
+                }
+                notify(message, 10000, "Security alert");
```
sjaeckel, line 674: Shouldn't this be enclosed in a `if (prefs_get_boolean(PREF_NOTIFY_SECURITY_ALERTS))`? (`PREF_NOTIFY_SECURITY_ALERTS` does not exist yet, maybe we should choose another name!?)

src/xmpp/presence.c
```diff
@@ -656,6 +658,69 @@ _available_handler(xmpp_stanza_t* const stanza)
...
+                        if (software_version->os || software_version->os_version) {
+                            win_newline(console);
+                        }
+                    }
+
+                    caps_destroy(caps);
+                }
+
+                win_println(console, THEME_RED_BOLD, "|", "If it wasn't you, change your password. Use: /changepassword");
+                win_println(console, THEME_GREEN, "|", "If it was you, change max devices to get an alarm, use: /account set %s session_alarm %d", account_name, res_count);
```
sjaeckel, line 720:
```suggestion
                win_println(console, THEME_GREEN, "|", "If it was you, update the `session_alarm` limit that determines when to trigger this alarm, use: /account set %s session_alarm %d", account_name, res_count);
```

sjaeckel (changes requested):
Basically LGTM.

Maybe think of putting most of the newly added code of `src/xmpp/presence.c` into its own static function? Maybe also not :)

src/xmpp/presence.c
jubalh: Hmm yes actually the statusbar indicates a change in the console window if something is printed there and the user should check there. I even wouldn't print the notice about console. Since it's like this in all (?) commands.

src/xmpp/presence.c
jubalh: I didn't quite understand what it does. Will read the docu again later :D

src/xmpp/presence.c
H3rnand3zzz: I think on the next PR with /privacy command we can integrate the setting

src/command/cmd_funcs.c
```diff
@@ -904,6 +904,28 @@ _account_set_auth(char* account_name, char* policy)
     return TRUE;
 }
 
+gboolean
+_account_set_max_sessions(char* account_name, char* max_sessions_raw)
+{
+    int max_sessions;
+    char* err_msg = NULL;
+    gboolean res = strtoi_range(max_sessions_raw, &max_sessions, -1, 100000, &err_msg);
```
sjaeckel, line 912:
```suggestion
    gboolean res = strtoi_range(max_sessions_raw, &max_sessions, 0, INT_MAX, &err_msg);
```
Why `-1` as lower boundary?

Also this just made me realize that this should've been an `unsigned int` internally and not a signed one... but we can keep it like that.

@jubalh maybe we should think of introducing usage of `g_key_file_[gs]et_uint64()` for such values!?

src/xmpp/presence.c
sjaeckel: Maybe add it to the tasklist in #1825 ? or open a new tracking issue for the `/privacy` command?

sjaeckel (changes requested):
Sorry, that's a bit late, I overlooked it at the first review.

src/command/cmd_funcs.c
jubalh: Sounds like a good idea! Also the `__attribute__` that you mentioned in another comment is a good idea!

jubalh:
Ok I'll create an issue for `/privacy`.

src/xmpp/presence.c
jubalh: i added a comment to the new privacy issue

jubalh (approved):
Please change the min max and then good to go

src/command/cmd_funcs.c
H3rnand3zzz: > Why `-1` as lower boundary?
> 
> Also this just made me realize that this should've been an `unsigned int` internally and not a signed one... but we can keep it like that for now.
> 
> @jubalh maybe we should think of introducing usage of `g_key_file_[gs]et_uint64()` for such values!?

`-1` looked logical to set it as "turned off" as in some programs for certain parameters. `100000` sessions sounded as logical limitation because I can't think of a scenario when someone would want to set more sessions. And the idea was to show for user that sets it to`9999999999` that it's possible to turn off by setting it to `-1` or `0`.

sjaeckel: Interesting interpretation, but `0` already turns the feature off!? Two "off" values doesn't make sense.

I would've expected that `-1` could mean "enable monitoring, but only report new devices" or something like that.

Please change the range to `0, INT_MAX` and we're fine to go with this PR.

sjaeckel: and btw. I also can't think of a use-case where `100000` sessions aren't enough, but I don't like the idea of just choosing value X of a limitation when there's only a technical limitation (`INT_MAX` in our case) in theory.

Fun fact: by using `100000` you were breaking a(nother very theoretical) case where someone wants to compile profanity for a 16bit CPU, since it's possible on such an architecture that an `int` is only able to hold values from `-32768 to 32767` :)

H3rnand3zzz:
Thank you for reviews, changes are applied

jubalh:
Great! Thank you!

PR #1829, opened 2023-04-14
Don't add the same command twice to history

H3rnand3zzz:
See commit messages.

### DIFF
--- a/src/ui/inputwin.c
+++ b/src/ui/inputwin.c
@@ -568,12 +568,15 @@ _inp_rl_startup_hook(void)
 static void
 _inp_rl_linehandler(char* line)
 {
-    if (line && *line) {
-        if (!get_password) {
-            add_history(line);
-        }
-    }
     inp_line = line;
+    if (!line || !*line || get_password) {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


jubalh:
>  Fix history duplicates

History is actually a broad term in our client :)
Maybe be a little more specific like: `Don't add the same command twice to history` or something.

> Bug is fixed by introducing check.

I don't consider this a bug. For example in bash you can also configure this behaviour. If you want to each line you type added or remove duplicates.

BTW did you check whether readline maybe even has an option for this? Maybe via inputrc?

jubalh (approved):
Commit message still refercnes "bug" otherwise good to go :)

sjaeckel (approved):
> BTW did you check whether readline maybe even has an option for this? Maybe via inputrc?

Did you check?

H3rnand3zzz:
> > BTW did you check whether readline maybe even has an option for this? Maybe via inputrc?
> 
> Did you check?

https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File-Syntax.html
Haven't found any mention of setting like this one

jubalh:
Thank you!

Issue #1264, opened 2020-02-07
[feature request] Allow utf8 symbols as OMEMO char


mdosch:
I'd like to be able to set utf8 symbols (e.g. the padlock emoji) as OMEMO char.

PR #1831, opened 2023-04-15
Add encryption section to man page

jubalh:
The goal should be that users can make a more informed decision about which encryption they actually want to use.

We can also use this to document usage details of implementation quirks, if any.

### DIFF
--- a/docs/profanity.1
+++ b/docs/profanity.1
@@ -106,6 +106,27 @@ for interactive history search and
 for reloading inputrc without restart.
 .SH USING PROFANITY
 The user guide can be found at <https://profanity-im.github.io/userguide.html>.
+.SH ENCRYPTION
+Profanity supports various kinds of encryption: OMEMO, OTR, PGP, OX.
+You can only enable one of them per correspondent at a time.
+.TP
+.BR OMEMO
+OMEMO (/omemo) is defined in XEP-0384. It uses an implementation of the Signal protocol for key management and to synchronize messages among different clients. It works even when other clients are offline. And offers Perfect Forward Secrecy and Plausible deniability. Servers need to support PEP (XEP-0163).
+We implement the "siacs" version of OMEMO. Version 0.3.0, which is currently the widest adopted option.
+OMEMO is the only encryption option in Profanity that also supports encryption for MUCs (XEP-0045) and file transfer via HTTP upload (XEP-0363).
+.TP
+.BR OTR
+OTR (/otr) is defined in XEP-0364. It uses a combination of the AES symmetric-key algorithm, the Diffie–Hellman key exchange, and the SHA-1 hash function. It offers deniable authentication and Perfect Forward Secrecy. To initialize a session both clients need to be online at the same time. A session is between two clients, so multiclient chats are not working. Which is a feature not a bug. OTR does by design not work for MUCs.
+.TP
+.BR OpenPGP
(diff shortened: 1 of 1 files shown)


jubalh:
Reviews/ideas/improvements welcome!

docs/profanity.1
```diff
@@ -106,6 +106,24 @@ for interactive history search and
 for reloading inputrc without restart.
 .SH USING PROFANITY
 The user guide can be found at <https://profanity-im.github.io/userguide.html>.
+.SH ENCRYPTION
+Profanity supports various kinds of encryption: OMEMO, OTR, PGP, OX.
+You can only enable one of them per correspondent at a time.
+.TP
+.BR OMEMO
+OMEMO (/omemo) is defined in XEP-0384. It uses an implementation of the Signal protocol for key management and to synchronize messages among different clients. It works even when other clients are offline. And offers Perfect Forward Secrecy and Plausible deniability. Servers need to support PEP (XEP-0163).
```
mdosch, line 114: Maybe mention the version of the OMEMO protocol implemented as this makes quite a difference: https://wiki.xmpp.org/web/Tech_pages/OMEMO

mdosch, line 114: Maybe also mention that OMEMO is the only (? no idea about OTR) which is supported in MUC and which supports http-upload.

docs/profanity.1
```diff
@@ -106,6 +106,24 @@ for interactive history search and
...
+.TP
+.BR OTR
+OTR (/otr) is defined in XEP-0364. It uses a combination of the AES symmetric-key algorithm, the Diffie–Hellman key exchange, and the SHA-1 hash function. It offers deniable authentication and Perfect Forward Secrecy. To initialize a session both clients need to be online at the same time. A session is between two clients, so multiclient chats are not working. Which is a feature not a bug.
+.TP
+.BR OpenPGP
+OpenPGP (/pgp) is defined in XEP-0027. Is uses a public and secret key. It is also known as Legacy OpenPGP and has been depcrecated. It doesn't provide protection against replay attacks.
```
mdosch, line 120: Maybe mention that http-upload and MUC are not supported (Conversations supports http-upload with PGP but afaik this is not standardized).

docs/profanity.1
```diff
@@ -106,6 +106,24 @@ for interactive history search and
...
+.TP
+.BR OX
+OX (/ox) is defined in XEP-0373 and XEP-0374. It's a more modern way to use OpenPGP on XMPP and tries to fix the shortcomings of legacy XEP-0027. Servers need to support PEP (XEP-0163).
```
mdosch, line 123: Maybe mention that http-upload and MUC are not (yet) standardized.

docs/profanity.1
H3rnand3zzz: OTR does not support MUC

H3rnand3zzz:
It might make sense to differentiate protocols that require server support from those that don't. And maybe even provide something like comparison table: server support requirement, forward secrecy, plausible deniability, file transfer, replay attacks vulnerability.

It makes sense to warn users that by default messages are not E2E encrypted and can be captured by server or even provider if connection to the server is not secure.

Maybe we can even highlight unencrypted messages in the future (for example, by using 🔒 for encrypted and 🔓 for unencrypted messages).

jubalh:
> It might make sense to differentiate protocols that require server support from those that don't

I did. That's when I wrote about PEP for example.

> And maybe even provide something like comparison table

It's on the linked xmpp page.

> Maybe we can even highlight unencrypted messages in the future (for example, by using lock for encrypted and unlock for unencrypted messages).

We do. It's at the left side where usually we see `-`. I believe the default for encryption is `~`. And since https://github.com/profanity-im/profanity/issues/1264 we also support exactly the character you mentioned.

jubalh:
Thanks for the reviews!

PR #1835, opened 2023-04-16
Fix crash, add consistency to logging

H3rnand3zzz:
See commit messages.


jubalh:
`sv_ev_delayed_private_message()` is about private MUC PMs. Which we don't want to log AFAIK.
The reason is that MUCs can be anonymous and you can't know the JID. And also people don't need to register their nick to their JID.
So it can happen that me1@somewhere.org logs into muc@somewhere.org as `nick1`. The he goes offline and later you1@somewhere.org logs in with the same nick.

jubalh (approved):

PR #1837, opened 2023-04-18
Fix two recently introduced memleaks

jubalh:
Got introduced with 5d3c8ce7c.

### DIFF
--- a/src/ui/inputwin.c
+++ b/src/ui/inputwin.c
@@ -577,6 +577,7 @@ _inp_rl_linehandler(char* line)
     if (last == NULL || strcmp(last->line, line) != 0) {
         add_history(line);
     }
+    free(history);
 }
 
 static gboolean shift_tab = FALSE;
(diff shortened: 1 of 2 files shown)


src/xmpp/stanza.c
```diff
@@ -995,6 +995,9 @@ stanza_create_caps_query_element(xmpp_ctx_t* ctx)
 #endif
         }
     }
+
+    account_free(account);
```
H3rnand3zzz, line 999: It should work. Ideally, in the future we should just get 1 parameter instead of fetching full account.

src/ui/inputwin.c
```diff
@@ -577,6 +577,7 @@ _inp_rl_linehandler(char* line)
     if (last == NULL || strcmp(last->line, line) != 0) {
         add_history(line);
     }
+    free(history);
```
H3rnand3zzz, line 580: If the code does not break, we should probably clean it.

H3rnand3zzz (approved):

PR #1843, opened 2023-05-02
Add ability to disable avatar publishing

IsaacM88:
Add "/avatar disable" to comply with point "3.5 Publisher Disables Avatar Publishing" in XEP-0084.

**src/command/cmd_defs.c:2416**
Add "disable" argument. Reword the "/avatar" command description so it flows better.

**src/command/cmd_ac.c:1101**
Add "disable" to the "/avatar" autocomplete dictionary.

**src/command/cmd_funcs.c:9277**
Split "/avatar" commands into two groups with an if statement; those with a parameter and those without. "cons_bad_cmd_usage()" is in both groups, which is messy. "disable" has similar logic to "set", but it includes a failure message.

**src/xmpp/avatar.c:152**
"avatar_publishing_disable()" uses the same logic to publish metadata as in "avatar_set()".

**src/xmpp/avatar.c:238**
Add a message to inform users when they do not receive an avatar after using "/avatar get" and "/avatar open". In case of a failure, the user will be subscribed to future avatar updates as long as they continue to use their current instance of profanity. Adding "caps_remove_feature()" after "cons_show()" will unsubscribe the user from avatar updates and prevent the user from downloading an avatar unexpectedly hours later without issuing an "/avatar" command.

**src/xmpp/stanza.c:2698**
The new "disable" function follows the same logic as "stanza_create_avatar_metadata_publish_iq()".

* Use "/avatar set" to upload an avatar.
* Use "/avatar get" on a second Profanity instance using another JID. The avatar should download.
* Use "/avatar disable" on the first Profanity instance. A _disabled_ message should appear.
* Use "/avatar get" again on the second instance. A _failure_ message should appear.
* Use "/avatar set" again on the first instance. The avatar should automatically download on the second instance.

# I ran valgrind when using my new feature
Yes

### DIFF
--- a/src/xmpp/avatar.c
+++ b/src/xmpp/avatar.c
@@ -151,6 +151,17 @@ avatar_set(const char* path)
 #endif
 
 gboolean
+avatar_publishing_disable()
+{
+    xmpp_ctx_t* const ctx = connection_get_ctx();
+    xmpp_stanza_t* iq = stanza_disable_avatar_publish_iq(ctx);
+    iq_send_stanza(iq);
+    xmpp_stanza_release(iq);
+
+    return TRUE;
+}
+
+gboolean
 avatar_get_by_nick(const char* nick, gboolean open)
 {
     // in case we set the feature, remove it
@@ -224,6 +235,9 @@ _avatar_metadata_handler(xmpp_stanza_t* const stanza, void* const userdata)
                         _avatar_request_item_by_id(from, data);
                     }
                 }
+            } else {
+                cons_show("We couldn't get the user's avatar, possibly because they haven't set one or have disabled avatar publishing. "
+                          "During this Profanity session, you will receive future changes to this user's avatar.");
             }
         }
     }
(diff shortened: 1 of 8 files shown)


src/xmpp/avatar.c
```diff
@@ -224,6 +235,9 @@ _avatar_metadata_handler(xmpp_stanza_t* const stanza, void* const userdata)
                         _avatar_request_item_by_id(from, data);
                     }
                 }
+            } else {
+                cons_show("We couldn't get the user's avatar, possibly because they haven't set one or have disabled avatar publishing. "
+                          "During this Profanity session, you will receive any future changes to this user's avatar.");
```
jubalh, line 240: something 's wrong with this sentence I think (but I'm not native speaker)

jubalh:
Cool! Thank you!
I did a quick review and it looks good, but will do another one tomorrow (?).

jubalh (approved):

jubalh:
Thanks for your contribution!

Issue #1846, opened 2023-05-09
`/reconnect now` doesn't work


H3rnand3zzz:
At this point command `/reconnect now` doesn't work. The problem is that the command causes program to enter invalid state (nor connected, nor disconnected).

As an additional problem, auto-reconnect by itself doesn't work. It looks like `libstrophe` struggles to handle connection issues. It handles disconnect caused by the server, but timeouts behaviour is questionable.

```
prof: DBG: Input received: /reconnect now
prof: DBG: Attempting reconnect with account <account>
prof: INF: Connecting without account, JID: <jid>
prof: INF: Connecting as <jid>
prof: DBG: Connecting with flags (0x2):
prof: DBG:   XMPP_CONN_FLAG_MANDATORY_TLS
conn: ERR: Flags can be set only for disconnected connection
prof: ERR: libstrophe doesn't accept this combination of flags: 0x2
```

After this command, program enters bad state. You are unable to connect and to disconnect.

On command `/connect <account>`
```
prof: DBG: Input received: /connect <account>
prof: INF: Connecting using account: <account>
prof: INF: Connecting as <jid>
prof: DBG: Connecting with flags (0x2):
prof: DBG:   XMPP_CONN_FLAG_MANDATORY_TLS
conn: ERR: Flags can be set only for disconnected connection
prof: ERR: libstrophe doesn't accept this combination of flags: 0x2
prof: INF: Connection attempt for <jid> failed
```

On command `/disconnect`
`prof: DBG: Input received: /disconnect`
And in the console `You are not currently connected.`

PR #1847, opened 2023-05-10
Minor improvements

sjaeckel:
- [x] I ran valgrind when using my new feature

Check commit messages for details.

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -6575,7 +6575,8 @@ cmd_reconnect(ProfWin* window, const char* const command, gchar** args)
     int intval = 0;
     char* err_msg = NULL;
     if (g_strcmp0(value, "now") == 0) {
-        session_reconnect_now();
+        cons_show("Reconnecting now.");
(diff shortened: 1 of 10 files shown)


jubalh (approved):
Interesting stuff!
Code looks good. I would prefer to have a little more text in the commit messages of "fix reconnect" and "window NULL issue" though.
In the first one a little explanation or thought process and in the second maybe a little description  what's happening there.
Sure I know which bug this is but might be that in some years I will forget or someone else tries to figure out what's done..

H3rnand3zzz:
LGTM. Checked `/reconnect now` and auto reconnect after forceful server disconnect. Both cases worked great, without any problems.

jubalh:
> It might be a good idea to add reference to https://github.com/profanity-im/profanity/issues/1846 in the /fix `/reconnect now` commit message.

Agreed.

sjaeckel:
> I would prefer to have a little more text in the commit messages

better?

jubalh:
> better?

perfect!

PR #1848, opened 2023-05-11
Minor improvements

sjaeckel:
- [x] I ran valgrind when using my new feature

Please check commit messages for details.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -371,7 +371,6 @@ AC_CHECK_LIB([expect], [exp_expectl], [AM_CONDITIONAL([HAVE_EXPECT], [true])],
     [AC_MSG_NOTICE([libexpect not found, will not be able to run functional tests])])
 
 ## Default parameters
-AM_CFLAGS="$AM_CFLAGS -Wall -Wno-deprecated-declarations -std=gnu99"
 AM_CFLAGS="$AM_CFLAGS -Wall -Wno-deprecated-declarations -std=gnu99 -ggdb3"
 AM_LDFLAGS="$AM_LDFLAGS -export-dynamic"
 
@@ -380,9 +379,10 @@ AS_IF([test "x$PACKAGE_STATUS" = xdevelopment],
 AS_IF([test "x$PLATFORM" = xosx],
     [AM_CFLAGS="$AM_CFLAGS -Qunused-arguments"])
 
-AM_CFLAGS="$AM_CFLAGS $PTHREAD_CFLAGS $glib_CFLAGS $gio_CFLAGS $curl_CFLAGS ${SQLITE_CFLAGS} $libstrophe_CFLAGS"
+AM_CFLAGS="$AM_CFLAGS $PTHREAD_CFLAGS $glib_CFLAGS $gio_CFLAGS $curl_CFLAGS ${SQLITE_CFLAGS}"
 AM_CFLAGS="$AM_CFLAGS $libnotify_CFLAGS ${GTK_CFLAGS} $python_CFLAGS"
 AM_CFLAGS="$AM_CFLAGS -DTHEMES_PATH=\"\\\"$THEMES_PATH\\\"\" -DICONS_PATH=\"\\\"$ICONS_PATH\\\"\" -DGLOBAL_PYTHON_PLUGINS_PATH=\"\\\"$GLOBAL_PYTHON_PLUGINS_PATH\\\"\" -DGLOBAL_C_PLUGINS_PATH=\"\\\"$GLOBAL_C_PLUGINS_PATH\\\"\""
+AM_CFLAGS="$AM_CFLAGS $CFLAGS"
 LIBS="$glib_LIBS $gio_LIBS $PTHREAD_LIBS $curl_LIBS $libnotify_LIBS $python_LIBS ${GTK_LIBS} ${SQLITE_LIBS} $LIBS"
 
 AC_SUBST(AM_LDFLAGS)
--- a/src/common.h
+++ b/src/common.h
@@ -60,6 +60,9 @@ void auto_free_char(char** str);
 #define STR_MAYBE_NULL(p) (p)
 #endif
 
+/* Our own define of MB_CUR_MAX but this time at compile time */
+#define PROF_MB_CUR_MAX 8
+
 // assume malloc stores at most 8 bytes for size of allocated memory
 // and page size is at least 4KB
 #define READ_BUF_SIZE 4088
--- a/src/common.h
+++ b/src/common.h
@@ -47,6 +47,9 @@
 
 #define ARRAY_SIZE(x) (sizeof(x) / sizeof(x[0]))
 
+#define PROF_STRINGIFY_(n) #n
+#define PROF_STRINGIFY(n)  PROF_STRINGIFY_(n)
+
 void auto_free_gchar(gchar** str);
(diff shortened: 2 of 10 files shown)


src/common.h
```diff
@@ -47,6 +47,9 @@
 
 #define ARRAY_SIZE(x) (sizeof(x) / sizeof(x[0]))
 
+#define PROF_STRINGIFY_(n) #n
```
jubalh, line 50: I don't understand what this is doing.

configure.ac
```diff
@@ -380,9 +379,10 @@ AS_IF([test "x$PACKAGE_STATUS" = xdevelopment],
 AS_IF([test "x$PLATFORM" = xosx],
     [AM_CFLAGS="$AM_CFLAGS -Qunused-arguments"])
 
-AM_CFLAGS="$AM_CFLAGS $PTHREAD_CFLAGS $glib_CFLAGS $gio_CFLAGS $curl_CFLAGS ${SQLITE_CFLAGS} $libstrophe_CFLAGS"
```
jubalh, line 383: do we need the libstrophe cflags? when do they get passed?

src/xmpp/iq.c
```diff
@@ -2761,7 +2761,7 @@ _mam_rsm_id_handler(xmpp_stanza_t* const stanza, void* const userdata)
                 // Convert to iso8601
                 start_str[strlen(start_str) - 3] = '\0';
             }
-            auto_char char* end_str = NULL;
```
jubalh, line 2764: Ouch yes! Missed that in the last review. Because it is actually data->end_datestr.

jubalh (commented):
* `remove VLA & calm Valgrind` is interrsting!
 * I would rename the `clean-up` to something that includes `configure file`.
 * `fix -Werror=maybe-uninitialized` is actually optimizing the code and "fixing" this "error" as a side effect right? I think this is especially note worthy since the error is actually no error as far as I can see but a false positive.

src/common.h
sjaeckel: https://gcc.gnu.org/onlinedocs/cpp/Stringizing.html

configure.ac
sjaeckel: they get put into `CFLAGS` here https://github.com/profanity-im/profanity/blob/f3265565e853ecf993f12093c6a0c94acf3f212e/configure.ac#L155-L156

src/xmpp/iq.c
sjaeckel: nope, it isn't. It's free'd inside `log_database_get_previous_chat()`

H3rnand3zzz:
Could you, please, rename ab3c0d2daae2dfb8cb96a40549af5220b1c3350f to something like `Consider global CFLAGS and use libstrophe CFLAGS`?

sjaeckel:
Wtf, you're rewriting my history and edit crap into it which isn't true?

sjaeckel:
And you even leave the signed-off part in!
I guess I have to pgp sign my commits from now on.

jubalh:
> Wtf, you're rewriting my history and edit crap into it which isn't true?

Sorry I wasn't aware that this was so bad.
I specifcially marked it with my name by "edit by jubalh". Should be clear then that it's not from you, right?

My thinking was that you are probably busy and don't care about these minor changes like naming commits more expressively. But for me it is important to be able to scan the history quickly and also to understand it better in retrospect.

Also things like starting with upper case which is also mentioned in our CONTRIBUTING.md (under "this guideline") which you don't seem to care about.

So my intention was not to bother you with these "details", since I think in last PRs I repeatedly mentioned them but in the new PR it was still happening. So I figured that you must be busy but don't have the energy/time for these "minor things" and it would help you if I just do it myself.

jubalh:
Just to summarize what I changed in this PR was:
* Use upper case at the start of the line
* In the review at https://github.com/profanity-im/profanity/pull/1848#pullrequestreview-1422304062 I mentioned `I would rename the clean-up to something that includes configure file` to have a more descriptive commit message. So I renamed this one to `Consider global CFLAGS and use libstrophe CFLAGS` (without notice anywhere that I did that. Bad..)
* In the review at https://github.com/profanity-im/profanity/pull/1848#pullrequestreview-1422304062 I said `fix -Werror=maybe-uninitialized is actually optimizing the code and "fixing" this "error" as a side effect right? I think this is especially note worthy since the error is actually no error as far as I can see but a false positive.` so I added `Edit by jubalh:
I still think the error message is a false positive. But this commit improves the
code a lot.` to that commit. In this case I highlighted that this edit is done by me and reflects my current understanding and opinion.

The reason I did that was:
* The in-code reviews were resolved. But these comments didn't get any feedback. I assumed that you don't care about them (because in last PRs by others that you also reviewed I made similar improvements suggestions. And because some of the things are mentioned in the contributing.md).
* Through our conversations I assumed that you are quite busy with work and don't have energy/time to spent of descriptions, you want to just improve the code and be done with it.
* The PR was set to "edits from maintainers are allowed"

> I would've bothered much less if what you wrote was true, but this headline is a lie now. (... and the grammatical mistake/typo is still not corrected ;) )
Haha ok bad about the typo :D Which part is not true now?

> Also I doubt I've ever not complied with rewording commit messages if asked.

I didn't want to say that you didn't comply. I'm pretty sure you always did. I only feel like I have to several times point to changes mentioned in contributing.md -> https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53
And I felt like probably this is just annoying nitpicking to you and you just want to improve the actual code. So I didn't want to bother you with it (again).

> IIUC that's only possible if a PR originates from a fork, since we're working on the same repo that's not possible.

That might be but it was your decision to create a branch that is part of this repo (and thus apparently, I wasn't aware, not have this feature). Now that I'm thinking about it it also makes sense that in a branch from the same repo doesn't have "allow edits from maintainers" since it's the default! It's like "this branch belongs to all maintainers" thus "edits from maintainers are allowed". So I think if you wouldn't want that a branch from a fork would be the proper way to signify that. I also remember that in the past I specifically mentioned something similar about creating PRs from a fork instead.

To me it is a branch on the main repo belongs to everybody on the main repo.

Another part of why I did what I did was: it is morning and I shouldn't have done this decision so early. Seems like then it's easy to overlook things or make bad decisions. I was asked if I could merge this PR soon, so I thought "ok let's just do those things myself. Seems like everybody will be more happy with this solution. The branch will be merged soon and sjaeckel doesn't have to bother about this stuff that he doesn't care about (mentioned reasons above).

To me it still feels like there were many indicators that these edits would be okay. But in the future I will just ping people more about making the edits themselves or ask them if I should make them.

sjaeckel:
> Which part is not true now?

`and use libstrophe CFLAGS`

>  I only feel like I have to several times point to changes mentioned in contributing.md

As already said: if there's no tool to enforce it, it won't happen (consistently).

> And I felt like probably this is just annoying nitpicking to you and you just want to improve the actual code. So I didn't want to bother you with it (again).

IMO this also belongs to a PR and the PR shouldn't be merged until everything is done by the author. (Sure there are cases where the author disappears, but that's something different.)

> [...] it is morning and [it seems] like then it's easy to overlook things or make bad decisions.

I couldn't agree more.

> [...] I will just ping people more about making the edits themselves or ask them if I should make them

:+1:

PR #1850, opened 2023-05-13
Extend `/pgp` command to make key exchange procedure easier

H3rnand3zzz:
See commit messages


jubalh:
Hi! Didn't have time to review yet, only skimmed it. But I would suggest (for the future if you don't want to do it for this PR) to split the commits.
I think this should actually be two or three commits :)

The reason is that it makes it easier to revert commits. Let's say for some (imagined) reason we would like to remove the pgp exchange procedure. Then we could just do `git revert xxx`. But if all three changes are in one commit we would also loose the typo fixes and the improvements not directly related to the exchange procedure. So that means we would have to revert by hand.

So it's good practise to have such things in individual commits. Even when it's unlikely that a revert will happen :)

src/command/cmd_defs.c
```diff
@@ -1719,7 +1720,9 @@ static const struct cmd_t command_defs[] = {
               { "log on|off", "Enable or disable plaintext logging of PGP encrypted messages." },
               { "log redact", "Log PGP encrypted messages, but replace the contents with [redacted]. This is the default." },
               { "char <char>", "Set the character to be displayed next to PGP encrypted messages." },
-              { "sendfile on|off", "Allow /sendfile to send unencrypted files while otherwise using PGP." })
+              { "sendfile on|off", "Allow /sendfile to send unencrypted files while otherwise using PGP." },
+              { "autoimport on|off", "Autoimport PGP keys from messages." },
```
jubalh, line 1724: I'm actually not sure but maybe we should mention here "This is compatible with PSI+"? But really not sure if this is good or not. Usually things are described in a XEP but this isn't so I thought maybe it's interesting for users which other clients work with this. On the other hand I for some reason don't like to tell which other clients support something since we have to keep this information up to date..

src/command/cmd_ac.c
```diff
@@ -920,6 +923,10 @@ cmd_ac_init(void)
     autocomplete_add(pgp_sendfile_ac, "on");
     autocomplete_add(pgp_sendfile_ac, "off");
 
+    pgp_autoimport_ac = autocomplete_new();
```
jubalh, line 926: For autocompletes that only support on/off you can actually use `_cmd_ac_complete_params()`.

src/command/cmd_defs.c
jubalh: Well I think lets not do that. It might become outdated. Especially on some distros that updates every few decades.

jubalh:
@sjaeckel ping

src/command/cmd_funcs.c
```diff
@@ -8606,24 +8668,37 @@ _update_presence(const resource_presence_t resource_presence,
...
-        cons_show("%s enabled.", display);
+    }
+
+    if (g_ascii_strcasecmp(arg, "on") == 0) {
+        cons_show("%s %senabled.", display, prev_state ? "is already " : "");
         prefs_set_boolean(pref, TRUE);
-    } else if (g_strcmp0(arg, "off") == 0) {
-        cons_show("%s disabled.", display);
+    } else if (g_ascii_strcasecmp(arg, "off") == 0) {
+        cons_show("%s %sdisabled.", display, !prev_state ? "is already " : "");
```
sjaeckel, line 8695: @jubalh prefers case sensitivity for arguments: https://github.com/profanity-im/profanity/pull/1780#discussion_r1039389092 :shrug:

src/pgp/gpg.c
```diff
@@ -303,67 +307,62 @@ p_gpg_free_key(ProfPGPKey* key)
...
+ *
+ * @note This function may perform additional operations, such as autocomplete, related to the retrieved keys.
+ */
 GHashTable*
 p_gpg_list_keys(void)
 {
     gpgme_error_t error;
     GHashTable* result = g_hash_table_new_full(g_str_hash, g_str_equal, free, (GDestroyNotify)p_gpg_free_key);
 
+    // Create a GPGME context
```
sjaeckel, line 331: https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/

> `Rule 1: Comments should not duplicate the code.`

This applies to all comments in the code you added in this PR that I've found so far.

sjaeckel (changes requested):

jubalh:
@sjaeckel I'm uncertain if you plan to review again :)
Otherwise I would merge.

src/command/cmd_funcs.c
```diff
@@ -7579,7 +7589,59 @@ cmd_pgp(ProfWin* window, const char* const command, gchar** args)
...
+            assert(chatwin->memcheck == PROFCHATWIN_MEMCHECK);
+        }
+
+        ProfAccount* account = accounts_get_account(session_get_account_name());
+
+        if (account->pgp_keyid == NULL) {
+            cons_show_error("Please, set the PGP key first using /account set %s pgpkeyid <pgpkeyid>. To list pgp keys, use /pgp keys.", account->name);
+            account_free(account);
+            return TRUE;
+        }
```
sjaeckel, line 7633: Would it maybe make more sense to do this check before changing the window focus?

src/command/cmd_funcs.c
```diff
@@ -7533,14 +7540,17 @@ cmd_pgp(ProfWin* window, const char* const command, gchar** args)
         }
 
         ProfAccount* account = accounts_get_account(session_get_account_name());
-        char* err_str = NULL;
+        auto_char char* err_str = NULL;
```
sjaeckel, line 7543: This declaration could/should go one block down, before its actual usage ... really minor ...

src/pgp/gpg.c
```diff
@@ -769,6 +750,200 @@ p_gpg_format_fp_str(char* fp)
...
+    }
+
+    gpgme_data_t key_data;
+    error = gpgme_data_new_from_mem(&key_data, buffer, strlen(buffer), 1);
+    if (error != GPG_ERR_NO_ERROR) {
+        log_error("GPG: Error creating GPGME data from buffer");
+        goto out;
+    }
+
+    gpgme_import_result_t import_result;
```
sjaeckel, line 850: This declaration could/should should also go down to where it's actually used first.

src/pgp/gpg.c
```diff
@@ -769,6 +750,200 @@ p_gpg_format_fp_str(char* fp)
...
+ * Convert a gpgme_data_t object to a null-terminated char* string.
+ *
+ * @param data The gpgme_data_t object to convert.
+ * @return The converted string allocated using malloc, which should be freed by the caller.
+ *         If an error occurs or the data is empty, NULL is returned and errors are written to the error log.
+ */
+static char*
+_gpgme_data_to_char(gpgme_data_t data)
+{
+    char* buffer = NULL;
```
sjaeckel, line 932: This one as well.

sjaeckel (commented):
I like the refactor of `_cmd_set_boolean_preference()` :+1: 

Would it maybe make sense to have a `/pgp import` command which lets you select one of the last messages in a private chat and then tries to import that? ... in case a user doesn't want to auto-import all public keys. But that'd be for another PR I guess.

src/command/cmd_funcs.c
H3rnand3zzz: window focus?

H3rnand3zzz:
@jubalh review suggestions are applied. After almost 2 month in waiting, maybe it's time to merge?

jubalh (changes requested):
Thanks for your patience!

So far I only looked at the code but now realize there could be a change in the commit messages as well.

In the first commit (Add /pgp sendpub command ) I'm missing a description what the command will do/why it is useful. And also with which other clients it is compatible (if it applies and you are aware of this info. I think you said it works with PSI).

Same for `Add optional pgp public key autoimport`.

I think it deserves a short description of what it will do/why its wanted.

After this I think good to go. Sorry again for the long waiting time!

src/command/cmd_funcs.c
```diff
@@ -7579,7 +7589,60 @@ cmd_pgp(ProfWin* window, const char* const command, gchar** args)
...
+        }
+
+        ProfAccount* account = accounts_get_account(session_get_account_name());
+
+        if (account->pgp_keyid == NULL) {
+            cons_show_error("Please, set the PGP key first using /account set %s pgpkeyid <pgpkeyid>. To list pgp keys, use /pgp keys.", account->name);
+            account_free(account);
+            return TRUE;
+        }
+        cons_show("account->pgp_keyid = %s", account->pgp_keyid);
```
sjaeckel, line 7615: That's more like a debug statement, so this can go. Or should it stay? Then please give it a more descriptive message that's understandable for a regular user.

sjaeckel (changes requested):
Sorry for the next bunch of change requests, but this PR is so huge, it's not easy to review. Also there are two sets of independent changes in there, which makes it even harder. That's really where the PR focused review approach falls short. If this were a commit based approach it'd be cleaner, but it is as it is and therefore :shrug: 

Still I suggest to rebase on top of master and re-order the commits. Put "Refactor `_cmd_set_boolean_preference` function" as the first commit and create a separate branch & PR for that single commit, which can then be quickly merged, so the "real changes" of this feature become visible.

src/pgp/gpg.c
```diff
@@ -381,6 +375,7 @@ p_gpg_list_keys(void)
 
     gpgme_release(ctx);
 
+    // TODO: move autocomplete in other place
```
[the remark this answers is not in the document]
H3rnand3zzz: It might be considered unintended functionality, the function is a bit too bloated.

src/command/cmd_funcs.c
H3rnand3zzz: Debug statement, I guess.

sjaeckel:
> Still I suggest to rebase on top of master and re-order the commits. Put "Refactor `_cmd_set_boolean_preference` function" as the first commit and create a separate branch & PR for that single commit, which can then be quickly merged, so the "real changes" of this feature become visible.

If you open a separate PR for those unrelated changes, maybe also include [Cleanup char* to auto_gchar gchar* for prefs_get_string](https://github.com/profanity-im/profanity/pull/1857/commits/be88cf9d29d924e873507e58f4447a09dde77d49)"and [Fix memory leak and cleanup](https://github.com/profanity-im/profanity/pull/1857/commits/f98e9b6415e7cfd1e1638c4c6271d4cf6fa9fa45) of #1857

src/pgp/gpg.c
```diff
@@ -431,13 +426,7 @@ p_gpg_valid_key(const char* const keyid, char** err_str)
...
     if (error || key == NULL) {
         log_error("GPG: Failed to get key. %s %s", gpgme_strsource(error), gpgme_strerror(error));
-        *err_str = strdup(gpgme_strerror(error));
-        gpgme_release(ctx);
-        return FALSE;
-    }
-
-    if (key == NULL) {
-        *err_str = strdup("Unknown error");
+        *err_str = error ? strdup(gpgme_strerror(error)) : "Unknown error";
```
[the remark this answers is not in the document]
H3rnand3zzz: Thanks, valid point, fixed.

src/pgp/gpg.c
```diff
@@ -769,6 +750,198 @@ p_gpg_format_fp_str(char* fp)
...
+{
+    size_t buffer_size = 0;
+    char* gpgme_buffer = gpgme_data_release_and_get_mem(data, &buffer_size);
+
+    if (!gpgme_buffer) {
+        log_error("GPG: Unable to extract gpgmedata.");
+        return NULL;
+    }
+
+    char* buffer = strndup(gpgme_buffer, buffer_size);
```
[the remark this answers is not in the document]
H3rnand3zzz: We don't return buffer size, it varies, so it doesn't make sense, as later it's only used as a null-terminated string.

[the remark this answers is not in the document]
sjaeckel: > We don't return buffer size, it varies, so it doesn't make sense

What do you mean by 'it'?

The idea behind the change is that the `strndup()` approach will take longer than the `memcpy()` one since it has to look at each byte whether it terminates the string, but that's not needed.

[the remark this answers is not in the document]
H3rnand3zzz: applied

src/pgp/gpg.c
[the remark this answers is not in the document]
H3rnand3zzz: in chat @sjaeckel pointed out another case with (un)conditional error assignment, I fixed it out as well.

jubalh (approved):
That's great descriptive commit messages now! Thanks :)

sjaeckel (approved):

jubalh:
Thanks for your contribution!

H3rnand3zzz:
Thank you for reviews

Issue #1819, opened 2023-04-05
`gchar*` vs. `char*`


sjaeckel:
Currently there's a mixed usage of GLib and stdlib API's when dealing with strings. In general this also applies to other types, but I'm not aware of any other cases where type casts from GLib to stdlib happen or would matter.

There are some functions using GLib API's but treating the returned value as `char*`.

This will cause problems if the memory allocator of GLib is different than standard `malloc()`, which apparently can be the case if running on Windows 

[glib/gmem.c:39-52](https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gmem.c#L39-L52)
```
#if defined(HAVE_MEMALIGN) || defined(HAVE__ALIGNED_MALLOC)
/* Required for _aligned_malloc() and _aligned_free() on Windows */
#include <malloc.h>
#endif

#ifdef HAVE__ALIGNED_MALLOC
/* _aligned_malloc() takes parameters of aligned_malloc() in reverse order */
# define aligned_alloc(alignment, size) _aligned_malloc (size, alignment)

/* _aligned_malloc()'ed memory must be freed by _align_free() on MSVC */
# define aligned_free(x) _aligned_free (x)
#else
# define aligned_free(x) free (x)
#endif
```

This is a huge task, as it involves touching most of the profanity code base, so it will have to be done over a longer stretch of time.

Another option to tackle this, could also be to define a rule, that we'll never make profanity build on platforms which don't use `malloc()` in the GLib memory allocator ;)

H3rnand3zzz:
Can you, please, provide details on correct usage with examples? Theoretically, we can catch all the mistakes made even with regex and fix it.

And I think that we should add these details into documentation since "auto_char" is something unique, as far as my knowledge goes.

H3rnand3zzz:
`[^g]char\s?\*.+\sg_`
150+ occurrences

sjaeckel:
> Can you, please, provide details on correct usage with examples? Theoretically, we can catch all the mistakes made even with regex and fix it.

That's a regex I'd like to see :grinning: 

What we have to do is: if we use a GLib API which returns newly allocated memory, make sure that this memory is later on free'd via `g_free()`.

One example module which just came to my mind and violates this in some places is `src/config/preferences.c`, e.g. here:

https://github.com/profanity-im/profanity/blob/5f078f95f2b0d77f1f94fbe69aaa0b6ae605b398/src/config/preferences.c#L904-L911

[...]

PR #1858, opened 2023-06-15
Major cleanup

H3rnand3zzz:
The changes are intended to fix most of the cases of #1819 (incorrect usage of `gchar` instead of `char` and vice versa leading to incorrect memory `free`ing).

- [x] Fix memory leaks/remove markers in a new commit
- [x] Fix double free's introduced in changes
- [x] Second commit for `prefs_get_string` as `gchar`
- [x] Third commit `auto_gchar` instead of `g_free` when suitable.
- [x] Fourth commit for jids
- [x] Fifth commit for gcharv
- [x] Arch check
- [x] Self-review


sjaeckel:
I really like how this is evolving!

Just as a heads up: IMO this PR gets too huge and must be split up into multiple ones. But only when you're finished and I guess the best would be 1 PR per commit as they're already separating the topics. (Minus the fixup commit(s) which should be squashed into the appropriate topic ones). 👍
... But maybe I'm also telling you what you had already planned to do 😆

H3rnand3zzz:
Since it's going to be split up, I will keep it "drafted" and then close/something. Please, don't review this one.

jubalh:
I'm not quite certain what's the plan here.
But if this is going to be split then the most important part should be this:

A commit should do one thing. For example:
```
[Cleanup prefs_get_string](https://github.com/profanity-im/profanity/pull/1858/commits/b57d9a5ef022cb4c2d39cb550a27610ecbbeac7c)

Improve usage of `gchar` and `char` by setting them correctly
Increase usage of `auto_gchar` and `auto_char`
Fix 2 mem leaks (rosterwin.c, avatar.c)
Improve documentation
```

is doing a lot of unrelated things. And I don't quite see why they are in one commit. This makes it hard to review. To pick and choose changes.


Similar for:

```
[Big cleanup char to auto_char](https://github.com/profanity-im/profanity/pull/1858/commits/5207b275e180dcd194aea813540396f90c965b51)

Replace `gchar` and `g_free` to `auto_gchar`
Improve readability
Improve documentation for certain functions to increase transparency
Correct certain  `char` functions/variables to `gchar`
Fix most of the cases for 1819 issue
```

I think you need a commit
* to change char to auto_char
* fix memleaks (if unrelated to the other commit)
* change jid to auto_jid
* improve documentation
* change g_strfreev() to auto_gcharv

something like this. I'm not sure if i missed something because really its hard to read :)

And for me it would be okay if such commits would be in one cleanup PR. But sure if this touches too many lines/code then it can also help to split it to several PRs. What I mean is that the one atomic change per commit is the more important splitting.

jubalh:
For example in https://github.com/profanity-im/profanity/pull/1863 you now have only one commit. But you do 2 things there:
* Improve the documentation and change the variable name of a function
* Change from char to auto_char to automatically free it.

These things are not related and should be separate commits.

jubalh:
Your other PR got merged. First commit can be dropped I guess.
Then the rest cleaned up. You dont need a new PR for this. I will review here. It shouldnt be too many lines left after the chat->autochar change.

src/config/files.c
```diff
@@ -155,15 +155,6 @@ files_get_config_path(const char* const config_base)
...
 
-/**
- * Get the full path by appending the given location to the project base path.
- *
- * @param location The location (directory or file) to append to the project base path.
- * @return The full path obtained by appending the location to the project base path.
- *
- * @note The returned value must be freed using g_free when it is no longer needed
- *       to prevent memory leaks.
- */
```
jubalh, line 166: Comment about function should be in the .c file and not the header.
The dev is reading the .c file and shouldnt have to switch to the header just to read the comment.
Maybe some IDEs take it from the .h file but I'm not a fan of this.
True for all other cases in this commit as well.

And please break up this commit. The docu change should be a commit of its own. And not mixed with the char one.

jubalh (changes requested):
Again please dont mix in one commit several things that dont belong together.

Split out the change int he first (two?) commits.

And put the description of the functions above the functions in the .c file.
Also please be aware that we decided in https://github.com/profanity-im/profanity/issues/1859 that we don't want doxygen comments.

So far I accepted them in cases where I believed that our function name is really not descriptive enough and your comments were a good explanation.
I would still prefer them written in another way but its also fine by me to have them like this. But the goal will not be to have doxygen comments for each function.

H3rnand3zzz:
> Looks good! But there are still function description comments in one .h file :)

it was for macros, since they are used only in .h file

src/config/preferences.c
```diff
@@ -565,15 +579,21 @@ prefs_get_string_with_option(preference_t pref, gchar* option)
...
 
+/**
+ * @brief Sets or deletes a string value for the given preference in the preference file.
+ *
+ * @param pref      The preference to set the value for.
+ * @param new_value The new string value to set. Pass NULL to remove the key.
+ */
 void
-prefs_set_string(preference_t pref, char* value)
+prefs_set_string(preference_t pref, gchar* new_value)
```
jubalh, line 589: I think `value` was quite descriptive ;)

src/config/preferences.c
```diff
@@ -540,14 +545,23 @@ prefs_get_string(preference_t pref)
...
+ * @brief Retrieves a localized string preference value.
+ *
+ * @param pref   The preference to retrieve the value for.
+ * @param locale The option to consider.
+ * @return       Returns the value associated with pref translated in the given locale if available.
+ *
+ * @note Remember to free the returned string using `auto_gchar` or `g_free()`.
+ */
 gchar*
-prefs_get_string_with_option(preference_t pref, gchar* option)
```
jubalh, line 544: It looks like we even don't use this function anymore?

jubalh (approved):
Huge PR to review. I hope I didn't miss anything. LGTM.

jubalh:
> it was for macros, since they are used only in .h file

I see! My bad!

PR #1865, opened 2023-07-10
Fix OMEMO autodetection in autotools

pasis:
This commit fixes few issues related to OMEMO autodetection:

1. Absence of libsignal-protocol-c doesn't turn OMEMO off, just prints a notice message. It should (a) set BUILD_OMEMO=false and (b) terminate with an error on --enable-omemo=yes.
2. Check for gcrypt continues even if libsignal-protocol-c fails. In this case, LIBS variable can be updated with unneeded -lgcrypt.
3. Similarly to item 2., if libsignal-protocol-c is present, but gcrypt isn't, variable LIBS is updated with unneeded library.

* step 1 install libgcrypt
* step 2 remove libsignal-protocol-c
* step 3 configure and build

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -323,19 +323,20 @@ AS_IF([test "x$enable_gdk_pixbuf" != xno],
 dnl feature: omemo
 AM_CONDITIONAL([BUILD_OMEMO], [false])
 if test "x$enable_omemo" != xno; then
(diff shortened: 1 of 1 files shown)


jubalh (approved):

jubalh:
Thanks!

PR #1867, opened 2023-07-13
Add PEP nicknames from XEP-0172

bodqhrohro:
* Setup telegabber
* Switch to the roster window
* Change some chat name at the Telegram side
* Look if it's changed in the roster

# I ran valgrind when using my new feature
```
==4147910== Memcheck, a memory error detector
==4147910== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==4147910== Using Valgrind-3.19.0 and LibVEX; rerun with -h for copyright info
==4147910== Command: profanity
==4147910== 
==4147910== 
==4147910== HEAP SUMMARY:
==4147910==     in use at exit: 1,849,573 bytes in 1,554 blocks
==4147910==   total heap usage: 1,551,248 allocs, 1,549,694 frees, 87,668,537 bytes allocated
==4147910== 
==4147910== LEAK SUMMARY:
==4147910==    definitely lost: 1,080 bytes in 15 blocks
==4147910==    indirectly lost: 2,125 bytes in 41 blocks
==4147910==      possibly lost: 832 bytes in 14 blocks
==4147910==    still reachable: 1,842,968 bytes in 1,457 blocks
==4147910==         suppressed: 0 bytes in 0 blocks
==4147910== Rerun with --leak-check=full to see details of leaked memory
==4147910== 
==4147910== For lists of detected and suppressed errors, rerun with: -s
==4147910== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

### DIFF
--- a/src/xmpp/nickname.c
+++ b/src/xmpp/nickname.c
@@ -1,0 +1,74 @@
+/*
+ * nickname.c
+ * vim: expandtab:ts=4:sts=4:sw=4
+ *
+ * Copyright (C) 2023 Bohdan Horbesko <bodqhrohro@gmail.com>
+ *
+ * This file is part of Profanity.
+ *
+ * Profanity is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Profanity is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Profanity.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * In addition, as a special exception, the copyright holders give permission to
+ * link the code of portions of this program with the OpenSSL library under
+ * certain conditions as described in each individual source file, and
+ * distribute linked combinations including the two.
+ *
+ * You must obey the GNU General Public License in all respects for all of the
+ * code used other than OpenSSL. If you modify file(s) with this exception, you
+ * may extend this exception to your version of the file(s), but you are not
+ * obligated to do so. If you do not wish to do so, delete this exception
+ * statement from your version. If you delete this exception statement from all
+ * source files in the program, then also delete it here.
+ *
+ */
+
+#include "xmpp/message.h"
+#include "xmpp/roster_list.h"
+#include "xmpp/stanza.h"
+#include "ui/ui.h"
+
+static int _pep_nick_handler(xmpp_stanza_t* const stanza, void* const userdata);
+
+void
+nickname_pep_subscribe(void)
+{
+    message_pubsub_event_handler_add(STANZA_NS_NICK, _pep_nick_handler, NULL, NULL);
+}
+
+static int
+_pep_nick_handler(xmpp_stanza_t* const stanza, void* const userdata)
+{
+    const char* from = xmpp_stanza_get_attribute(stanza, STANZA_ATTR_FROM);
+    if (from) {
+        xmpp_stanza_t* event = xmpp_stanza_get_child_by_name_and_ns(stanza, STANZA_NAME_EVENT, STANZA_NS_PUBSUB_EVENT);
+        if (event) {
+            xmpp_stanza_t* items = xmpp_stanza_get_child_by_name(event, STANZA_NAME_ITEMS);
+            if (items) {
+                xmpp_stanza_t* item = xmpp_stanza_get_child_by_name(items, STANZA_NAME_ITEM);
+                if (item) {
+                    xmpp_stanza_t* nick = xmpp_stanza_get_child_by_name_and_ns(item, STANZA_NAME_NICK, STANZA_NS_NICK);
+                    if (nick) {
+                        auto_char char* text = xmpp_stanza_get_text(nick);
+
+                        PContact contact = roster_get_contact(from);
+                        if (contact) {
+                            roster_change_name(contact, text);
+                        }
+                    }
+                }
+            }
+        }
+    }
+    return 1;
+}
--- a/Makefile.am
+++ b/Makefile.am
@@ -21,6 +21,7 @@ core_sources = \
 	src/xmpp/avatar.c src/xmpp/avatar.h \
 	src/xmpp/ox.c src/xmpp/ox.h \
 	src/xmpp/vcard.c src/xmpp/vcard.h src/xmpp/vcard_funcs.h \
+	src/xmpp/nickname.c src/xmpp/nickname.h \
 	src/event/common.c src/event/common.h \
 	src/event/server_events.c src/event/server_events.h \
 	src/event/client_events.c src/event/client_events.h \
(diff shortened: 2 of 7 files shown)


src/xmpp/nickname.c
```diff
@@ -0,0 +1,74 @@
...
+{
+    const char* from = xmpp_stanza_get_attribute(stanza, STANZA_ATTR_FROM);
+    if (from) {
+        xmpp_stanza_t* event = xmpp_stanza_get_child_by_name_and_ns(stanza, STANZA_NAME_EVENT, STANZA_NS_PUBSUB_EVENT);
+        if (event) {
+            xmpp_stanza_t* items = xmpp_stanza_get_child_by_name(event, STANZA_NAME_ITEMS);
+            if (items) {
+                xmpp_stanza_t* item = xmpp_stanza_get_child_by_name(items, STANZA_NAME_ITEM);
+                if (item) {
+                    xmpp_stanza_t* nick = xmpp_stanza_get_child_by_name_and_ns(item, STANZA_NAME_NICK, STANZA_NS_NICK);
```
sjaeckel, line 60: This could be simplified by using `xmpp_stanza_get_child_by_path()`
```suggestion
        xmpp_stanza_t* nick = xmpp_stanza_get_child_by_path(stanza, XMPP_STANZA_NAME_IN_NS(STANZA_NAME_EVENT, STANZA_NS_PUBSUB_EVENT), STANZA_NAME_ITEMS, STANZA_NAME_ITEM, XMPP_STANZA_NAME_IN_NS(STANZA_NAME_NICK, STANZA_NS_NICK), NULL);
```

sjaeckel (changes requested):
:+1: 

Will you also add the possibility to set such a nickname? as FMU now you'd have to use a different client to set it.

> `==4147910==    definitely lost: 1,080 bytes in 15 blocks`

where do those lost blocks originate? I'm mostly certain they're not lost in your feature, but still it'd be nice to know. Feel free to attach a logfile.

bodqhrohro:
> Feel free to attach a logfile.

[prof1.txt](https://github.com/profanity-im/profanity/files/12036005/prof1.txt)

src/xmpp/nickname.c
bodqhrohro: Nah, `nick` is NULL this way, and it's non-trivial to debug where exactly does it fail (looks OK to me, huh :/), so I'll probably leave it intact for now.

jubalh (changes requested):
Thanks for your feature!

Regarding CI:
* Please check https://github.com/profanity-im/profanity/blob/master/CONTRIBUTING.md on how to run clang format.
* You need to add a nickname_pep_subscribe (stub) to the tests. You can grep for avatar_pep_subscribe for example as inspiration

jubalh:
A couple of questions:

How should this play together with `/nick` (setting your own nickname in a MUC) `/roster nick` (assigning a nickname to a user in your roster). Should a nick sent by the contact or the nick I assign via `/roster nick` have a higher priority?

> Will you also add the possibility to set such a nickname? as FMU now you'd have to use a different client to set it.

?

jubalh:
From our MUC:
```
jubalh: pep.: do you support xep0172 in poezio? I would have some questions how to handle it: https://github.com/profanity-im/profanity/pull/1867#issuecomment-1633801969
pep.: I think we do but not really. I'd like to. I started something here about it: https://codeberg.org/joinjabber/support/issues/6
pep.: But for your question,  
pep.: If a nick is set manually in a MUC I'd use that rather than the global set nick
pep.: Same for the roster.
pep.: In general, if a user has done something manually for a more detailed level I'd use that rather than the global setting (172)
jubalh: pep.: I agree. Thanks for your input!
```

So I think we need to check if the user did set a nickname for person in their roster and only if they didn't use the nick that we get over this XEP. Same for MUC.

Could you add such functionality?

And do you plan to also implement the setting of the nick?

If you need any help or guidance don't hesitate to ask here or contact us in the MUC.

bodqhrohro:
> How should this play together with `/nick` (setting your own nickname in a MUC) `/roster nick` (assigning a nickname to a user in your roster). Should a nick sent by the contact or the nick I assign via `/roster nick` have a higher priority?

From the experience with my transport, users have different expectations. Some indeed prefer to set contact nicknames manually, and are frustrated when they are updated automatically. Other users are rather frustrated when they don't see relevant updates of the nickname (especially if they have saved a nickname provided in a subscription request, which in some clients is equal to setting it manually).

I suppose it should be a configuration option then, and the default should be to respect the nickname from the roster first then, as recommended by Modern XMPP.

> And do you plan to also implement the setting of the nick?

Maybe, but later. Setting the nickname via PubSub seems to need some server-side support, so I'll need to prepare it for testing first.

jubalh:
> I suppose it should be a configuration option then, and the default should be to respect the nickname from the roster first then, as recommended by Modern XMPP.

Yes configuration option would be good.

Taking roster nick in MUC doesn't make much sense to me since it could be that I want to reply to a user and write "John Tolkien: how are you doing?" because in my roster he is named like that. But everyone else sees his nickname as "JRRT" because that's what he set when joining.

> Maybe, but later. Setting the nickname via PubSub seems to need some server-side support, so I'll need to prepare it for testing first.

Ok, great.

H3rnand3zzz (changes requested):
Please, fix the nesting, add XEP support information to .doap file and disco (XEP-0030, XEP-0115, if it's recommended/required by the XEP that you are adding).

Thank you for your contribution.

src/xmpp/nickname.c
sjaeckel: Indeed, as I forgot that you have to give the full path, i.e. this should work:

```suggestion
        xmpp_stanza_t* nick = xmpp_stanza_get_child_by_path(stanza, STANZA_NAME_MESSAGE, XMPP_STANZA_NAME_IN_NS(STANZA_NAME_EVENT, STANZA_NS_PUBSUB_EVENT), STANZA_NAME_ITEMS, STANZA_NAME_ITEM, XMPP_STANZA_NAME_IN_NS(STANZA_NAME_NICK, STANZA_NS_NICK), NULL);
```

jubalh:
@bodqhrohro ping :)

bodqhrohro:
I see no real-life use for setting the nickname via PubSub and thus not sure what to do and why.

I forgot that my transport also sends the nickname in presences, and I implemented PubSub later than that specifically for Gajim as Gajim ignores the nickname in presences and uses only nicknames from PubSub. The person who I made this patch for finally implemented their own patch which takes the nickname from presences, as they found it easier; not sure if they intend to merge it upstream at all. Later I discovered that this actually violates the standard, as it specifies that regular non-subscribe presences MUST NOT include the nick.

jubalh:
Do I understand you right that you think your PR is obsolete?

bodqhrohro:
I'm not sure about that, as the other patch violates the XEP, but as of now it Just Works™. PEP is also not a great solution for the task, given that the roster name is anyway the highest priority according to modernxmpp spec, yet it works in Gajim. Slidge implements roster sync via a special privilege, and we probably would go this way eventually too (but that won't work for the users from other servers).

As of now, I'll probably close this PR and reopen it later if we come up with some decision. When more important features are implemented, I'll anyway return to the problem of presence flood, as they consume traffic excessively, and those out-of-spec nickname presences are a part of it.

jubalh:
> As of now, I'll probably close this PR and reopen it later if we come up with some decision. 

Feel free to reopen or create a new patch when you feel like it!
All improvements are welcome.

PR #1869, opened 2023-07-18
Fix crash if we don't get `first` stanza during RSM

jubalh:


sjaeckel (approved):
As indicated by XEP 0059, the `first` element is optional, so this patch fixes a bug.

Issue #1880, opened 2023-08-27
`/plugins install` without connected account crashes


jubalh:
Reason is: https://github.com/profanity-im/profanity/commit/3a86b8c29
Created to fix: https://github.com/profanity-im/profanity/issues/1624

It wants to get the tls setting based on account but doesn't check whether account != NULL.

Options:
1) `/plugins install` should only work when we are connected and respect the tls setting according to account
2) it should take the tls setting from the default account and if none is set cancel
3) it should always validate the cert. Since it's actually an unrelated decision whether the user wants to validate the certificate of his xmpp server vs any web server where he wants to install his plugins from
4) a new setting

Personally I like 1 the most.

jubalh:
@sjaeckel opinion? :)

sjaeckel:
> @sjaeckel opinion? :)

c.f. https://github.com/profanity-im/profanity/pull/1882#pullrequestreview-1597717430 :)

PR #1882, opened 2023-08-28
Don't crash when using `/plugins install` disconnected

### LINKED ISSUE #1624: /sendfile does not work with self signed cert

jubalh:
When we are not connected and run `/plugins install` we crash because we get the account struct to check for the (xmpp) tls setting. To apply that to the http (etc) connection to download the plugin from a server.

This got introduced in 3a86b8c29 to fix #1624.

There are several ways to handle this (some described in 1880) in this patch I took the route that it will use secure connection when we are nto connected and will only check the tls.trust account setting if we are connected.

Fix https://github.com/profanity-im/profanity/issues/1880

I ran valgrind when using my new feature
no

### DIFF
--- a/src/tools/http_download.c
+++ b/src/tools/http_download.c
@@ -134,7 +134,10 @@ http_file_get(void* userdata)
     gchar* cert_path = prefs_get_string(PREF_TLS_CERTPATH);
     gchar* cafile = cafile_get_name();
     ProfAccount* account = accounts_get_account(session_get_account_name());
-    gboolean insecure = account->tls_policy && strcmp(account->tls_policy, "trust") == 0;
+    gboolean insecure = FALSE;
+    if (account) {
+        insecure = account->tls_policy && strcmp(account->tls_policy, "trust") == 0;
+    }
     account_free(account);
     pthread_mutex_unlock(&lock);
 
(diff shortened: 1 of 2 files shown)


sjaeckel (approved):
I like that approach!

Issue #1877, opened 2023-08-23
Profanity fails to properly detect URLs wrapped in <>


mdosch:
If an URL is wrapped in <>, e.g.:

```
Fürst Pückler bezeichnet: (<https://de.wikipedia.org/wiki/F%C3%BCrst_P%C3%BCckler>)
```

profanitys `/url open <tab>` will try expand to `/url open https://de.wikipedia.org/wiki/F%C3%BCrst_P%C3%BCckler>)`.

I think `<` and `>` should not be considered as part of an URL as they are unsafe characters that must be escaped:

> Characters can be unsafe for a number of reasons.  The space
> character is unsafe because significant spaces may disappear and
> insignificant spaces may be introduced when URLs are transcribed or
> typeset or subjected to the treatment of word-processing programs.
> The characters "<" and ">" are unsafe because they are used as the
> delimiters around URLs in free text; […]
https://datatracker.ietf.org/doc/html/rfc1738

## Expected Behavior

`/url open <tab>` with the previous example should expand to `/url open https://de.wikipedia.org/wiki/F%C3%BCrst_P%C3%BCckler`.

## Current Behavior

`/url open <tab>` expands to `/url open https://de.wikipedia.org/wiki/F%C3%BCrst_P%C3%BCckler>)`.

## Steps to Reproduce (for bugs)

1. See explanation at the beginning.

## Environment
* Debian Testing/Unstable

```
profanity --version
Profanity, version 0.14.0dev.master.492efc43
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2023 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Disabled
PGP support: Enabled (libgpgme 1.18.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Disabled
GTK icons/clipboard: Disabled
GDK Pixbuf: Enabled
```

jubalh:
[...]

PR #1883, opened 2023-08-28
Cut URLs before adding to automcomp at `>`

jubalh:
First I tried with g_uri_parse() and g_uri_to_string() but then I learned that GUri validation API is only for things that are part of a proper URL.

Let's cut the string at `>` since they are sometimes enclosed in `<>`.

Fix https://github.com/profanity-im/profanity/issues/1877

### DIFF
--- a/src/ui/window_list.c
+++ b/src/ui/window_list.c
@@ -1310,7 +1310,8 @@ wins_add_urls_ac(const ProfWin* const win, const ProfMessage* const message, con
     GRegex* regex;
     GMatchInfo* match_info;
 
-    regex = g_regex_new("(https?|aesgcm)://\\S+", 0, 0, NULL);
+    // https://stackoverflow.com/questions/43588699/regex-for-matching-any-url-character
+    regex = g_regex_new("(https?|aesgcm)://[\\w\\-.~:/?#\\[\\]@!$&'()*+,;=%]+", 0, 0, NULL);
     g_regex_match(regex, message->plain, 0, &match_info);
 
     while (g_match_info_matches(match_info)) {


jubalh:
Maybe adjusting the regex used at https://github.com/profanity-im/profanity/pull/1883/files#diff-1a7db58e738a75a1ee3630328ced5287e85e1f30168c13d6d1d680d5f7c2a830L1313 would be better. If anyone is a regex master here .. ;)

sjaeckel:
> Maybe adjusting the regex used at https://github.com/profanity-im/profanity/pull/1883/files#diff-1a7db58e738a75a1ee3630328ced5287e85e1f30168c13d6d1d680d5f7c2a830L1313 would be better. If anyone is a regex master here .. ;)

... now we have two problems ... 

```
$ git diff -U1
diff --git a/src/ui/window_list.c b/src/ui/window_list.c
index a00fcb43..2206c6d8 100644
--- a/src/ui/window_list.c
+++ b/src/ui/window_list.c
@@ -1312,3 +1312,3 @@ wins_add_urls_ac(const ProfWin* const win, const ProfMessage* const message, con
 
-    regex = g_regex_new("(https?|aesgcm)://\\S+", 0, 0, NULL);
+    regex = g_regex_new("(https?|aesgcm)://[\\w\\-.~:/?#\\[\\]@!$&'()*+,;=%]+", 0, 0, NULL);
     g_regex_match(regex, message->plain, 0, &match_info);

```

No regex master here, https://stackoverflow.com/q/43588699 did most of the work, I only had to change the escaping.

sjaeckel:
Could you please add a comment that references the SO thread?

PR #1887, opened 2023-09-08
Fix OMEMO keyfile loading

sjaeckel:
We have to lazy load the keyfiles as they will maybe be written at one point and should be initialized by that time.

Reported by @H3rnand3zzz 

Fixup of ca2df180d81f1b30142e63cd7621a00eb9248913 resp. #1886


H3rnand3zzz:
Indeed fixes the issue.

PR #1891, opened 2023-10-04
Minor improvements

sjaeckel:
Check commit messages for details.

### DIFF
--- a/src/config/accounts.c
+++ b/src/config/accounts.c
@@ -481,7 +481,7 @@ static void
 _accounts_set_string_option(const char* account_name, const char* const option, const char* const value)
 {
     if (accounts_account_exists(account_name)) {
-        g_key_file_set_string(accounts, account_name, option, value);
+        g_key_file_set_string(accounts, account_name, option, value ?: "");
         _save_accounts();
     }
 }
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -300,7 +300,8 @@ sv_ev_room_history(ProfMessage* message)
     ProfMucWin* mucwin = wins_get_muc(message->from_jid->barejid);
     if (mucwin) {
         // if this is the first successful connection
-        if (ev_is_first_connect()) {
+        // or for some reason the `last_msg_timestamp` is not initialized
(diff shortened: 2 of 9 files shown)


src/config/accounts.c
```diff
@@ -481,7 +481,7 @@ static void
 _accounts_set_string_option(const char* account_name, const char* const option, const char* const value)
 {
     if (accounts_account_exists(account_name)) {
-        g_key_file_set_string(accounts, account_name, option, value);
+        g_key_file_set_string(accounts, account_name, option, value ?: "");
```
jubalh, line 484: Isn't here a second "value" missing? Or is this some even shorter syntax? :)

sjaeckel: That's a gcc specific extension of the ternary operator https://gcc.gnu.org/onlinedocs/gcc/Conditionals.html

src/event/server_events.c
```diff
@@ -300,7 +300,8 @@ sv_ev_room_history(ProfMessage* message)
     ProfMucWin* mucwin = wins_get_muc(message->from_jid->barejid);
     if (mucwin) {
         // if this is the first successful connection
-        if (ev_is_first_connect()) {
+        // or for some reason the `last_msg_timestamp` is not initialized
+        if (ev_is_first_connect() || !mucwin->last_msg_timestamp) {
```
jubalh, line 304: This seems not right to me.
We want to set it inside this if loop actually?

jubalh: d53c477e4 changed this btw.

sjaeckel: > [d53c477](https://github.com/profanity-im/profanity/commit/d53c477e4e2589e15267fe2b3de4bc59fe32aa3e) changed this btw.

maybe it did, but now we have cases where `g_date_time_compare()` is called with first argument set to `NULL`, which means that something is off.

> This seems not right to me.
> We want to set it inside this if statement actually?

What is not right here? We want to initialize `mucwin->last_msg_timestamp` if it is `NULL`.

jubalh: Somehow I'm too confused right now to completely understand it :) But I'll just trust you on this ;)

Issue #1899, opened 2023-10-17
Save all received messages to DB even if duplicate ID is used


jubalh:
This is to track a discussion that we had in the MUC and via PM.

Most of the things were already mentioned in
* https://github.com/profanity-im/profanity/pull/1893#issuecomment-1752434740
* https://github.com/profanity-im/profanity/pull/1893#issuecomment-1755831972

So I believe what needs to be done is:
* Save all messages even with duplicate IDs in DB (don't require <message id=""> to be unique)
* If the server generated stanza-id from XEP-0359: Unique and Stable Stanza IDs is not unique then we use log_warn and print it to console
* With things like Last Message Correction we just use the newest/latest message with that ID

It still would be valuable to find out what other clients do in this case and write it here on the issue.

Issue will most likely be worked on by @H3rnand3zzz  (only can assign you after commented on this thread).

jubalh:
Detailed report :)

It might be that one of us is misunderstanding something.

I don't see the value of `replaces_db_id`.
When we get an LMC message from the server it contains the ID that it should replace. It does *not* contain your self created `replaces_db_id`. So I don't see how you want to associate the LMC message with the correct previously received message. Basically the problem will stay the same. Because you will have several messages in the database with the same  ID. And some of them will have an internal `replaces_db_id`. But then you still don't know WHICH of those to replace because the sender couldn't inform you about that.

Even *if* that would work then you wouldn't need replaces_db_id to have an unique idientifier you could just use the DB ID field, right?

The abstract of XEP-0308 sais: `This specification defines a method for indicating that a message is a correction of the last sent message.`.

So what I believe is the correct behaviour is to replace the *newest* message with the *specified ID* as mentioned in https://github.com/profanity-im/profanity/pull/1893#issuecomment-1755831972 a week ago :)

And as mentioned above (and in the 1893 PR) it would be interesting to check what Dino, Conversations, Gajim are doing. My guess is that they will do exactly that.

H3rnand3zzz:
> I don't see the value of `replaces_db_id`. When we get an LMC message from the server it contains the ID that it should replace. It does _not_ contain your self created `replaces_db_id`. So I don't see how you want to associate the LMC message with the correct previously received message. Basically the problem will stay the same. Because you will have several messages in the database with the same ID. And some of them will have an internal `replaces_db_id`. But then you still don't know WHICH of those to replace because the sender couldn't inform you about that.

[...]

PR #1902, opened 2023-10-26
Change DB structure

H3rnand3zzz:
**Backup your DB before performing any testing.**

All tests should be done with the final version.
 - [x] Pass [all tests](https://gist.github.com/H3rnand3zzz/08c1308fa5b6e6adfdaa009092341ed1)
image
 - [x] Test migration
 - [x] Test sending + editing with `/receipts` on/off
 - [x] Test empty DB

Even though it does not drop any tables, I want to be on the cautious side and avoid any potential data loss by current or future profanity developers.

It's not ready for release yet and just shows PoC. The point of the PR is to share the idea for further discussion and development. I am not even sure if it works properly or even works in some parts. In 1 day I plan to tidy it up and make it generally better.

Read commit message for further details. Discussion links are going to be attached here and in commit message later.

# LMC is (not) fun

### Abstract

The initial problem is that when we receive messages from old client, they can have duplicate IDs and currently we don't even write them to DB, but it goes much further than that: current DB visualization (message history) relies on heavy JOIN with redacted messages. Explanation for necessity of `replaces_db_id` can be found in earlier discussions (#1893, but primarily in #1899), but to simplify, it would improve performance as well as give another benefit which is going to be discussed further. I will provide explanation for message editing (chained edits vs direct edits), short explanation for duplicated IDs problem and explained new message storage approach.

### Chained Edits vs Direct Edits
**Chained Edit** is edit that points to previous edit on second and further edits. E.g.
| ID | Message                | replaces_db_id |
|----|------------------------|----------------|
| 1  | Hlo                    | NULL           |
| 2  | Hello                  | 1              |
| 3  | Hello. How are you?    | **2**          |

We have to traverse all the edits until we can find the original message. This is not an ideal solution and LMC was not designed to be this way. Quoting the XEP,
`If the same message is to be corrected several times, the id of the original message is used in each case (e.g. If a message of id 1 is corrected by a message of id 2, a subsequent correction should correct 1, not 2).`

Though this misconception is so common that even Profanity, as well as some other popular XMPP clients (e.g. Psi) use this approach for subsequent edits.

Note: sometimes it's regarded in conversation as edit->edit->edit->original.

**Direct edit** is edit that always points to the original message
| ID | Message                | replaces_db_id |
|----|------------------------|----------------|
| 1  | Hlo                    | NULL           |
| 2  | Hello                  | 1              |
| 3  | Hello. How are you?    | **1**          |

This approach allows us to always find the original message without any need to follow chain of edits. Some clients (e.g. Gajim) follow this XEP.

But as some clients make chained edits with different stanza ids, the best way for us to handle both ways would be introducing `replaces_db_id` and using it for direct-edit approach internally, thus accepting both ways for editing messages.
| ID | Stanza | Message                | replaces_db_id | replaces_stanza_id |
|----|--------|------------------------|----------------|--------------------|
| 1  | aa     | Hlo                    | NULL           | NULL               |
| 2  | xx     | Hello                  | 1              | aa                 |
| 3  | zz     | Hello. How are you?    | **1**          | xx                 |

Here we would check xx's replace_db_id and set `replaces_db_id` as 1 (because it's the original message). Thus we don't rely on external client's approach and can always save messages in a suitable manner (performance- and design-wise).

### Duplicated ID problem
Duplicated ID is not a problem within itself. We can easily put messages with duplicated IDs in the DB as even DB field for `stanza_id` is not set to be `UNIQUE`. The problem starts with message replacement. When we get message replacement (LMC) for the message with `stanza_id` that is present 2 or more times, with old approach it wouldn't be (or barely would be) possible to replace them correctly on visualization, not to mention trying to stay performant while doing so. `replaces_db_id` could be a key to solving this problem, but it still would be just as an inefficient visualization with JOIN which can be called on every scroll up.

## New DB approach
Introduced fields:
 - `replaces_db_id` database ID for correcting message of the original message
 - `replaced_by_db_id` database ID for original message of the last correcting message

This way we connect all chained edits as direct edits, hence always have reference to the original message. `replaced_by_db_id` provides ability to replace messages quickly and correctly (JOIN only with latest change wasn't possible even with `replaces_db_id`).

<details> 
  <summary>Editable message storage approach (unacceptable, rejected by design police, left for historical ref.)</summary>
 - `replace_id` -> `replaces_stanza_id` stanza ID of the replaced message
 - `replaces_db_id` db ID of the replaced message
 - `original_message` original message before replacement
 - `message` now is used to store visual message (replaced by LMC)

So to show messages, we just need to select. But to store previous message, we can use `original_message` which stores a message which was in the original's message column at the moment of change. E.g. "Hlo" being replaced to "hello", and then to "Hello world!" we save "Hlo" to `original_message` field of the LMC and set the original message to "hello", after this on further change we will save "hello" as an `original_message` and set the original's message to "Hello world". In the DB it will look like that:
| ID | Message       | replaces_db_id | original_message |
|----|---------------|----------------|------------------|
| 1  | Hello world!  | NULL           | NULL             |
| 2  |               | 1              | hlo              |
| 3  |               | 1              | hello            |

Despite looking **unintuitive** (theoretically, `original_message` could be renamed to `previous_message`), we have performant solution that keeps full edit history, performant on visualization (just a select) and robust enough to allow direct linking despite external chain link.

<hr />
</details>

## Other changes
Edited visualization of LMC, allowing to receive messages that do not have original message counterpart in the DB (as original might be lost, e.g. received by other client during disconnect). I couldn't find good way to allow deeper (3+ edits) chained edits, so it's now similar to gajim's approach (every second edit we display a new message and then it can be redact once ad infinitum). In future we can use DB ID as a key for buffer instead of stanza ids.

### DIFF
--- a/src/database.c
+++ b/src/database.c
@@ -36,6 +36,7 @@
 #include "config.h"
 
 #include <sys/stat.h>
+#include <sys/statvfs.h>
 #include <sqlite3.h>
 #include <glib.h>
 #include <stdio.h>
@@ -58,8 +59,13 @@ static void _add_to_db(ProfMessage* message, char* type, const Jid* const from_j
 static char* _get_db_filename(ProfAccount* account);
 static prof_msg_type_t _get_message_type_type(const char* const type);
 static prof_enc_t _get_message_enc_type(const char* const encstr);
+static int _get_db_version(void);
+static gboolean _migrate_to_v2(void);
+static gboolean _check_available_space_for_db_migration(char* path_to_db);
 
 #define auto_sqlite __attribute__((__cleanup__(auto_free_sqlite)))
+
+static const int latest_version = 2;
 
 static void
 auto_free_sqlite(gchar** str)
@@ -97,32 +103,101 @@ log_database_init(ProfAccount* account)
     }
 
     char* err_msg;
-    // id is the ID of DB the entry
-    // from_jid is the senders jid
-    // to_jid is the receivers jid
-    // from_resource is the senders resource
-    // to_jid is the receivers resource
-    // message is the message text
-    // timestamp the timestamp like "2020/03/24 11:12:14"
+
+    int db_version = _get_db_version();
+    if (db_version == latest_version) {
+        return TRUE;
+    }
+
+    // ChatLogs Table
+    // Contains all chat messages
+    //
+    // id is primary key
+    // from_jid is the sender's jid
+    // to_jid is the receiver's jid
+    // from_resource is the sender's resource
+    // to_resource is the receiver's resource
+    // message is the message's text
+    // timestamp is the timestamp like "2020/03/24 11:12:14"
     // type is there to distinguish: message (chat), MUC message (muc), muc pm (mucpm)
     // stanza_id is the ID in <message>
     // archive_id is the stanza-id from from XEP-0359: Unique and Stable Stanza IDs used for XEP-0313: Message Archive Management
-    // replace_id is the ID from XEP-0308: Last Message Correction
     // encryption is to distinguish: none, omemo, otr, pgp
     // marked_read is 0/1 whether a message has been marked as read via XEP-0333: Chat Markers
-    char* query = "CREATE TABLE IF NOT EXISTS `ChatLogs` ( `id` INTEGER PRIMARY KEY AUTOINCREMENT, `from_jid` TEXT NOT NULL, `to_jid` TEXT NOT NULL, `from_resource` TEXT, `to_resource` TEXT, `message` TEXT, `timestamp` TEXT, `type` TEXT, `stanza_id` TEXT, `archive_id` TEXT, `replace_id` TEXT, `encryption` TEXT, `marked_read` INTEGER)";
+    // replace_id is the ID from XEP-0308: Last Message Correction
+    // replaces_db_id is ID (primary key) of the original message that LMC message corrects/replaces
+    // replaced_by_db_id is ID (primary key) of the last correcting (LMC) message for the original message
+    char* query = "CREATE TABLE IF NOT EXISTS `ChatLogs` ("
+                  "`id` INTEGER PRIMARY KEY AUTOINCREMENT, "
+                  "`from_jid` TEXT NOT NULL, "
+                  "`to_jid` TEXT NOT NULL, "
+                  "`from_resource` TEXT, "
+                  "`to_resource` TEXT, "
+                  "`message` TEXT, "
+                  "`timestamp` TEXT, "
+                  "`type` TEXT, "
+                  "`stanza_id` TEXT, "
+                  "`archive_id` TEXT, "
+                  "`encryption` TEXT, "
+                  "`marked_read` INTEGER, "
+                  "`replace_id` TEXT, "
+                  "`replaces_db_id` INTEGER, "
+                  "`replaced_by_db_id` INTEGER)";
     if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
         goto out;
     }
 
-    query = "CREATE TABLE IF NOT EXISTS `DbVersion` ( `dv_id` INTEGER PRIMARY KEY, `version` INTEGER UNIQUE)";
+    query = "CREATE TRIGGER IF NOT EXISTS update_corrected_message "
+            "AFTER INSERT ON ChatLogs "
+            "FOR EACH ROW "
+            "WHEN NEW.replaces_db_id IS NOT NULL "
+            "BEGIN "
+            "UPDATE ChatLogs "
+            "SET replaced_by_db_id = NEW.id "
+            "WHERE id = NEW.replaces_db_id; "
+            "END;";
+    if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
+        log_error("Unable to add `update_corrected_message` trigger.");
+        goto out;
+    }
+
+    query = "CREATE INDEX IF NOT EXISTS ChatLogs_timestamp_IDX ON `ChatLogs` (`timestamp`)";
+    if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
+        log_error("Unable to create index for timestamp.");
+        goto out;
+    }
+    query = "CREATE INDEX IF NOT EXISTS ChatLogs_to_from_jid_IDX ON `ChatLogs` (`to_jid`, `from_jid`)";
+    if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
+        log_error("Unable to create index for to_jid.");
+        goto out;
+    }
+
+    query = "CREATE TABLE IF NOT EXISTS `DbVersion` (`dv_id` INTEGER PRIMARY KEY, `version` INTEGER UNIQUE)";
     if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
         goto out;
     }
 
-    query = "INSERT OR IGNORE INTO `DbVersion` (`version`) VALUES('1')";
-    if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
+    if (db_version == -1) {
+        query = "INSERT OR IGNORE INTO `DbVersion` (`version`) VALUES ('2')";
+        if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
+            goto out;
+        }
+        db_version = _get_db_version();
+    }
+
+    // Unlikely event, but we don't want to migrate if we are just unable to determine the DB version
+    if (db_version == -1) {
+        cons_show_error("DB Initialization Error: Unable to check DB version.");
         goto out;
+    }
+
+    if (db_version < latest_version) {
+        cons_show("Migrating database schema. This operation may take a while...");
+        if (db_version < 2 && (!_check_available_space_for_db_migration(filename) || !_migrate_to_v2())) {
+            cons_show_error("Database Initialization Error: Unable to migrate database to version 2. Please, check error logs for details.");
+            goto out;
+        }
+        cons_show("Database schema migration was successful.");
     }
 
     log_debug("Initialized SQLite database: %s", filename);
@@ -130,10 +205,10 @@ log_database_init(ProfAccount* account)
 
 out:
     if (err_msg) {
-        log_error("SQLite error: %s", err_msg);
+        log_error("SQLite error in log_database_init(): %s", err_msg);
         sqlite3_free(err_msg);
     } else {
-        log_error("Unknown SQLite error");
+        log_error("Unknown SQLite error in log_database_init().");
     }
     return FALSE;
 }
@@ -202,17 +277,17 @@ ProfMessage*
 log_database_get_limits_info(const gchar* const contact_barejid, gboolean is_last)
 {
     sqlite3_stmt* stmt = NULL;
-    gchar* query;
     const char* jid = connection_get_fulljid();
     auto_jid Jid* myjid = jid_create(jid);
     if (!myjid)
         return NULL;
 
-    if (is_last) {
-        query = sqlite3_mprintf("SELECT * FROM (SELECT `archive_id`, `timestamp` from `ChatLogs` WHERE (`from_jid` = '%q' AND `to_jid` = '%q') OR (`from_jid` = '%q' AND `to_jid` = '%q') ORDER BY `timestamp` DESC LIMIT 1) ORDER BY `timestamp` ASC;", contact_barejid, myjid->barejid, myjid->barejid, contact_barejid);
-    } else {
-        query = sqlite3_mprintf("SELECT * FROM (SELECT `archive_id`, `timestamp` from `ChatLogs` WHERE (`from_jid` = '%q' AND `to_jid` = '%q') OR (`from_jid` = '%q' AND `to_jid` = '%q') ORDER BY `timestamp` ASC LIMIT 1) ORDER BY `timestamp` ASC;", contact_barejid, myjid->barejid, myjid->barejid, contact_barejid);
-    }
+    const char* order = is_last ? "DESC" : "ASC";
+    auto_sqlite char* query = sqlite3_mprintf("SELECT `archive_id`, `timestamp` FROM `ChatLogs` WHERE "
+                                              "(`from_jid` = '%q' AND `to_jid` = '%q') OR "
+                                              "(`from_jid` = '%q' AND `to_jid` = '%q') "
+                                              "ORDER BY `timestamp` %s LIMIT 1;",
+                                              contact_barejid, myjid->barejid, myjid->barejid, contact_barejid, order);
 
     if (!query) {
         log_error("Could not allocate memory for SQL query in log_database_get_limits_info()");
@@ -221,7 +296,7 @@ log_database_get_limits_info(const gchar* const contact_barejid, gboolean is_las
 
     int rc = sqlite3_prepare_v2(g_chatlog_database, query, -1, &stmt, NULL);
     if (rc != SQLITE_OK) {
-        log_error("Unknown SQLite error in log_database_get_last_info()");
+        log_error("Unknown SQLite error in log_database_get_last_info().");
         return NULL;
     }
 
@@ -235,7 +310,6 @@ log_database_get_limits_info(const gchar* const contact_barejid, gboolean is_las
         msg->timestamp = g_date_time_new_from_iso8601(date, NULL);
     }
     sqlite3_finalize(stmt);
-    sqlite3_free(query);
 
     return msg;
 }
@@ -259,9 +333,9 @@ log_database_get_previous_chat(const gchar* const contact_barejid, const char* s
     auto_gchar gchar* end_date_fmt = end_time ? end_time : g_date_time_format_iso8601(now);
     auto_sqlite gchar* query = sqlite3_mprintf("SELECT * FROM ("
                                                "SELECT COALESCE(B.`message`, A.`message`) AS message, "
-                                               "A.`timestamp`, A.`from_jid`, A.`type`, A.`encryption` FROM `ChatLogs` AS A "
-                                               "LEFT JOIN `ChatLogs` AS B ON (A.`stanza_id` = B.`replace_id` AND A.`from_jid` = B.`from_jid`) "
-                                               "WHERE A.`replace_id` = '' "
+                                               "A.`timestamp`, A.`from_jid`, A.`to_jid`, A.`type`, A.`encryption` FROM `ChatLogs` AS A "
+                                               "LEFT JOIN `ChatLogs` AS B ON (A.`replaced_by_db_id` = B.`id` AND A.`from_jid` = B.`from_jid`) "
+                                               "WHERE (A.`replaces_db_id` IS NULL OR A.`replaces_db_id` = '') "
                                                "AND ((A.`from_jid` = '%q' AND A.`to_jid` = '%q') OR (A.`from_jid` = '%q' AND A.`to_jid` = '%q')) "
                                                "AND A.`timestamp` < '%q' "
                                                "AND (%Q IS NULL OR A.`timestamp` > %Q) "
@@ -278,22 +352,23 @@ log_database_get_previous_chat(const gchar* const contact_barejid, const char* s
 
     int rc = sqlite3_prepare_v2(g_chatlog_database, query, -1, &stmt, NULL);
     if (rc != SQLITE_OK) {
-        log_error("Unknown SQLite error in log_database_get_previous_chat()");
+        log_error("SQLite error in log_database_get_previous_chat(): %s", sqlite3_errmsg(g_chatlog_database));
         return NULL;
     }
 
     GSList* history = NULL;
 
     while (sqlite3_step(stmt) == SQLITE_ROW) {
-        // TODO: also save to jid. since now part of profmessage
         char* message = (char*)sqlite3_column_text(stmt, 0);
         char* date = (char*)sqlite3_column_text(stmt, 1);
         char* from = (char*)sqlite3_column_text(stmt, 2);
-        char* type = (char*)sqlite3_column_text(stmt, 3);
-        char* encryption = (char*)sqlite3_column_text(stmt, 4);
+        char* to_jid = (char*)sqlite3_column_text(stmt, 3);
+        char* type = (char*)sqlite3_column_text(stmt, 4);
+        char* encryption = (char*)sqlite3_column_text(stmt, 5);
 
         ProfMessage* msg = message_init();
         msg->from_jid = jid_create(from);
+        msg->to_jid = jid_create(to_jid);
         msg->plain = strdup(message);
         msg->timestamp = g_date_time_new_from_iso8601(date, NULL);
         msg->type = _get_message_type_type(type);
@@ -375,6 +450,7 @@ static void
 _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Jid* const to_jid)
 {
     auto_gchar gchar* pref_dblog = prefs_get_string(PREF_DBLOG);
+    sqlite_int64 original_message_id = -1;
 
     if (g_strcmp0(pref_dblog, "off") == 0) {
         return;
@@ -407,10 +483,10 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
         type = (char*)_get_message_type_str(message->type);
     }
 
-    // Check LMC validity (XEP-0308)
+    // Apply LMC and check its validity (XEP-0308)
     if (message->replace_id) {
-        auto_sqlite char* replace_check_query = sqlite3_mprintf("SELECT `from_jid` FROM `ChatLogs` WHERE `stanza_id` = '%q'",
-                                                                message->replace_id ? message->replace_id : "");
+        auto_sqlite char* replace_check_query = sqlite3_mprintf("SELECT `id`, `from_jid`, `replaces_db_id` FROM `ChatLogs` WHERE `stanza_id` = %Q ORDER BY `timestamp` DESC LIMIT 1",
+                                                                message->replace_id);
 
         if (!replace_check_query) {
             log_error("Could not allocate memory for SQL replace query in log_database_add()");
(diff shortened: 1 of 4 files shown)


src/database.c
```diff
@@ -97,12 +101,16 @@ log_database_init(ProfAccount* account)
     }
 
     char* err_msg;
+
+    // ChatLogs Table
+    // Contains visual representation of chat logs
```
jubalh, line 106: Not sure what `visual representation of chat logs` should exactly mean..

src/database.c
```diff
@@ -115,16 +123,39 @@ log_database_init(ProfAccount* account)
         goto out;
     }
 
-    query = "CREATE TABLE IF NOT EXISTS `DbVersion` ( `dv_id` INTEGER PRIMARY KEY, `version` INTEGER UNIQUE)";
+    gboolean new_installation = (_get_db_version() == -1);
```
jubalh, line 126: It doesn't need to be a new installation. It can also be that it's just a a new user who gets a new database. So I would call it `first_run`, `first_start` or `db_not_existant` or.. I'm not sure :)

src/database.c
H3rnand3zzz: By "visual representation" I mean the chat messages you see when you open a chat window (i.e. corrected version)

src/database.c
H3rnand3zzz: Adressed

src/database.c
```diff
@@ -115,16 +123,39 @@ log_database_init(ProfAccount* account)
...
+
+    int db_version = _get_db_version();
+    if (db_version == -1) {
+        cons_show_error("DB Initialization Error: Unable to check DB version.");
+        goto out;
+    }
+
+    if (db_version < 2) {
+        if (database_exists) {
+            cons_show("Updating DB to V2. Please, wait...");
```
jubalh, line 148: User shouldn't be bothered with too much information and text. This should go as seemless as possible. I would maybe write `Updating Database schema to new version` or something like that. So that we can leave this text also vor a V3 and so on.

jubalh: Generally lets write `database` and not `DB` to the user. Also in the messages below.

src/database.c
```diff
@@ -455,12 +515,12 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
     }
 
     if (duplicate_exists) {
-        log_warning("Duplicate stanza-id found for the message. stanza_id: %s; archive_id: %s; sender: %s; content: %s", message->id, message->stanzaid, from_jid->barejid, message->plain);
-        return;
+        log_error("Duplicate server stanza-id found for the message. stanza_id: %s; archive_id: %s; sender: %s; content: %s", message->id, message->stanzaid, from_jid->barejid, message->plain);
```
jubalh, line 518: `server stanza-id` sounds confusing. Just leave stanza-id since its always set by server.

src/database.c
```diff
@@ -493,3 +555,159 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+        if (sqlite3_step(statement) == SQLITE_ROW) {
+            current_version = sqlite3_column_int(statement, 0);
+        }
+        sqlite3_finalize(statement);
+    }
+    return current_version;
+}
+
+/**
+ * @brief Migration to V2 that introduces new fields and reorganizes message storage.
```
jubalh, line 576: Remove the @brief and leave the rest. Its a nice description.

src/database.c
```diff
@@ -493,3 +555,159 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+ * `replace_id` -> `replaces_stanza_id` stanza ID of the replaced message
+ * `replaces_db_id` db ID of the replaced message
+ * `original_message` original message before replacement
+ * `message` now is used to store visual message (replaced by LMC)
+ *
+ * @return gboolean with migration success status.
```
jubalh, line 582: Remove `@return gboolean with migration success status.` It doesnt tell us anything. It doesnt tell whether true is success or false so we know exactly the same like just reading the function header.
And as previously said in reviews we dont want doxygen style docu.

src/database.c
```diff
@@ -493,3 +555,159 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+ */
+static gboolean
+_migrate_to_v2(void)
+{
+    char* err_msg;
+
+    // Make V2 table
+    char* query = "CREATE TABLE `ChatLogsV2` (`id` INTEGER PRIMARY KEY AUTOINCREMENT, `from_jid` TEXT NOT NULL, `to_jid` TEXT NOT NULL, `from_resource` TEXT, `to_resource` TEXT, `message` TEXT, `timestamp` TEXT, `type` TEXT, `stanza_id` TEXT, `archive_id` TEXT, `encryption` TEXT, `marked_read` INTEGER, `original_message` TEXT, `replaces_db_id` INTEGER, `replaces_stanza_id` TEXT)";
```
jubalh, line 590: It does not make sense to call it ChatLogsV2. We want the version only in the db_version. Not in the name of the table. The table should be called ChatLogs.

src/database.c
```diff
@@ -493,3 +555,159 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+    }
+
+    // Set V2 table as main
+    query = "ALTER TABLE ChatLogsV2 RENAME TO ChatLogs;";
+    if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
+        log_error("[DB Migration] Unable to rename new table.");
+        goto cleanup;
+    }
+
+    // TODO: drop old table (_bck). Only after sufficient testing this part should be added
```
jubalh, line 659: Before merging this should be activated. We dont want the db file to become too big.

src/database.c
```diff
@@ -493,3 +555,159 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+            log_error("[DB Migration] Unable to delete old DB backup.");
+            return FALSE;
+        }
+
+    */
+
+    // Set new DB version (V2)
+    query = "UPDATE `DbVersion` SET `version` = 2";
+    if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
+        log_error("[DB Migration] Unable to update DB Version.");
```
jubalh, line 672: In other places you write `log_error("SQLite error: %s", err_msg);` here you write `log_error("[DB Migration] Unable to update DB Version.");` this should be unified. Just use SQLite error again. It is clear that it is during migration because it is about database version.

src/database.c
```diff
@@ -493,3 +555,159 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+        log_error("SQLite error: %s", err_msg);
+        sqlite3_free(err_msg);
+        err_msg = NULL;
+    } else {
+        log_error("Unknown SQLite error.");
+    }
+
+    query = "DROP TABLE `ChatLogsV2`;";
+    if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
+        log_error("[DB Migration] Unable to drop V2 table: %s", err_msg);
```
jubalh, line 689: same like above

src/database.c
```diff
@@ -493,3 +555,159 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+        sqlite3_free(err_msg); // TODO: test this part, looks suspicious
```
jubalh, line 690: test should be done or todo removed before merging

jubalh (changes requested):
I added some comments for a first superficial review. Didn't check new DB layout yet.
Generally please write `database` to the user and not DB. I also don't like the `V2` the user shouldnt be faced with such details. And generally I like `version 2` even in in code comments.

src/database.c
H3rnand3zzz: > It does not make sense to call it ChatLogsV2. We want the version only in the db_version. Not in the name of the table. The table should be called ChatLogs in the end. During transaction could have the V2 there or something like "transformation" or something. But in the end it should be ChatLogs again.

Well, it's renamed later to "ChatLogs" (once all the fields are transformed), so I guess it's resolved.

H3rnand3zzz:
Rebased on master (since last update was history-related), in the next update I am going to address the issues

src/database.c
H3rnand3zzz: Line removed

src/database.c
H3rnand3zzz: > User shouldn't be bothered with too much information and text. This should go as seemless as possible. I would maybe write `Updating Database schema to new version` or something like that. So that we can leave this text also vor a V3 and so on.

version 2 is good since we know which exact version update was problematic (e.g. it can stuck).

jubalh:
Now comes to mind: I think it is `migrate` instead of `update` when you change the database layout. 
Also it would be good to have the migration in a function.

So lets say later we have version 3. But a user is using a Linux that has a long update cycle, like Debian. So it might be that he is using profanity X with database version 1 and then after a dist upgrade gets profanity Z with database version 4.
So upon start we should check the db version. Then apply the migrations for version 2 and then apply the migrations to version 3 and then to 4.

Probably each of this should be a separate function that we call upon need?

src/database.c
H3rnand3zzz: [DB Migration] messages show us place in the code, SQLite error shows us what was wrong with SQLite query.
So it should look something like that:
```log
[DB Migration] Unable to update DB Version.
SQLite error: unknown field `version`.
```

I can probably remove [DB Migration] prefix (though the prefix would make searching for errors in logs much easier), but I don't want to leave just 1 error.

H3rnand3zzz:
> Now comes to mind: I think it is `migrate` instead of `update` when you change the database layout. Also it would be good to have the migration in a function.
> 
> So lets say later we have version 3. But a user is using a Linux that has a long update cycle, like Debian. So it might be that he is using profanity X with database version 1 and then after a dist upgrade gets profanity Z with database version 4. So upon start we should check the db version. Then apply the migrations for version 2 and then apply the migrations to version 3 and then to 4.
> 
> Probably each of this should be a separate function that we call upon need?

That's why I did < 2 version check, and called it `_migrate_to_v2`, we can just add `_migrate_to_v3` without touching the original migration function. We can probably do something like that:
```c
   int LAST_VERSION = 5;

    if (db_version < LAST_VERSION) {
        if (database_exists) {
            cons_show("Updating database schema. This operation may take a while...");
        }
        if(!_check_available_space_for_db_migration(filename)) {
          cons_show_error("Non enough space for database update...");
        }
        
        for(i = 2; i < LAST_VERSION+1; i++){
            if (version < i && !_migrate_to_v(i)) { // or some other way to concatenate function with version
                cons_show_error("Database Initialization Error: Unable to update database to version %d. Please, check error logs for details.", ver);
                goto out;
            }
        }
        if (database_exists) {
            cons_show("Database schema update was successful.");
        }
    }
```

src/database.c
jubalh: I think the user just needs to know that the DB is migrated. In the log file he can find the information from which version to which it is happening in case of an error. So I would probably write Migrate instead of Update here too.

H3rnand3zzz:
```c
            cons_show("Migrating database schema. This operation may take a while...");
```
But I want to keep this message to avoid user's panic if program is stuck for (e.g.) 5-50 seconds.

H3rnand3zzz:
All suggested changes are applied.

jubalh:
I pulled the PR and ran it locally.
I messaged `one` and did 2 corrections. I messaged `a` and did one correction.
Nothing is shown as a correction in the UI when _sending_:

```
18:03:54 - me: one
18:04:00 - me: two
18:04:07 - me: three
18:04:12 - me: a
18:04:17 - me: b
```

The receiving end (old profanity version) still displays it correctly.

jubalh (changes requested):
Seems like UI doesnt display LMC messages from sender correctly anymore.

H3rnand3zzz:
> Seems like UI doesnt display LMC messages from sender correctly anymore.

Still unable to reproduce. Please, respond in the chat so we can figure the problem out.

jubalh:
* Run profanity from master branch
* Copy the db to a backup
* Switch to your branch with commit `87c0c8ca0dfcdf19667015c7c6c0fb35eef5b1f2`
* Compile and run
* `/msg mytestacc@domain.org` (this test account runs Profanity on master)
* `hi`
* `/correct Hi`
* `/correct strange`
* `Another`
* `/correct Another!`

UI of sending account with 87c0c8ca0dfcdf19667015c7c6c0fb35eef5b1f2:

scr1

DB:

scr2

UI of receiving account on master:

scr3

H3rnand3zzz:
@jubalh 
I repeated this step-by-step.
image
image

jubalh:
>  Fix has been applied.

Much better now it works :)

src/database.c
```diff
@@ -419,25 +449,34 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+        if (SQLITE_OK != sqlite3_prepare_v2(g_chatlog_database, replace_check_query, -1, &lmc_stmt, NULL)) {
+            log_error("SQLite error in _add_to_db() on selecting original message: %s", sqlite3_errmsg(g_chatlog_database));
+            return;
+        }
+
+        if (sqlite3_step(lmc_stmt) == SQLITE_ROW) {
+            original_message_id = sqlite3_column_int(lmc_stmt, 0);
+            const char* from_jid_orig = (const char*)sqlite3_column_text(lmc_stmt, 1);
+
+            // Handle non-XEP-compliant replacement messages (edit->edit->original)
```
jubalh, line 461: Didn't check this part in detail. I remember @sjaeckel mentioned something about not fixing others bugs in the MUC.

src/ui/window.c
```diff
@@ -1335,19 +1335,23 @@ win_show_status_string(ProfWin* window, const char* const from,
     win_appendln(window, presence_colour, "");
 }
 
-static void
+/** Corrects the visual representation of a message with prior check for sender validity.
+ *
+ * @returns TRUE if the message was successfully corrected and should not be printed, FALSE otherwise
```
jubalh, line 1340: Still this doxygen `@returns`. But not important I can change that after merging.

src/database.c
```diff
@@ -491,5 +531,175 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+        log_error("[DB Migration] Unable to create index for to_jid.");
+        goto cleanup;
+    }
+    query = "CREATE INDEX ChatLogs_from_jid_IDX ON `ChatLogsV2` (`from_jid`)";
+    if (SQLITE_OK != sqlite3_exec(g_chatlog_database, query, NULL, 0, &err_msg)) {
+        log_error("[DB Migration] Unable to create index for from_jid.");
+        goto cleanup;
+    }
+
+    // Select all messages form V1 to V2
```
jubalh, line 613: small typo `form`

jubalh (approved):
I added some comments but for me it is good enough for master now.

src/database.c
H3rnand3zzz: > Didn't check this part in detail. I remember @sjaeckel mentioned something about not fixing others bugs in the MUC.

It's also our bug as well. At the very least we should be compatible with older profanity versions.

src/ui/window.c
H3rnand3zzz: > Still this doxygen `@returns`. But not important I can change that after merging.

it's doxygen style, but it doesn't take excessive space, but at the same time, provide better readability

src/database.c
H3rnand3zzz: fixed

H3rnand3zzz:
Fixed typo, improved logging.

src/database.c
```diff
@@ -469,6 +508,7 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
                                               date_fmt ? date_fmt : "",
                                               message->id ? message->id : "",
                                               message->stanzaid ? message->stanzaid : "",
+                                              original_message_id == -1 ? "" : sqlite3_mprintf("%d", original_message_id),
```
sjaeckel, line 511: Quoting the sqlite documentation:

> The sqlite3_mprintf() and sqlite3_vmprintf() routines write their
** results into memory obtained from [sqlite3_malloc64()].
** The strings returned by these two routines should be
** released by [sqlite3_free()]

src/database.c
```diff
@@ -481,15 +521,185 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
         }
+
+        sqlite3_int64 last_inserted_id = sqlite3_last_insert_rowid(g_chatlog_database);
+        if (last_inserted_id < 1) {
+            log_error("Message was inserted without error, but unable to get last inserted ID (got: %d)", last_inserted_id);
+            return;
+        }
+
+        if (message->replace_id) {
+            // TODO: maybe put as a trigger
```
sjaeckel, line 542: Please do so.

I'd prefer having a final solution merged instead of re-iterating again and again since introducing this would bump the db version again, or not?

src/database.c
```diff
@@ -481,15 +521,185 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+ * `replaces_db_id` database ID for correcting message of the original message
+ * `replaced_by_db_id` database ID for original message of the last correcting message
+ */
+static gboolean
+_migrate_to_v2(void)
+{
+    char* err_msg;
+
+    // Make ChatLogsV2 table which will be renamed to ChatLogs on transformation success
+    char* query = "CREATE TABLE `ChatLogsV2` (`id` INTEGER PRIMARY KEY AUTOINCREMENT, `from_jid` TEXT NOT NULL, `to_jid` TEXT NOT NULL, `from_resource` TEXT, `to_resource` TEXT, `message` TEXT, `timestamp` TEXT, `type` TEXT, `stanza_id` TEXT, `archive_id` TEXT, `encryption` TEXT, `marked_read` INTEGER, `replaces_db_id` INTEGER, `replace_id` TEXT, `replaced_by_db_id` INTEGER)";
```
sjaeckel, line 588: I'd prefer if the name of the 'migration table' would be more clear. This looks like it'd be a legit table name ... maybe use `ChatLogs_migration_v1_to_v2`!?

src/database.c
```diff
@@ -481,15 +521,185 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
+cleanup:
+    if (err_msg) {
+        log_error("SQLite error in _migrate_to_v2(): %s", err_msg);
+        sqlite3_free(err_msg);
+        err_msg = NULL;
+    } else {
+        log_error("Unknown SQLite error in _migrate_to_v2().");
+    }
+
+    query = "DROP TABLE `ChatLogsV2`;";
```
sjaeckel, line 679: Would it maybe make sense to turn around the logic and keep this table on failure, but drop it if it exists before starting the migration?

This would allow investigating the migration table after a failure.

src/ui/window.c
```diff
@@ -1464,9 +1460,7 @@ win_print_outgoing(ProfWin* window, const char* show_char, const char* const id,
     GDateTime* timestamp = g_date_time_new_now_local();
 
     const char* myjid = connection_get_fulljid();
-    if (replace_id) {
-        _win_correct(window, message, id, replace_id, myjid);
-    } else {
+    if (!replace_id || !_win_correct(window, message, id, replace_id, myjid)) {
```
sjaeckel, line 1463: Why isn't `PREF_CORRECTION_ALLOW` checked here?

src/ui/window.c
```diff
@@ -1625,10 +1619,15 @@ win_print_outgoing_with_receipt(ProfWin* window, const char* show_char, const ch
     receipt->received = FALSE;
 
     const char* myjid = connection_get_fulljid();
+    gboolean corrected = FALSE;
     if (replace_id) {
-        _win_correct(window, message, id, replace_id, myjid);
-        free(receipt); // TODO: probably we should use this in _win_correct()
-    } else {
+        corrected = _win_correct(window, message, id, replace_id, myjid);
```
sjaeckel, line 1624: Why isn't `PREF_CORRECTION_ALLOW` checked here?

src/ui/window.c
```diff
@@ -1406,9 +1408,7 @@ win_print_incoming(ProfWin* window, const char* const display_name_from, ProfMes
             enc_char = strdup("-");
         }
 
-        if (prefs_get_boolean(PREF_CORRECTION_ALLOW) && message->replace_id) {
-            _win_correct(window, message->plain, message->id, message->replace_id, message->from_jid->barejid);
-        } else {
+        if (!prefs_get_boolean(PREF_CORRECTION_ALLOW) || !message->replace_id || !_win_correct(window, message->plain, message->id, message->replace_id, message->from_jid->barejid)) {
```
sjaeckel, line 1411: How about moving those first two checks inside `_win_correct()` and simply always call it?

src/database.c
```diff
@@ -419,25 +449,34 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
         }
+        sqlite3_finalize(lmc_stmt);
     }
 
     // Check for duplicate messages
-    auto_sqlite char* duplicate_check_query = sqlite3_mprintf("SELECT 1 FROM `ChatLogs` WHERE (`archive_id` = '%q' AND `archive_id` != '') OR (`stanza_id` = '%q' AND `stanza_id` != '')",
-                                                              message->stanzaid ? message->stanzaid : "",
-                                                              message->id ? message->id : "");
+    auto_sqlite char* duplicate_check_query = sqlite3_mprintf("SELECT 1 FROM `ChatLogs` WHERE (`archive_id` = '%q' AND `archive_id` != '')",
+                                                              message->stanzaid ? message->stanzaid : "");
```
sjaeckel, line 479: Shouldn't we maybe handle the case where `message->stanzaid == ("" or NULL)` before getting here? FMU this should never happen!

Then the `AND 'archive_id' != ''` part could be dropped.

src/database.c
```diff
@@ -97,13 +101,17 @@ log_database_init(ProfAccount* account)
...
     char* err_msg;
-    // id is the ID of DB the entry
-    // from_jid is the senders jid
-    // to_jid is the receivers jid
-    // from_resource is the senders resource
-    // to_jid is the receivers resource
-    // message is the message text
-    // timestamp the timestamp like "2020/03/24 11:12:14"
+
+    // ChatLogs Table (Version 1)
```
sjaeckel, line 105: Why would we always want to start off with `version 1`? Wouldn't it make more sense to check the version first, if we're at `none` or `<latest>` we're fine and directly create `version <latest>` style tables.

If we're at `version <not latest>` we start the migration.

sjaeckel (changes requested):
Thanks for doing the work! Basically looks OK to me, but IMO needs some further refinement before it's ready to be merged.

src/database.c
H3rnand3zzz: > Would it maybe make sense to turn around the logic and keep this table on failure, but drop it if it exists before starting the migration?
> 
> This would allow investigating the migration table after a failure.

In case of an error, we are going to keep the original table + migration error in the logs. We can just update the app based on this data and all that user will have to do is update app, after this original table would migrate without mistakes without need for user to interact with DB.

src/ui/window.c
H3rnand3zzz: It's checked priorly, in the cmd_correct. That's a prior mess, I didn't want to change the logic and set checking everywhere in cmd_function/replace function, as it's a design solution and it would require further negotiations etc.

src/ui/window.c
H3rnand3zzz: Same as above.

src/ui/window.c
H3rnand3zzz: That would make sense.

src/database.c
```diff
@@ -115,25 +123,46 @@ log_database_init(ProfAccount* account)
...
+
+    if (db_version < 2) {
+        if (database_exists) {
+            cons_show("Migrating database schema. This operation may take a while...");
+        }
+        if (!_check_available_space_for_db_migration(filename) || !_migrate_to_v2()) {
+            cons_show_error("Database Initialization Error: Unable to migrate database to version 2. Please, check error logs for details.");
+            goto out;
+        }
+        if (database_exists) {
```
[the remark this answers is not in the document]
H3rnand3zzz: That's intended.

src/database.c
```diff
@@ -455,12 +494,12 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
     }
 
     if (duplicate_exists) {
-        log_warning("Duplicate stanza-id found for the message. stanza_id: %s; archive_id: %s; sender: %s; content: %s", message->id, message->stanzaid, from_jid->barejid, message->plain);
-        return;
+        log_error("Duplicate stanza-id found for the message. stanza_id: %s; archive_id: %s; sender: %s; content: %s", message->id, message->stanzaid, from_jid->barejid, message->plain);
+        cons_show_error("Got a message with duplicate server ID from %s.", from_jid->fulljid);
```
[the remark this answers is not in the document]
H3rnand3zzz: It's [XEP-0359](https://xmpp.org/extensions/xep-0359.html)'s stanza-id. @jubalh, which name should be used here?

src/database.c
H3rnand3zzz: Done.

src/database.c
```diff
@@ -419,25 +449,34 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
...
-        if (SQLITE_OK == sqlite3_prepare_v2(g_chatlog_database, replace_check_query, -1, &lmc_stmt, NULL)) {
-            if (sqlite3_step(lmc_stmt) == SQLITE_ROW) {
-                const char* from_jid_orig = (const char*)sqlite3_column_text(lmc_stmt, 0);
+        if (SQLITE_OK != sqlite3_prepare_v2(g_chatlog_database, replace_check_query, -1, &lmc_stmt, NULL)) {
+            log_error("SQLite error in _add_to_db() on selecting original message: %s", sqlite3_errmsg(g_chatlog_database));
+            return;
+        }
+
+        if (sqlite3_step(lmc_stmt) == SQLITE_ROW) {
+            original_message_id = sqlite3_column_int(lmc_stmt, 0);
```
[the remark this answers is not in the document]
H3rnand3zzz: Done.

src/database.c
H3rnand3zzz: Thanks, done.

src/database.c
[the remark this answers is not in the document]
jubalh: I agree that `stanza-id` is the correct term used in the XEP and probably users will not be aware of our internal DB column name. So I think `stanza-id` is good.

src/database.c
H3rnand3zzz: That's a common practice with migrations, since not only the DB structure changes, but also triggers, indexes are added etc. Totally it would be a big DRY violation and it would just make all new migrations harder, since changes would need to be duplicated.

src/database.c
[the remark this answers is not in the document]
H3rnand3zzz: Made it persistent. Stanza-id with explanation that it's a server-generated, so no ambiguity here.

src/ui/window.c
H3rnand3zzz: Moved the check, but without prefs_get_boolean, as it would duplicate checking and be inefficient.

src/ui/window.c
H3rnand3zzz: I will mark it as "resolved" and once the decision is made on another similar comment, I will apply the change at all places.

src/ui/window.c
H3rnand3zzz: @jubalh, where the check for pref should be done? In cmd_correct it shows a good error, thus being more user-friendly. As for me, I am not sure that we should make this check in the internal function, since it's needed for some other functions (e.g. download progress)

src/ui/window.c
H3rnand3zzz: Same as above, let's continue discussion here https://github.com/profanity-im/profanity/pull/1902#discussion_r1381536586

H3rnand3zzz:
Thank you for the review.

src/ui/window.c
sjaeckel: IMO that check doesn't necessarily only have to be once. If it's better to be displayed in the UI, please do it once where the error reporting is good, but also do it later in the code (maybe a second time) if it improves the code overall.

src/database.c
H3rnand3zzz: Now it's `TRANSACTION`, so it's resolved, I guess.

src/database.c
H3rnand3zzz: Done.

src/ui/window.c
jubalh: Would be good to have it solved correctly but I agree that its not the point of your PR. It would be great to have it in a 2nd commit though. I'll have to leave now and will merge in the evening if all things are addressed and @sjaeckel gave approval as well. Thanks for the work everybody!

H3rnand3zzz:
Last version has been successfully retested (tests list is in the PR message).

sjaeckel (approved):
Good work! Thanks

jubalh:
Thanks everybody for the detailed work and reviewing and patience.

PR #1906, opened 2023-10-30
Fix loading of encryption type

sjaeckel:
The column of the encryption type in the result is 4, not 3.

Fixup of 6375b2719fd16cbedf531ff197f6d82ffbcc873a


jubalh (approved):

jubalh:
Thanks!

PR #1910, opened 2023-11-03
Show encryption for historical messages

H3rnand3zzz:
See commit message.

- [x] No need to run valgrind when using my new feature

* Use page up in the chat window until you see historical message. Encryption used should be displayed accordingly.


jubalh (approved):
Would have been cool to have this in 2 commits. One that adds the new function in the existing places. And then another one that makes encryption char work for history (and also uses that new function).
But its okay.

PR #1912, opened 2023-11-06
Add new mode for statusbar tab behaviour

### LINKED ISSUE #1283: Always show focused window number in statusbar
Profanity version: 3c56b289
libstrophe: strophe/libstrophe@c92fc84

## Expected Behavior
Show the number of the focused window always, even if > 10.

## Current Behavior
Only numbers from 1 to 10 (0) are displayed in the statusbar.

## Possible Solution
Either display the number always alongside other numbers, or alongside the room name maybe (sounds even better).

### LINKED ISSUE #1764: Statusbar - not displaying as configured
# Statusbar not displaying as configued

This issue is not a software bug, but it is a UX design flaw.

`/statusbar` is configued to hide all read chat and display the unread chat:

```
[time] - Show tab names (/statusbar)             : ON
[time] - Show tab numbers (/statusbar)           : ON
[time] - Show tab with no actions (/statusbar)   : ON
[time] - Max tabs (/statusbar)                   : 10
[time] - Max tab length (/statusbar)             : OFF
[time] - Self statusbar display (/statusbar)     : jid
[time] - Chat tab display (/statusbar)           : user
[time] - Room tab display (/statusbar)           : room
```

With the above configuration, `/statusbar` display the number and name of the current window on the right side and only display additional items when new message are recieved.

## Context

`/statusbar` is working just fine when user opens less than 10 chat windows which is the max limit of `/statusbar maxtabs`. Because there are only 10 numbers printed on the physical keyboard. When user have more than the limit, by default, it shows `[>]` to signal there are more windows than can be shown.

The problem comes when the unread messages are assigned in windows that are above the `/statusbar maxtabs`.

At the moment I look there and either feel confused or think that I have no new message to read.

It felt like I broke my back bone in 5 places. (no it doesn't.)

## Current Behavior

It is showing `[>]`. 

This is annoying because this is the corner of the screen that I look to check and decide before giving the command `/win` + `[number|name]` to open the very window I want read.

At the moment, I have to go to the console window and run `/wins` before I can decide which window I want to open. 
In terms of speed, this costs at least 12-key-strokes.
In terms of human relationship, this might costs a heart break if the user missed the new message.

## Expected Behavior

As a user I expect it to show the maximum of 10 windows. This includes my current window AND as many unread windows as possible BEFORE signal to me that there are even more new messages to read.

1. Open less than the limit.
2. Recieve a new message.
3. Open more than the limit.
4. Move the new message over the limit with `/wins swap`.
5. Feel confused on why it vanished.

## Environment
profanity 0.13.0.7dev.local.ffe55e9

doesn't matter.

* glib version
doesn't matter.

## Possible Solution

Code in English:
```
window_max = 10;
window_displayed = 0;
window_unread = ??;
more = "[>]";

display_statusbar() {
    // Some-magical-cursor-movements here.

    // Display the n° || name of current window.
    print_window_current();
    window_displayed = 1

    // Display all the unread.
    if (window_unread > 0) {
        while (window_unread > window_displayed) {
            while (window_displayed < window_max) {
                print_window_newest();

                // some-magical-pointers passing here.

                window_displayed++;
            }  

            // Display signal if there are still more.
            if (window_max < window_displayed + window_unread && window_unread > 0) 
                print("%s", more);
            window_unread--;
        }
    }
}
```

H3rnand3zzz:
Display current pages dynamically so users can navigate easily in-depth.

- [x] Indicate unread message to the left (**[<]**) and to the right (**[>]**) of the range.

* Use the following commands (for easier testing, you can add them to `scripts/msg` folder and `/account set YOUR_ACCOUNT startscript msg` (BEWARE: can cause occasional strophe-related crash, use 10-15 windows for safety and make `/statusbar maxtabs 5` to see the effect better)
```
/msg 1
/msg 3
/msg 4
/msg 5
/msg 6
/msg 7
/msg 8
/msg 9
/msg 10
/msg 11
/msg 12
/msg 13
/msg 14
/msg 15
/msg 16
/msg 17
/msg 18
/msg 19
```
* navigate through pages easily
* use `/statusbar hide name`, `/statusbar show name`, `/statusbar hide number` to test all the possible mods.

Potentially fixes #1764.
Fixes #1283

### Win 6 (or 7, I don't know and remember and have no way to know except manually counting)
I am on tab 6 (or 7, I don't know and remember and have no way to know except manually counting)

### 2 tabs more left to the previous position
I am 2 tabs more left to the previous position

### Win 3
I am on tab 3

### Win 6
image

### Win 8 with new message in console
image

### Win 2 with new message in console
image

### Win 4 with new message in console and from a "friend"
image

### DIFF
--- a/src/command/cmd_defs.c
+++ b/src/command/cmd_defs.c
@@ -1289,7 +1289,7 @@ static const struct cmd_t command_defs[] = {
               "/statusbar hide name|number|read",
               "/statusbar maxtabs <value>",
               "/statusbar tablen <value>",
-              "/statusbar tabmode default|actlist",
+              "/statusbar tabmode default|dynamic|actlist",
               "/statusbar self user|barejid|fulljid|off",
               "/statusbar chat user|jid",
               "/statusbar room title bookmark|jid|localpart|name",
@@ -1300,7 +1300,7 @@ static const struct cmd_t command_defs[] = {
       CMD_ARGS(
               { "maxtabs <value>", "Set the maximum number of tabs to display, <value> must be between 0 and 10." },
               { "tablen <value>", "Set the maximum number of characters to show as the tab name, 0 sets to unlimited." },
-              { "tabmode default|actlist", "Set the mode how the 'active tabs' are shown." },
+              { "tabmode default|dynamic|actlist", "Set the mode tabs are shown. `dynamic` is a mode that displays tabs conveniently around current tab, thus providing proper pagination. `actlist` setting shows only active tabs. `default` setting always shows tabs in 1 to max_tabs range." },
               { "show|hide name", "Show or hide names in tabs." },
               { "show|hide number", "Show or hide numbers in tabs." },
               { "show|hide read", "Show or hide inactive tabs." },
--- a/src/ui/statusbar.c
+++ b/src/ui/statusbar.c
@@ -78,15 +78,18 @@ static GTimeZone* tz;
 static StatusBar* statusbar;
 static WINDOW* statusbar_win;
 
+void _get_range_bounds(int* start, int* end);
 static int _status_bar_draw_time(int pos);
 static int _status_bar_draw_maintext(int pos);
 static int _status_bar_draw_bracket(gboolean current, int pos, const char* ch);
-static int _status_bar_draw_extended_tabs(int pos);
+static int _status_bar_draw_extended_tabs(int pos, gboolean prefix, int start, int end);
 static int _status_bar_draw_tab(StatusBarTab* tab, int pos, int num);
+static int _status_bar_draw_tabs(int pos);
 static void _destroy_tab(StatusBarTab* tab);
-static int _tabs_width(void);
+static int _tabs_width(int start, int end);
+static unsigned int _count_digits(int number);
+static unsigned int _count_digits_in_range(int start, int end);
 static char* _display_name(StatusBarTab* tab);
-static gboolean _extended_new(void);
 static gboolean _tabmode_is_actlist(void);
 
 void
@@ -282,26 +285,40 @@ status_bar_draw(void)
     werase(statusbar_win);
     wbkgd(statusbar_win, theme_attrs(THEME_STATUS_TEXT));
 
+    gint max_tabs = prefs_get_statusbartabs();
     int pos = 1;
 
     pos = _status_bar_draw_time(pos);
-
     pos = _status_bar_draw_maintext(pos);
+    if (max_tabs != 0)
+        pos = _status_bar_draw_tabs(pos);
 
+    wnoutrefresh(statusbar_win);
+    inp_put_back();
+}
+
+static int
+_status_bar_draw_tabs(int pos)
+{
     if (!_tabmode_is_actlist()) {
-        pos = getmaxx(stdscr) - _tabs_width();
+        int start, end;
+        _get_range_bounds(&start, &end);
+
+        pos = getmaxx(stdscr) - _tabs_width(start, end);
         if (pos < 0) {
             pos = 0;
         }
-        gint max_tabs = prefs_get_statusbartabs();
-        for (int i = 1; i <= max_tabs; i++) {
+
+        pos = _status_bar_draw_extended_tabs(pos, TRUE, start, end);
+
+        for (int i = start; i <= end; i++) {
             StatusBarTab* tab = g_hash_table_lookup(statusbar->tabs, GINT_TO_POINTER(i));
             if (tab) {
                 pos = _status_bar_draw_tab(tab, pos, i);
             }
         }
 
-        _status_bar_draw_extended_tabs(pos);
+        pos = _status_bar_draw_extended_tabs(pos, FALSE, start, end);
     } else {
         pos++;
         guint print_act = 0;
@@ -339,13 +356,13 @@ status_bar_draw(void)
             pos = _status_bar_draw_bracket(FALSE, pos, "]");
         }
     }
-
-    wnoutrefresh(statusbar_win);
-    inp_put_back();
+    return pos;
 }
 
+// Checks if there are highlighted (unread) messages on the (left side if `left_side` is true, else right side)
+// within the current displayed tabs range.
 static gboolean
-_extended_new(void)
+_has_new_msgs_beyond_range_on_side(gboolean left_side, int display_tabs_start, int display_tabs_end)
 {
     gint max_tabs = prefs_get_statusbartabs();
     int tabs_count = g_hash_table_size(statusbar->tabs);
@@ -353,7 +370,10 @@ _extended_new(void)
         return FALSE;
     }
 
-    for (int i = max_tabs + 1; i <= tabs_count; i++) {
+    int start = left_side ? 1 : display_tabs_end + 1;
+    int end = left_side ? display_tabs_start - 1 : tabs_count;
+
+    for (int i = start; i <= end; i++) {
         StatusBarTab* tab = g_hash_table_lookup(statusbar->tabs, GINT_TO_POINTER(i));
         if (tab && tab->highlight) {
             return TRUE;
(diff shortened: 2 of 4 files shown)


H3rnand3zzz:
I will provide comments here for each rebase to keep the context. Previous rebases were with master, just to keep the code up-to-date.

H3rnand3zzz:
## Revision 1
- Add function to count digits in numbers
- Add function to count digits in range of numbers
- Indicate unread message to the left (`[<]`) and to the right (`[>]`) of the range.
- Refactor tabs (divide `status_bar_draw()` with `_status_bar_draw_tabs()` function) and don't draw them if `/statusbar maxtabs` set to `0`.

## Revision 2
 - Remove excessive debug comments

H3rnand3zzz:
If code is ok, I will link issues in the commit message and try to add some details there.

H3rnand3zzz:
## Revision 3
Improve commit message.

H3rnand3zzz:
## Revision 4
Calculate range bounds only once and pass it as arguments to width calculating function to avoid excessive computations.

src/ui/statusbar.c
```diff
@@ -339,21 +356,23 @@ status_bar_draw(void)
             pos = _status_bar_draw_bracket(FALSE, pos, "]");
         }
     }
-
-    wnoutrefresh(statusbar_win);
-    inp_put_back();
+    return pos;
 }
 
+// Checks if there are messages to the side of the current displayed tabs range
```
jubalh, line 362: which side?

src/ui/statusbar.c
```diff
@@ -364,44 +383,49 @@ _extended_new(void)
...
+    if (!prefix && end > opened_tabs - 1) {
+        return pos;
+    }
+    gboolean is_current = FALSE;
 
-        pos = _status_bar_draw_bracket(is_current, pos, "[");
+    pos = _status_bar_draw_bracket(is_current, pos, "[");
 
-        int status_attrs;
-        if (is_current) {
```
jubalh, line 380: actually i liked those 3 comments to easily see whats what. in the new code its harder to find this.

src/ui/statusbar.c
H3rnand3zzz: > which side?
outside

jubalh: Its missing in the comment, no? :)

src/ui/statusbar.c
H3rnand3zzz: sorry, what comments?

H3rnand3zzz: anyway, there are only 2 states: new messages to the side, e.g. **[<]** or no new messages there `[<]` and `all other` comment doesn't make much sense and certainly doesn't improve code clarity.

src/ui/statusbar.c
H3rnand3zzz: > Its missing in the comment, no? :)

but it's kind of self-explanatory with `left_side` argument. I offered before to use doxygen style... I will add that the side is based on the argument.

src/ui/statusbar.c
H3rnand3zzz: Hopefully now the comments are satisfying.

H3rnand3zzz:
### Revision 5
Addressed the issues that were pointed out in the code review.

H3rnand3zzz:
### Revision 6
Rebased on master

sjaeckel (commented):
1. IIUC this changes the default behavior of the statusbar.
TBH I don't think that we should do that w/o at least having an option to disable it and IMO this new style shouldn't be the new default, but an option that can be enabled.

2. I didn't look yet in detail at the changes, but I've one question: Did you test whether the `actlist` mode also still works? I guess so, since it seems to be a completely different code path, but I prefer to be on the safe side.

H3rnand3zzz:
> 1. IIUC this changes the default behavior of the statusbar.
>        TBH I don't think that we should do that w/o at least having an option to disable it and IMO this new style shouldn't be the new default, but an option that can be enabled.

How is the change making it worse? Any objective reason why it shouldn't be the default except that it changes old clunky behaviour that doesn't even show current tab when you exceed 10 tabs (`maxtabs` maximum)?

jubalh:
The description of this PR is quite confusing to me. I read it several times and still wasn't sure what it does and how it changes the current behaviour.
AFAIK we already have `<` and `>` when a message in a window to the left/right appears.

So I pulled the changes and tested what happens.
What I saw is that if I have maxtabs set to 2. Then in our old behaviour we always saw the first 2 windows no matter which one was the active one. With this PR it always displays the current window and the closest ones around it.

So I think for some people this can be helpful to know where they are.
I'm not sure whether it should be the default behaviour or should be configurable.

What I didn't understand:
When I set maxtabs to 2. And navigate to window 3. Then I see window 2 and 4 also in the tabs on the statusbar. Didn't I configure it to show max 2 tabs? Before this PR we only display 2 tabs.

This is the only change I noticed when trying the PR. Was there something else that I did miss?

jubalh:
Quite often your PRs/issues contain a huge wall of text but somehow fail at bringing the information across to the reader.

If I understand your PR now correctly, then it should have been quite easy to understand it with a better description.
Something along the lines:
```
This PR changes the current behaviour which displays always the first tabs until `maxtabs` to display the X tabs around the currently selected tab. So if we are having a maxtab of 1 and the actively selected window is 2. Then 2 is displayed only.

So far we only displayed `>` to indicate there are more windows. Since this PR modifies the tabs that are displayed we also have an `<` indicator now for the indicating windows to the left side of the currently selected ones.
```

It's quite late and I'm tired so the words might not be perfect. But I guess you get what I mean with a better explanation.

You instead say:

```
    Indicate unread message to the left ([<]) and to the right ([>]) of the range.
```

And I thought: ok, dont we do that already? Or whats the current behaviour? Should I change my whole settings to understand which options are available and check how it differs?

And:

```
navigate through pages easily
```

Where I dont understand what it means at all. I can always navigate through the open windows easily...

So now that I understood what actually changes: I don't think this should just be the new default behaviour without any notice of it anywhere. No notice in the commit message or PR that the default behaviour changes  and how...

The proposed change itself is fine. But right now I believe that there should be a setting to use this behaviour or our current default. Then there is actlist too etc. So there will be 3 ways. There needs to be a well thought through setting with explanation for it.

H3rnand3zzz:
> The proposed change itself is fine. But right now I believe that there should be a setting to use this behaviour or our current default. Then there is actlist too etc. So there will be 3 ways. There needs to be a well thought through setting with explanation for it.

There is no objective reason to keep old behaviour. Current PR fixes 2+ issues, it's not a feature, I reclassify it as a "fix".

jubalh:
> There is no objective reason to keep old behaviour. Current PR fixes 2+ issues, it's not a feature, I reclassify it as a "fix".

One reason could be that I open my windows in a certain order. Could be the 5 first windows are the ones that are most important to me. I always want to see the tabs for those windows. I immediately want to see when somebody writes me on those windows. The rest might be windows that I don't care much about. That I cycle through with alt+a when I have extra free time. Your PR would break this workflow.

Another reason could be that I don't like the UI to change all the time. I don't want that the titlebar/statusbar changes when I change the focus of a window. Maybe I only want that the content of the window changes and the rest should look stable.

So there are 2 objective reasons. If I think harder I could probably find more.

Just because something isn't your use case or to your liking, doesn't mean it doesn't exist.

H3rnand3zzz:
`static` tabmode has been added as requested.

jubalh:
Much better. But now you did the opposite of:

> IIUC this changes the default behavior of the statusbar.
> TBH I don't think that we should do that w/o at least having an option to disable it and IMO this new style shouldn't be > the new default, but an option that can be enabled.

And the text is misleading again. Since you write `static` mode has been added. But compared to master it is actually your `dynamic` mode that has been added. So I would change the second commit so that `static` is still the default. And the new option is called `dynamic`.

Like this the behaviour will stay the same for users but they can choose to change to the dynamic or actlist modes.

jubalh (changes requested):
Please change 2nd commit to actually preserve the default behaviour as default and add an option for your new mode with the name `dynamic`.

H3rnand3zzz:
# Poll
Old style to the left, new style to the right. Upvote for a new style, downvote for old style. Don't mind the margin, pay attention only to the pages being displayed (ease of navigation).

### Win 1
image

### Win 5
image

### Win 7
image

### Win 8
image

H3rnand3zzz:
## New revision
Changed commit. Currently it makes a good history (first introduced a change that fixes the issues, then introduce new change that makes new behaviour non-default), commits contain sufficient explanations for the changes.
`default` is renamed to `static`, as it makes more sense based on the current changes.

src/command/cmd_defs.c
```diff
@@ -1300,7 +1300,7 @@ static const struct cmd_t command_defs[] = {
       CMD_ARGS(
               { "maxtabs <value>", "Set the maximum number of tabs to display, <value> must be between 0 and 10." },
               { "tablen <value>", "Set the maximum number of characters to show as the tab name, 0 sets to unlimited." },
-              { "tabmode default|actlist", "Set the mode how the 'active tabs' are shown." },
+              { "tabmode static|dynamic|actlist", "Set the mode tabs are shown. `dynamic` is a mode that modernizes paging by displaying tabs around current tab. `actlist` setting shows only active tabs. `static` setting always shows tabs in 1 to max_tabs range." },
```
jubalh, line 1303: Maybe describe only what it does. Remove `is a mode that modernizes paging by` displaying-> displays.

src/config/preferences.c
```diff
@@ -2276,7 +2276,7 @@ _get_default_string(preference_t pref)
     case PREF_STATUSBAR_ROOM_TITLE:
         return "name";
     case PREF_STATUSBAR_TABMODE:
-        return "default";
+        return "static";
```
jubalh, line 2279: users will have `default` in their config file I guess? So it won't find that setting. That means we need to add migration (there is a part where we do this for other settings already). However for simplicities sake I would just continue to use `default` for static. (So this also means to make the changes above). Alternative is to add migration.

jubalh (changes requested):
see comments

src/config/preferences.c
H3rnand3zzz: > users will have `default` in their config file I guess? So it won't find that setting. That means we need to add migration (there is a part where we do this for other settings already). However for simplicities sake I would just continue to use `default` for static. (So this also means to make the changes above). Alternative is to add migration.

in code logic, this doesn't matter since only "dynamic" and "actlist" are checked.

H3rnand3zzz:
# New revision
`static`->`default`, git commit updated, command description updated

H3rnand3zzz:
## New revision
 - Fix typo `paging`->`pagination`

jubalh:
Thanks for the changes and for the feature!

PR #1913, opened 2023-11-06
Fix `/omemo fingerprint` crashing in a MUC

sjaeckel:
Reported by a user in the profanity MUC.


jubalh:
Oh so the pointer was pointing to a random location and the auto_ tried to free it. Interesting situation. Will have to keep that in mind.

jubalh (approved):

PR #1915, opened 2023-11-07
Handle potential NULL pointers in database

sjaeckel:
Follow-up of 89dc7a46f13b452a1dba7fa8237ad5efbf8f3a2d resp. #1902

PR #1916, opened 2023-11-07
Improve handling of NULL pointers fetched from DB

H3rnand3zzz:
#1915 unsuccessfully tried to fix strdup(message) when message is NULL, since it lead to a chain reaction that can be ungracefully fixed like this: https://github.com/profanity-im/profanity/commit/c1c2113a3a2c66db160757d7169ee18610b3c29a or instead we could use strdup("") to substitute an empty message.


jubalh (changes requested):
Low quality description and commit message without any references. You know better than this.

H3rnand3zzz:
> Low quality description and commit message without any references. You know better than this.

Addressed.

jubalh (approved):

Issue #1925, opened 2023-11-11
Duplicate stanza-id warning for every message.


mdosch:
With profanity commit bba8278bb74a3b286b5a53004a1e88b051c66c2d I get the warning `Got a message with duplicate (server-generated) stanza-id from [REDACTED]` for every 1-1 message I received:

```
grep 9dd00770-38b1-4f5a-894c-1e03e708be99 ~/.local/share/profanity/logs/profanity.log
[...]

PR #1926, opened 2023-11-11
Do not check for unique `stanza-id` for MAM messages 

H3rnand3zzz:
See commit message for details.

Fix https://github.com/profanity-im/profanity/issues/1925

### DIFF
--- a/src/database.c
+++ b/src/database.c
@@ -528,7 +528,8 @@ _add_to_db(ProfMessage* message, char* type, const Jid* const from_jid, const Ji
 
     // stanza-id (XEP-0359) doesn't have to be present in the message.
     // But if it's duplicated, it's a serious server-side problem, so we better track it.
-    if (message->stanzaid) {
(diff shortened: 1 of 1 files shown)


jubalh:
I wouldn't write `[Fix Duplicate stanza-id errors flood for MAM](https://github.com/profanity-im/profanity/pull/1926/commits/44d72e3b226462217576879b15dd0463beb9f4b5)` as the commit header since this is just what fixes 1925. But not what this change is really about. I would write:

```
Do not check for unique stanza-id in case of MAM message

Since we might have MAM messages already received earlier it doesn't make sense to require unique stanza-ids there.
Fix https://github.com/profanity-im/profanity/issues/1925.
```

jubalh (changes requested):
please change commit message

H3rnand3zzz:
> please change commit message

done

jubalh (approved):

PR #1927, opened 2023-11-11
Handle alt+enter as newline char

H3rnand3zzz:
See commit message.

### How to test functionality
Write a message and use alt+enter during writing to put a new character symbol there. Plain "Enter" for sending message/command should work as well as before.

### Why alt+enter?
Well, you see, shift+enter and ctrl+enter aren't that bindable, as you can see by using `showkey --ascii` command. Though shift/ctrl variants might be more preferable since they are used in some popular applications, we just can't bind on them using GNU Readline, while alt+enter does the work as well. 

I will be glad if someone can improve the solution to take keybinding from preference and allow custom settings for ctrl+enter, shift+enter etc. Though even current solution will suffice in terms of functionality and you can get used to it quite quickly.

### DIFF
--- a/docs/profanity.1
+++ b/docs/profanity.1
@@ -80,6 +80,9 @@ Mark current window for later reading with an attention flag.
 .BI ALT+m
 Switch between windows marked with the attention flag.
 .TP
+.BI ALT+ENTER
+Add newline character without sending a message/command.
+.TP
 .BI ALT+c
 Run external editor (see
 .BR profanity-editor (1))
--- a/src/ui/inputwin.c
+++ b/src/ui/inputwin.c
@@ -137,6 +137,7 @@ static int _inp_rl_subwin_pagedown_handler(int count, int key);
 static int _inp_rl_startup_hook(void);
 static int _inp_rl_down_arrow_handler(int count, int key);
 static int _inp_rl_send_to_editor(int count, int key);
+static int _inp_rl_print_newline_symbol(int count, int key);
 
 void
 create_input_window(void)
@@ -482,6 +483,7 @@ _inp_rl_addfuncs(void)
     rl_add_funmap_entry("prof_win_close", _inp_rl_win_close_handler);
     rl_add_funmap_entry("prof_send_to_editor", _inp_rl_send_to_editor);
     rl_add_funmap_entry("prof_cut_to_history", _inp_rl_down_arrow_handler);
+    rl_add_funmap_entry("prof_print_newline_symbol", _inp_rl_print_newline_symbol);
 }
 
 // Readline callbacks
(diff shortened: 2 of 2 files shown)


jubalh (changes requested):
Should be mentioned in the keybinding section in `docs/profanity.1` as well.

H3rnand3zzz:
> Should be mentioned in the keybinding section in `docs/profanity.1` as well.

Done, thx for quick review :)

jubalh (approved):

sjaeckel:
> Could anyone confirm that this feature is still functional please?

works for me!

Are you sure you have profanity in focus in the window?

mrusme:
The global `/etc/inputrc` on the other hand defaults to this configuration:

```
# /etc/inputrc: initialization file for readline
#
# For more information on how this file works, please see the
# INITIALIZATION FILE section of the readline(3) man page
#
# Quick dirty little note:
#  To get the key sequence for binding, you can abuse bash.
#  While running bash, hit CTRL+V, and then type the key sequence.
#  So, typing 'ALT + left arrow' in Konsole gets you back:
#    ^[[1;3D
#  The readline entry to make this skip back a word will then be:
#    "\e[1;3D" backward-word
#
# Customization note:
#  You don't need to put all your changes in this file.  You can create
#  ~/.inputrc which starts off with the line:
#    $include /etc/inputrc
#  Then put all your own stuff after that.
#

# do not bell on tab-completion
#set bell-style none

set meta-flag on
set input-meta on
set convert-meta off
set output-meta on

# Completed names which are symbolic links to
# directories have a slash appended.
set mark-symlinked-directories on

$if mode=emacs

# for linux console and RH/Debian xterm
# allow the use of the Home/End keys
"\e[1~": beginning-of-line
"\e[4~": end-of-line
# map "page up" and "page down" to search history based on current cmdline
"\e[5~": history-search-backward
"\e[6~": history-search-forward
# allow the use of the Delete/Insert keys
"\e[3~": delete-char
"\e[2~": quoted-insert

# gnome / others (escape + arrow key)
"\e[5C": forward-word
"\e[5D": backward-word
# konsole / xterm / rxvt (escape + arrow key)
"\e\e[C": forward-word
"\e\e[D": backward-word
# gnome / konsole / others (control + arrow key)
"\e[1;5C": forward-word
"\e[1;5D": backward-word
# aterm / eterm (control + arrow key)
"\eOc": forward-word
"\eOd": backward-word

# konsole (alt + arrow key)
"\e[1;3C": forward-word
"\e[1;3D": backward-word

# Chromebooks remap alt + backspace so provide alternative (alt + k)
"\ek": backward-kill-word

$if term=rxvt
"\e[8~": end-of-line
$endif

# for non RH/Debian xterm, can't hurt for RH/Debian xterm
"\eOH": beginning-of-line
"\eOF": end-of-line

# for freebsd console
"\e[H": beginning-of-line
"\e[F": end-of-line
$endif

# fix Home and End for German users
"\e[7~": beginning-of-line
"\e[8~": end-of-line
```

jubalh:
I'm not sure why it doesn't work for you.
Did you try to empty the global inputrc just for testing purposes?

As an alternative you could use ALT+c to open the editor and edit the multiline message there.

PR #1929, opened 2023-11-12
Add custom `assert()` (`PROF_ASSERT()`)

### LINKED ISSUE #1895: Change assert behavior
Currently, we have many "asserts" in a code that would cause a crash and not every time it is possible to reproduce, as well as dump might be unavailable. I propose to use wrapper over a default "assert" (or even a custom solution) which would add logging, which can provide us additional data on the issues.

@sjaeckel @pasis , I would gladly hear your opinion on this considering your level of expertise in deep topics as such.

H3rnand3zzz:
See commit message.

Valgrind not needed in this case.

* Cause assert failure or add custom assert that will certainly fail.

Resolve #1895

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -74,7 +74,8 @@ core_sources = \
 	src/plugins/themes.c src/plugins/themes.h \
 	src/plugins/settings.c src/plugins/settings.h \
 	src/plugins/disco.c src/plugins/disco.h \
-	src/ui/tray.h src/ui/tray.c
+	src/ui/tray.h src/ui/tray.c \
+	src/trace.h
 
 unittest_sources = \
 	src/xmpp/contact.c src/xmpp/contact.h src/common.c \
--- a/src/chatlog.c
+++ b/src/chatlog.c
@@ -36,7 +36,6 @@
 
 #include "config.h"
 
-#include <assert.h>
 #include <errno.h>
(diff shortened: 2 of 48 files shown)


H3rnand3zzz:
@jubalh @sjaeckel @pasis, if someone kindly can find a time to review it, it would be amazing.

jubalh:
Personally I fail to see the reason for it.
For this reason, plus pasis being the author of pppoat2, I thought it's best if pasis and sjaeckel review it.

H3rnand3zzz:
> Personally I fail to see the reason for it. For this reason, plus pasis being the author of pppoat2, I thought it's best if pasis and sjaeckel review it.

The reason is on plain site: to make debugging easier. If someone has assert problem, he can provide profanity.log containing a line that shows where the problem has happened. Otherwise we need coredump, which not all the users can provide.

jubalh:
That's exactly what I mean. For me it is default that under Linux we check things using gdb, get backtrace, receive a coredump from users.
For me this and https://profanity-im.github.io/issues.html seems to be enough and seems to be the standard way in the Linux world.
So for me it is not visible why we need this and what downsides it could bring. Which is why I will let someone else review it :)

jubalh:
Not a big fan of this. And since noone else reviewed yet I'll close.

PR #1931, opened 2023-11-13
Introduce `/redraw` command

H3rnand3zzz:
See commit message.

### DIFF
--- a/src/command/cmd_defs.c
+++ b/src/command/cmd_defs.c
@@ -2729,6 +2729,18 @@ static const struct cmd_t command_defs[] = {
               "/privacy logging off",
               "/privacy os off")
     },
+
+    { CMD_PREAMBLE("/redraw",
+                   parse_args, 0, 0, NULL)
+      CMD_MAINFUNC(cmd_redraw)
+      CMD_TAGS(
+              CMD_TAG_UI)
+      CMD_SYN(
+              "/redraw")
+      CMD_DESC(
+              "Redraw user interface. Can be used when some other program interrupted profanity or wrote to the same terminal and the interface looks \"broken\"." )
+    },
+
     // NEXT-COMMAND (search helper)
 };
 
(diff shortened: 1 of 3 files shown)


src/command/cmd_defs.c
```diff
@@ -2729,6 +2729,18 @@ static const struct cmd_t command_defs[] = {
...
+
+    { CMD_PREAMBLE("/redraw",
+                   parse_args, 0, 0, NULL)
+      CMD_MAINFUNC(cmd_redraw)
+      CMD_TAGS(
+              CMD_TAG_UI)
+      CMD_SYN(
+              "/redraw")
+      CMD_DESC(
+              "Redraw User Interface. Might be used in case if some other program interrupted and interface is \"broken\"." )
```
jubalh, line 2741: I'm not sure `in case if` is correct.  I think the `if` can be canceled. Personally I probably would just use the first sentence.

jubalh: I also believe `User` and `Interface` should be lower case.

jubalh (changes requested):
Just description. Rest looks ok.

jubalh:
I think some users use a keybinding for their terminal. But probably a command can be helpful.

src/command/cmd_defs.c
sjaeckel:
```suggestion
              "Redraw user interface. Can be used when some other program interrupted profanity or wrote to the same terminal and the interface looks \"broken\"." )
```

I also would just use the first sentence, but that's because I'm lazy :)

Here we're writing the help text for our users, which also shows up in the man pages, and I like the idea to include the "why" some command exists.

sjaeckel (approved):
Can be merged after description is updated.

H3rnand3zzz:
updated

jubalh (approved):

PR #1932, opened 2023-11-14
Fix mem leak in `conf_string_list_add`

H3rnand3zzz:
See commit message.

### DIFF
--- a/src/config/conflists.c
+++ b/src/config/conflists.c
@@ -65,14 +65,16 @@ conf_string_list_add(GKeyFile* keyfile, const char* const group, const char* con
     }
 
     // Add item to the existing list
-    gchar** new_list = g_new(gchar*, length + 2);
+    const gchar** new_list = g_new(const gchar*, length + 2);
     for (gsize i = 0; i < length; ++i) {
-        new_list[i] = g_strdup(list[i]);
+        new_list[i] = list[i];
     }
-    new_list[length] = g_strdup(item);
+    new_list[length] = item;
     new_list[length + 1] = NULL;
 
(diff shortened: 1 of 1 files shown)


jubalh:
I checked the MUC and saw that @sjaeckel diff looked like:

           ```
           diff --git a/src/config/conflists.c b/src/config/conflists.c
           index 4032878e..838b611e 100644
           --- a/src/config/conflists.c
           +++ b/src/config/conflists.c
           @@ -51,3 +51,3 @@ conf_string_list_add(GKeyFile* keyfile, const char* const group, const char* con
                gsize length;
           -    auto_gcharv gchar** list = g_key_file_get_string_list(keyfile, group, key, &length, NULL);
           +    gchar** list = g_key_file_get_string_list(keyfile, group, key, &length, NULL);

           @@ -62,2 +62,3 @@ conf_string_list_add(GKeyFile* keyfile, const char* const group, const char* con
                    if (strcmp(list[i], item) == 0) {
           +            g_strfreev(list);
                        return FALSE;
           @@ -69,3 +70,3 @@ conf_string_list_add(GKeyFile* keyfile, const char* const group, const char* con
                for (gsize i = 0; i < length; ++i) {  
           -        new_list[i] = g_strdup(list[i]);
           +        new_list[i] = list[i];
                }
           @@ -73,2 +74,3 @@ conf_string_list_add(GKeyFile* keyfile, const char* const group, const char* con
                new_list[length + 1] = NULL;
           +    g_free(list);

           ```

H3rnand3zzz:
The solution has been improved.

sjaeckel (approved):

sjaeckel:
> The solution has been improved.

Indeed, looks very good now!

PR #1933, opened 2023-11-21
Improve string convertion error handling

H3rnand3zzz:
A bit hard to reproduce. You need to add a plugin with the following code (or any other plugin with `prof_pre_chat_message_display` present, such as emoticons.py):
```python
def prof_pre_chat_message_display(barejid, resource, message):
    return message
```

and then receive/send message with incorrectly encoded character, it can be any invalid UTF8 symbol. You'll see error in console, errors don't represent actual errors in plugins, but rather Profanity's shortcoming, though it make appear so that the problem is in plugin.

Actual proper handling would likely be using `y` instead of `s` format (see [reference](https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue))

> s ([str](https://docs.python.org/3/library/stdtypes.html#str) or None) [const char *]
>    Convert a null-terminated C string to a Python [str](https://docs.python.org/3/library/stdtypes.html#str) object using 'utf-8' encoding. If the C string pointer is NULL, None is used.

to

> y ([bytes](https://docs.python.org/3/library/stdtypes.html#bytes)) [const char *]
>     This converts a C string to a Python [bytes](https://docs.python.org/3/library/stdtypes.html#bytes) object. If the C string pointer is NULL, None is returned.

since there is a [problem](https://docs.python.org/3/c-api/arg.html) with `s`:
> s ([str](https://docs.python.org/3/library/stdtypes.html#str)) [const char *]
>    Convert a Unicode object to a C pointer to a character string. A pointer to an existing string is stored in the character pointer variable whose address you pass. The C string is NUL-terminated. The Python string must not contain embedded null code points; if it does, a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) exception is raised. Unicode objects are converted to C strings using 'utf-8' encoding. **If this conversion fails**, a [UnicodeError](https://docs.python.org/3/library/exceptions.html#UnicodeError) is raised.

In python such problem can be handled using `errors='ignore'` in [bytes.decode](https://docs.python.org/3/library/stdtypes.html#bytes.decode) or in a more sophisticated manner, depending on needs and realization.

### DIFF
--- a/src/plugins/python_plugins.c
+++ b/src/plugins/python_plugins.c
@@ -318,6 +318,11 @@ python_pre_chat_message_display_hook(ProfPlugin* plugin, const char* const barej
 {
     disable_python_threads();
     PyObject* p_args = Py_BuildValue("sss", barejid, resource, message);
+    if (!p_args) {
+        log_warning("Unable to convert strings in `python_pre_chat_message_display_hook` to 'UTF8'. barejid: %s, resource: %s, message: %s", barejid, resource, message);
+        allow_python_threads();
+        return NULL;
+    }
     PyObject* p_function;
 
     PyObject* p_module = plugin->module;
@@ -408,6 +413,12 @@ python_pre_room_message_display_hook(ProfPlugin* plugin, const char* const barej
 {
     disable_python_threads();
     PyObject* p_args = Py_BuildValue("sss", barejid, nick, message);
+    if (!p_args) {
+        log_warning("Unable to convert strings in `python_pre_room_message_display_hook` to 'UTF8'. barejid: %s, nick: %s, message: %s", barejid, nick, message);
+        allow_python_threads();
+        return NULL;
+    }
+
     PyObject* p_function;
 
     PyObject* p_module = plugin->module;


jubalh (approved):
Thanks for the good description. Actually I like it so much that I think maybe it would be good to add the description of the PR also at the end of the already existing commit description? I think combined they give a good picture of whats going on.

H3rnand3zzz:
> Thanks for the good description. Actually I like it so much that I think maybe it would be good to add the description of the PR also at the end of the already existing commit description? I think combined they give a good picture of whats going on.

It looks a bit too long for a commit description and formatting is probably going to be broken, so it would be better to just reference this PR link in the commit description + fix typos there. In a few hours I am going to do just that.

sjaeckel:
> It looks a bit too long for a commit description and formatting is probably going to be broken, so it would be better to just reference this PR link in the commit description + fix typos there

IMO there's no "too long" commit message if it's reasonable.

GH is not guaranteed to stay as long as the commit history and I'd also prefer to have it in the commit message.

H3rnand3zzz:
### New revision

Descriptions are concatenated.

jubalh:
Thanks!

Issue #1934, opened 2023-11-23
Broken scrolling


H3rnand3zzz:
### Problem
Long messages in history cause profanity to "underscroll" (on top/bottom does not actually scroll until that 

### Explanation
NCurses has let's call it a "buffer" with the size of `_maxy`, something like a hidden "window" and we can move inside of it. We have our "buffer" as well, our buffer is 200 lines (my new limit, before it was 1000), while ncurses' buffer is just 999 lines. When we get e.g. 10 100-line messages, buffer gets overwhelmed and when they are near the scrolling area end (e.g. end or start of history), we are unable to reach them, since profanity tries to "redraw" e.g. 3000 lines, while ncurses will only show last 999 of them (now I noticed that there is a problem even with my theory since last messages are affected by the problem as well).

### Potential solution: count lines in our buffer -- by counting '\n' symbols in each record
           and keeping "total" number, which is adjusted on each deletion/addition for faster
           calculations.

H3rnand3zzz:
Problem still persists on alt+scrolling (underscrolling)

PR #1935, opened 2023-11-23
Fix scrolling

H3rnand3zzz:
Current state:
 - Fixes "overscrolling"
 - Lag reduction (only necessary lines are being drawn)

Currently it's full of debugging lines with some issues which are to be fixed before release:
- [x] Fix "underscrolling"
- [x] Fix "overscrolling" (profanity jumps over big messages received from DB)
- [x] Change line calculation (made based on curses' position)
- [x] Cleanup debug lines
- [x] Merge
- [x] Fix a problem: extremely big message (e.g. 3k lines) causes profanity to jump back once scrolled on top

#### This line is for github CI/CD system to link the issue
Fix https://github.com/profanity-im/profanity/issues/1934

### DIFF
--- a/src/database.c
+++ b/src/database.c
@@ -333,7 +333,7 @@ log_database_get_previous_chat(const gchar* const contact_barejid, const char* s
     auto_gchar gchar* end_date_fmt = end_time ? end_time : g_date_time_format_iso8601(now);
     auto_sqlite gchar* query = sqlite3_mprintf("SELECT * FROM ("
                                                "SELECT COALESCE(B.`message`, A.`message`) AS message, "
-                                               "A.`timestamp`, A.`from_jid`, A.`to_jid`, A.`type`, A.`encryption` FROM `ChatLogs` AS A "
+                                               "A.`timestamp`, A.`from_jid`, A.`to_jid`, A.`type`, A.`encryption`, A.`stanza_id` FROM `ChatLogs` AS A "
                                                "LEFT JOIN `ChatLogs` AS B ON (A.`replaced_by_db_id` = B.`id` AND A.`from_jid` = B.`from_jid`) "
                                                "WHERE (A.`replaces_db_id` IS NULL) "
                                                "AND ((A.`from_jid` = %Q AND A.`to_jid` = %Q) OR (A.`from_jid` = %Q AND A.`to_jid` = %Q)) "
@@ -365,8 +365,10 @@ log_database_get_previous_chat(const gchar* const contact_barejid, const char* s
         char* to_jid = (char*)sqlite3_column_text(stmt, 3);
         char* type = (char*)sqlite3_column_text(stmt, 4);
         char* encryption = (char*)sqlite3_column_text(stmt, 5);
+        char* id = (char*)sqlite3_column_text(stmt, 6);
 
         ProfMessage* msg = message_init();
+        msg->id = id ? strdup(id) : NULL;
         msg->from_jid = jid_create(from);
         msg->to_jid = jid_create(to_jid);
         msg->plain = strdup(message ?: "");
--- a/src/ui/buffer.c
+++ b/src/ui/buffer.c
@@ -50,23 +50,29 @@
 #include <curses.h>
 #endif
 
+#include "log.h"
 #include "ui/window.h"
 #include "ui/buffer.h"
 
-#define BUFF_SIZE 200
+#define MAX_BUFFER_SIZE     200
+#define STRDUP_OR_NULL(str) ((str) ? strdup(str) : NULL)
 
 struct prof_buff_t
 {
     GSList* entries;
+    int lines;
 };
 
 static void _free_entry(ProfBuffEntry* entry);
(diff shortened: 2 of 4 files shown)


H3rnand3zzz:
### New revision

 - Git commit description is removed for now since a lot of things are subject to change.
 - New approach: let ncurses calculate lines instead of trying to do it by ourselves. Passing numbers of lines in the buffer helped to resolve overscrolling problem and can be handy in the future (jumping to some messages (unread, search), visual selection of messages etc.).
 - Underscrolling is broken again since some numbers are off in line calculation (though they are mostly correct, it's likely caused by some edge cases), requires further investigation.

H3rnand3zzz:
### New revision
 - Commit message that outlines changes
 - Merged with the latest changes
 - Added fetching `stanza_id` from the DB (changes are highly interconnected)
 - Removed debug messages

H3rnand3zzz:
### New revision
 - Comments are removed as per @sjaeckel's request.

jubalh:
> Pass id from DB (I am not sure if it's correct, probably stanza-id should be passed, but also we should consider LMC case)

What were your thoughts here exactly? I think id that you used is correct.

src/ui/window.c
```diff
@@ -656,15 +654,20 @@ win_page_up(ProfWin* window, int scroll_size)
...
             }
+
+            int buff_size = buffer_size(window->layout->buffer);
+            int offset_entry_id = buff_size > 10 ? 10 : buff_size - 1;
+            int offset = buffer_get_entry(window->layout->buffer, offset_entry_id)->y_end_pos;
+            *page_start = offset - page_space;
         }
     }
 
-    // went past beginning, show first page
```
jubalh, line 662: why is this comment removed? I think its helpful to quickly see whats going on here.

src/ui/window.c
```diff
@@ -656,15 +654,20 @@ win_page_up(ProfWin* window, int scroll_size)
...
-    if (*page_start < 0)
+    if (*page_start < 0) {
         *page_start = 0;
+    }
 
     window->layout->paged = 1;
-    // update only if position has changed
```
jubalh, line 667: same as above

H3rnand3zzz:
### New revision
Addressed the issues outlined in the review.

src/ui/window.c
```diff
@@ -715,7 +723,7 @@ win_page_down(ProfWin* window, int scroll_size)
         *page_start = total_rows - page_space - 1;
 
     window->layout->paged = 1;
-    // update only if position has changed
```
jubalh, line 718: same

H3rnand3zzz:
addressed

jubalh:
>  Fix a problem: extremely big message (e.g. 3k lines) causes profanity to jump back once scrolled on top

Was this already the case before your change?
I'm not even sure 3k lines fit on a screen? :D

H3rnand3zzz:
### New revision
 - Removed buffer limitation based on amount of lines in messages: it might affect users who don't use history, the approach has a good idea, but requires more careful implementation (e.g. limitation on amount of lines drawn as opposed to stored lines)
 - Description updated
 - Rebased on master

jubalh:
>  it might affect users who don't use history, the approach has a good idea, but requires more careful implementation

Affect how? Should we wait with merging/reviewing the code before this is solved? I would prefer to not include something "wrong" in master because several of our users use master to help us find bugs. So this might annoy them or make them report something that we are aware of.

H3rnand3zzz:
> > it might affect users who don't use history, the approach has a good idea, but requires more careful implementation
> 
> Affect how? Should we wait with merging/reviewing the code before this is solved? I would prefer to not include something "wrong" in master because several of our users use master to help us find bugs. So this might annoy them or make them report something that we are aware of.

It might've affected in a previous version, currently it was removed, hence no one should be affected in any negative way (at least I don't see a problem currently, I might've missed something obviously).

src/database.c
```diff
@@ -365,8 +365,10 @@ log_database_get_previous_chat(const gchar* const contact_barejid, const char* s
         char* to_jid = (char*)sqlite3_column_text(stmt, 3);
         char* type = (char*)sqlite3_column_text(stmt, 4);
         char* encryption = (char*)sqlite3_column_text(stmt, 5);
+        char* id = (char*)sqlite3_column_text(stmt, 6);
 
         ProfMessage* msg = message_init();
+        msg->id = id ? strdup(id) : NULL;
```
jubalh, line 371: I see you introduced `STRDUP_OR_NULL` in `buffer.c`. I wonder if it would be worth it to have this in common.h and use it here (and in other places) as well? @sjaeckel ?

src/ui/buffer.c
```diff
@@ -50,23 +50,30 @@
 #include <curses.h>
 #endif
 
+#include "log.h"
 #include "ui/window.h"
 #include "ui/buffer.h"
 
-#define BUFF_SIZE 200
+#define BUFF_SIZE           200
```
jubalh, line 57: Could you add a comment for `BUFF_SIZE` and `BUFF_SIZE_LINES` to short explain what they stand for?

H3rnand3zzz: I renamed the variable to self-describing name.

H3rnand3zzz:
### New revision
 - Fixed the issue pointed out above (buffer wasn't limited at all)

src/database.c
jubalh: @sjaeckel ping

PR #1937, opened 2023-12-05
Fix scrolling state at windows edge

H3rnand3zzz:
See commit message, it's descriptive.

No memory stuff -> no need for valgrind.

### DIFF
--- a/src/ui/win_types.h
+++ b/src/ui/win_types.h
@@ -146,9 +146,16 @@ typedef enum {
     WIN_VCARD
 } win_type_t;
 
+typedef enum {
+    WIN_SCROLL_INNER,
+    WIN_SCROLL_REACHED_TOP,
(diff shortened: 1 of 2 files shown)


jubalh:
BTW if you dont write "see commit message" and have only 1 commit. Then the commit message should be taken automatically. So with one commit it doesn't make sense to write this :)

src/ui/win_types.h
```diff
@@ -146,9 +146,16 @@ typedef enum {
     WIN_VCARD
 } win_type_t;
 
+typedef enum {
+    WIN_SCROLL_FREE,
```
jubalh, line 150: This means neither top nor bottom. Maybe an english native can suggest a better name than "free". Something like "inbetween", "medial", "inner" maybe?

H3rnand3zzz: Done

PR #1944, opened 2023-12-28
Fix when feature discovery is finished

### LINKED ISSUE #1940: Wrong error about server not supporting MAM.
On startup profanity shows the error `Server doesn't support MAM (urn:xmpp:mam:2).` although the server supports MAM. 

## Expected Behavior

It should only show that error if the server really doesn't support MAM.

## Current Behavior
On startup profanity shows the error `Server doesn't support MAM (urn:xmpp:mam:2).` although the server supports MAM. 

1. Start profanity and check messages in console window. (Don't know if this only happens in my set up or in general)

## Environment
* Debian Testing (trixie)

```
profanity --version
Profanity, version 0.14.0dev.master.0e66cbe4
Copyright (C) 2019 - 2023 Michael Vetter <jubalh@iodoru.org>.

Build information:
XMPP library: libstrophe
OTR support: Disabled
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Disabled
GTK icons/clipboard: Disabled
GDK Pixbuf: Enabled
```

sjaeckel:
Feature discovery was marked as finished once we received a reply to the initial request. The discovery mechanism allows to delegate the real feature discovery to another service running on different domain and those requests are created dynamically.

This was another instance causing the warning message described in #1940

Fixes #1940 (once more)

* have an account on a server that delegates feature discovery like hot-chilli
* send a 1-on-1 message to that account while being offline
* connect to that account

Before the patch you would receive the warning when connecting to the server and the server delivers the 1-on-1 message before feature discovery was finished.

### DIFF
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -58,6 +58,7 @@
 #include "ui/window.h"
 #include "tools/bookmark_ignore.h"
 #include "xmpp/xmpp.h"
+#include "xmpp/iq.h"
 #include "xmpp/muc.h"
 #include "xmpp/chat_session.h"
 #include "xmpp/roster_list.h"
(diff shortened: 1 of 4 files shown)


jubalh (approved):

jubalh:
Thanks!

Issue #1955, opened 2024-02-17
Slashguard doesn't work


mdosch:
I realized that slashguard stopped working

## Expected Behavior
Slashguard won't let you send a message like `s/test`.

## Current Behavior
It's possible to send `s/test`

## Steps to Reproduce (for bugs)

1. Type `s/test` into a chat.
2. See that the message is not blocked by slashguard but sent.

## Environment
Debian testing/unstable 

```
profanity --version
Profanity, version 0.14.0dev.master.569e37f0
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2024 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Disabled
PGP support: Enabled (libgpgme 1.18.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Disabled
GTK icons/clipboard: Disabled
GDK Pixbuf: Enabled
```

jubalh:
Probably has to do with 7eac636fc80016dbf18eea60a7d59576dd2a4925 and `inp_callback()`.

jubalh:
Solved in https://github.com/profanity-im/profanity/commit/88b26cfdb9bd8468c9b7f9131dea55ed35bbd751 by reverting gmainloop changes.

PR #1956, opened 2024-02-17
Fix slashguard

jubalh:
* Need more time to investigate the callback stuff.
* Need to check whether we need the slashguard check in two locations or not. If we do put it into a func.

Fix https://github.com/profanity-im/profanity/issues/1955


jubalh:
@H3rnand3zzz can you review?

Issue #1970, opened 2024-05-31
Reject subscription requests when using XEP-0377: Spam Reporting 


mdosch:
When I receive spam and and block/report via `/blocked report-abuse JID` or `/blocked report-spam JID` I'd also like subscription requests from the spammer/abuser to be denied if present.

## Expected Behavior
Deny the subscription request of the spammer/abuser if present.

## Current Behavior
I have to deny manually afterwards.

## Possible Solution
Automatically deny any subscription requests if someone is reported as spammer/abuser.

## Steps to Reproduce (for bugs)

1. Receive spam including a subscription request
2. Use `/blocked report-spam spammer@spamserver.example`
3. Check `/sub received`
4. See the spammers request still lingering there.

## Context
I'm currently hit by a spam wave and reporting/blocking with profanity is double the effort compared to conversations.

## Environment
* Debian Bookworm (stable)
* 
```
profanity --version
Profanity, version 0.14.0
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2023 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.18.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.11.2)
GTK icons/clipboard: Enabled
GDK Pixbuf: Enabled
```
* Server: Prosody trunk nightly build 1898 (2024-05-21, 6f840763fc73)

jubalh:
JFTR: Somehow I think the server should be doing this.

PR #1972, opened 2024-06-02
Remove subscription when reporting spam/abuse

jubalh:
When users receive a message and a subscription request from a new contact, and they use `/blocked report-abse JID` or `/blocked report-spam JID` they most likely want the subscription request to vanish as well.

According to mdosch Conversations is behaving like this too.

Fix: https://github.com/profanity-im/profanity/issues/1970


sjaeckel (approved):

jubalh:
@chenrui333 do you see why the build fails for macOS?

Issue #1974, opened 2024-06-11
statusbar actlist doesnt respect `/statusbar show` setting


jubalh:
statusbar actlist doesnt respect `/statusbar show` setting.

## Expected Behavior
`/statusbar tabmode default`
`/statusbar show name`
Shows the name in the statusbar tab.

`/statusbar tabmode actlist`
`/statusbar show name`
Should do the same

## Current Behavior
`/statusbar tabmode actlist`
`/statusbar show name`
Does not show the name in the statusbar tab.

sjaeckel:
FTR: I did this explicitly to only show numbers in #1813 since irssi's actlist also only shows numbers.

One could still want to see names instead, so this feature request is valid.

PR #1978, opened 2024-06-19
Make statusbar actlist respect the /statusbar show setting

jubalh:
This lets users use the actlist and decide if they want to see name or numbers.

Fix https://github.com/profanity-im/profanity/issues/1974


jubalh:
@weiss @sjaeckel could one of you review?

Issue #1946, opened 2024-01-09
Substantial memory leak on aarch64 and x86_64 architectures


kseistrup:
**TL**;**DR**: Profanity ate 1 GB swap in less than 15 days

I have profanity running on a RPi4 (aarch64 architecture) that is running ArchLinuxARM. Profanity receives notifications from e.g. cron jobs and casual interpersonal chat, altogether less than a couple of hundred messages daily — some days substantially less.

The RPi4 has 4 GB RAM + 1 GB swap. Usually no swap is used at all, so at first I didn't notice when `free -tk` reported that used and free swap had swapped places, so to speak: 100% used swap and 0% free. This was when the RPi4 and `profanity` had been up for slightly less than 15 days.

A `ps ux` revealed that the `profanity` process had gobbled up 2.5+ GB RAM, including all of the available 1 GB swap. Taking `profanity` down released all of the used swap, except ~40 MB.

## Expected Behavior

Ideally, `profanity` should not keep gobbling up RAM.

## Current Behavior

Launching `profanity` again, this is what I noticed (numbers taken from `ps ux`):

```txt
HH:MM     VSZ     RSS
-----··------··------
??:??  188208  163968
18:58  249260  225152
21:42  286228  262016
07:24  346020  321792
```

`HH:MM` is wallclock hours and minutes.

I tried the same thing on a x86_64 desktop pc, running ArchLinux:

```txt
HH:MM     VSZ     RSS
-----··------··------
??:??   55120   33976
18:13  280880   40264
22:13  316360   77128
07:21  360808  121544
```

## Possible Solution

I haven't got a clue.

## Steps to Reproduce (for bugs)

1. Launch `profanity` in a terminal emulator
2. Regularly run e.g. `ps ux | rg pro[f]anity` to see its increasing memory consumption

## Context

N/A

## Environment
* Give us the version and build information output generated by `profanity -v`

```sh
$ profanity -v
Profanity, version 0.14.0
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2023 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
[...]

PR #1979, opened 2024-06-19
Fix memleaks

jubalh:


jubalh:
I'm not sure what those glib list/hash map and ncurses related ones are that valgrind reports with:

```
G_DEBUG=gc-friendly G_SLICE=always-malloc valgrind --tool=memcheck --track-origins=yes --leak-check=full --leak-resolution=high --num-callers=30 --log-file=val --suppressions=prof.supp ./profanity
```

It seems to me that we always pass some free functions at init or call them later.
We added a lot of them to the surpress file for a reason probably.

But something has to cause: https://github.com/profanity-im/profanity/issues/1946

I'll try to look more but maybe we can merge this PR already.

Issue #1984, opened 2024-06-20
Enable substring autocompletion for `/msg`


jubalh:
People with huge contact lists most likely happen to have contacts starting with popular names more often.
They can use `/msg Arthur` to find the correct person.

But if they have 50 Arthurs it will take a while. Using the surname might be faster. So a substring search of `/msg Clarke<tab>` will be faster.

The question is if we want this as a general behaviour for roster contact completion or just for `/msg`? Opinions?

PR #1985, opened 2024-06-20
Implement substring autocompletion for /msg

jubalh:
People with huge contact lists most likely happen to have contacts starting with popular names more often. They can use /msg Arthur to find the correct person.

But if they have 50 Arthurs it will take a while. Using the surname might be faster. So a substring search of /msg Clarke<tab> will be faster.

Fix https://github.com/profanity-im/profanity/issues/1984

### DIFF
--- a/src/tools/autocomplete.c
+++ b/src/tools/autocomplete.c
@@ -57,7 +57,7 @@ struct autocomplete_t
     gchar* search_str;
 };
 
-static gchar* _search(Autocomplete ac, GList* curr, gboolean quote, search_direction direction);
+static gchar* _search(Autocomplete ac, GList* curr, gboolean quote, search_direction direction, gboolean substring);
 
 Autocomplete
 autocomplete_new(void)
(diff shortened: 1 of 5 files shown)


jubalh:
roster_contact_autocomplete() is used for:
* /msg
* /info
* /roster remvoe
* /roster group add
* /roster group remove
* /otr policy
* /pgp start
* /pgp sendpub
* /ox start
* /omemo start
* /omemo fingerprint
* /omemo trust
* /omemo untrust
* /win
* /invite send
* /status get
* /vcard get
* /vcard photo open
* /vcard photo save

@mdosch @weiss @sjaeckel 
What are your opinions? Should we generally make `roster_contact_autocomplete()` autocomplete substrings? Or should we do it as proposed in this PR. That we have two functions, one for the exact prefix and the other for substring.
If we do the latter, do we want it for more functions than `/msg`?

mdosch:
If you have a lot of Arthurs the issue of having to tab a lot till you find the right Arthur can happen with any command, so I think we should allow the substring search everywhere. But for the ordering I'd like to rank hits starting with the search term higher.

E.g. if you have "Arthur Anchor", Arthur Boyle", "Arthur Clarke", "Arthur Dent" and "Clark Kent" `/msg Clark<tab>` should show "Clark Kent" first and then "Arthur Clarke".

src/tools/autocomplete.c
```diff
@@ -245,7 +245,7 @@ autocomplete_contains(Autocomplete ac, const char* value)
 }
 
 gchar*
```
sjaeckel, line 247: `static`

sjaeckel (changes requested):
LGTM

and
1. I like the idea of @mdosch 
2. that's one of the cases where I think tests would make sense :)

PR #1998, opened 2024-10-03
Remove plenking

mdosch:
There were two cases of plenking (a space between a word and the punctuation
marks) in front of `…`.

Should not be necessary as only two strings were changed (space removed).

* If it builds it should be fine.


jubalh (approved):

jubalh:
Thanks!

PR #1999, opened 2024-10-12
Update .gitignore

andreaz98:
I know I should work on my own branch instead of PR to master but this file `core.20384`, which seems a random file, won't let me merge the OMEMO-colors feature (still a work in progress).

- [X] I ran valgrind when using my new feature

This isn't a functionality.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ breaks
 *.tar.*
 *.zip
 *.log*
+
+
+core.20384


sjaeckel:
This PR should not be merged. I'm not sure how or why this could prevent a merge without you committing the file to the branch you're trying to merge, but adding a single core dump file to the project gitignore is not an option.

I see two options, either we ignore all coredumps on project level or IMO the better way is that you configure your [global Git excludesfile](https://git-scm.com/docs/git-config/2.47.0#Documentation/git-config.txt-coreexcludesFile) and manage those kind of ignores yourself locally.

As I believe that we should go with the latter option, I'm closing this PR. Feel free to update this PR and re-open it if you believe that we should go with the former option.

PR #2007, opened 2024-11-27
Debug #1946

sjaeckel:
This is a starting point while debugging #1946.

The memory usage on current develop has increased significantly since 0.14.0.

The main memory saving was done by reverting d7e46d64fe6bc9926eb404dad6a4803e1fc2e8b2 which is not in 0.14.0.

I did not yet conclude on what the real problem is, but my guess is that the memory growth can only be tackled by rewriting some internal parts and we're not really leaking memory. AFAICT most memory is accounted for and is free'd on program termination.

NB: The valgrind output of a default build, executed with "our default valgrind settings" is kind of misleading. Each time a gnupg API is invoked it forks which causes the valgrind output to blow up and be hard to analyze. One way to improve the debug experience would be to disable following forks when running valgrind, another is to disable gnupg.

I've extended our default `prof.supp` by quite a bit, which must still be reviewed whether those excludes should be there or not.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -343,6 +343,14 @@ clean-local:
 	rm -f $(git_include) $(git_include).in
 endif
 
+.PHONY: my-prof.supp
+my-prof.supp:
+	@sed '/^# AUTO-GENERATED START/q' prof.supp > $@
+	@cat /usr/share/glib-2.0/valgrind/glib.supp >> $@
+	@wget -O- https://raw.githubusercontent.com/python/cpython/refs/tags/v`python3 --version | cut -d' ' -f2`/Misc/valgrind-python.supp >> $@
+	@test -z "@GTK_VERSION@" || wget -O- https://raw.githubusercontent.com/GNOME/gtk/refs/tags/@GTK_VERSION@/gtk.supp >> $@
+	@test -z "@GTK_VERSION@" || cat /usr/share/gtk-3.0/valgrind/gtk.supp >> $@
+
 check-unit: tests/unittests/unittests
 	tests/unittests/unittests
 
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,7 @@ toc_fragment.html
 
 # valgrind files
 valgrind.out
+my-prof.supp
 
 # Generate docs
 apidocs/python/_build/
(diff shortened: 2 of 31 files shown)


Makefile.am
```diff
@@ -343,6 +343,14 @@ clean-local:
 	rm -f $(git_include) $(git_include).in
 endif
 
+.PHONY: my-prof.supp
+my-prof.supp:
+	@sed '/^# AUTO-GENERATED START/q' prof.supp > $@
+	@cat /usr/share/glib-2.0/valgrind/glib.supp >> $@
+	@wget -O- https://raw.githubusercontent.com/python/cpython/refs/tags/v`python3 --version | cut -d' ' -f2`/Misc/valgrind-python.supp >> $@
+	@wget -O- https://raw.githubusercontent.com/GNOME/gtk/refs/tags/`pkg-config --modversion gtk+-3.0`/gtk.supp >> $@
```
jubalh, line 351: This one isn't working: `https://raw.githubusercontent.com/GNOME/gtk/refs/tags/gtk.supp`
This one is though: `https://raw.githubusercontent.com/GNOME/gtk/refs/heads/main/gtk.supp`

jubalh:
Seems to work also with reconnects.
For some reason I see both on master and on this branch that the roster isn't displayed correctly upon a reconnect. I'll try to investigate further.

I'll keep this branch running for some time to see if I find more.

Makefile.am
sjaeckel: > This one is though: `https://raw.githubusercontent.com/GNOME/gtk/refs/heads/main/gtk.supp`

You most likely don't want the supp file for main, but for "your installed gtk version".

> This one isn't working: `https://raw.githubusercontent.com/GNOME/gtk/refs/tags/gtk.supp`

This is wrongly extended from 
```
https://raw.githubusercontent.com/GNOME/gtk/refs/tags/`pkg-config --modversion gtk+-3.0`/gtk.supp
```
, so it seems like `pkg-config --modversion gtk+-3.0` does not work for you.

Do you know where this fails for you?

What do the following commands return?

```
pkg-config --modversion gtk+-3.0
pkg-config --modversion gtk+-2.0
```

jubalh: Ah indeed. `gtk3-devel` wasn't installed. With that it works fine.

Issue #2013, opened 2025-01-18
return value of `prof_get_current_nick` cannot be freed


ventosus:
Either the function definition in `profapi.h` is wrong, i.e. schould be a `const char*` ...

    char* (*prof_get_current_nick)(void);

... or a `strdup` is missing somewhere, because a `free` of the returned value crashes profanity.

PR #2014, opened 2025-01-19
Make `muc_nick` return a `const char* const`

ventosus:
In the documentation of the `muc_nick` function we can read:

> The nickname is owned by the chat room and should not be modified or freed

We should reflect that in the return type and `strdup` the value for the plugin API.

Fixes: https://github.com/profanity-im/profanity/issues/2013


jubalh (approved):

jubalh:
@ventosus can you make the coding style change?

ventosus:
> @ventosus can you make the coding style change?

Done.

jubalh:
Thank you!

Issue #1190, opened 2019-09-13
Profanity shows old messages as new when connecting to a MUC


mdosch:
When I join a MUC (after reconnect or a fresh start) old messages are often shown as new with a current time stamp.

## Expected Behavior
Old messages are not shown as new but are in the right place of the timeline.

## Current Behavior
After connect old messages are shown as just received with the current date and time.

## Context
A message from `12/09/2019 22:28:34"` is shown as received at 13.09.19 07:52:39. Probably related to the second msg with the delay.

![](https://files.mdosch.de:64183/upload/ibzymzp80lpRvNB2/2019-09-13-075306_scrot.png)

```
grep -r "Martin, we don" ~/.local/share/profanity/logs/
/home/martin/.local/share/profanity/logs/profanity.log.04:12/09/2019 22:28:34: xmpp: DBG: RECV: <message id="5e0efe4f361c4653b58fc9888b7bc145" lang="en" to="martin@mdosch.de/NM2oPdUi" type="groupchat" from="profanity@rooms.dismail.de/pep."><origin-id id="5e0efe4f361c4653b58fc9888b7bc145" xmlns="urn:xmpp:sid:0"/><body>Martin, we don't have the same target as dino or conversations users.</body><stanza-id id="5f5d31ad-6cb0-4ca2-a03f-5c1afaed754a" by="profanity@rooms.dismail.de" xmlns="urn:xmpp:sid:0"/></message>
/home/martin/.local/share/profanity/logs/profanity.log.05:13/09/2019 07:52:39: xmpp: DBG: RECV: <message id="5e0efe4f361c4653b58fc9888b7bc145" lang="en" to="martin@mdosch.de/bHDiwWui" type="groupchat" from="profanity@rooms.dismail.de/pep."><origin-id id="5e0efe4f361c4653b58fc9888b7bc145" xmlns="urn:xmpp:sid:0"/><body>Martin, we don't have the same target as dino or conversations users.</body><stanza-id id="5f5d31ad-6cb0-4ca2-a03f-5c1afaed754a" by="profanity@rooms.dismail.de" xmlns="urn:xmpp:sid:0"/><delay xmlns="urn:xmpp:delay" stamp="2019-09-12T20:28:34Z" from="rooms.dismail.de"/><x xmlns="jabber:x:delay" stamp="20190912T20:28:34" from="rooms.dismail.de"/></message>
```

## Environment
Debian bullseye (testing)

```
profanity --version
[...]

PR #2019, opened 2025-02-09
Bug Fix #1938

### LINKED ISSUE #1938: Wrong time stamps in messages (presumably) from MAM
When opening a MUC and receiving old messages (presumably from MAM) some messages showed the real time stamp, but some in between showed the current time stamp of the moment opening the chat.

_jLKZecbSUiictGeps-f3A

## Expected Behavior

All messages should show the time stamp when the message was sent.

## Current Behavior

Some messages show the time stamp from the moment the messages are received.

1. Profanity is not connected
2. Receive messages in a MUC you are participating in.
3. Connect profanity.
4. Compare time stamps of the fetched messages to the real time stamps (e.g. in another client).

## Environment
* Debian Testing (trixie)

```
profanity --version
Profanity, version 0.14.0dev.master.0e66cbe4
Copyright (C) 2019 - 2023 Michael Vetter <jubalh@iodoru.org>.

Build information:
XMPP library: libstrophe
OTR support: Disabled
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Disabled
GTK icons/clipboard: Disabled
GDK Pixbuf: Enabled
```

OdenDillenkoffer:
For the bug reported in #1938 there seems to be an issue of not grabbing the right time stamp. This is similar to the bug in #1190 of not grabbing the proper delay for MAM messages. To fix this, instead of discarding the original timestamp and trying to find the oldest delay, which may have been lost to a server quirk, we preserve the timestamp until an older delay is found. Then, and only then, will the timestamp update. Of course if it still can't find a delay and there is no timestamp, the fall back of grabbing the current time is still available. This should fix the bug experienced in #1938 and #1190 , given they look like they stem from a similar issue. 

* Step 1 - Profanity is not connected.
* Step 2 - Receive messages in a MUC
*Step 3 - Connect Profanity
Now the timestamps should be correct for the MUC.

Previous to this there were issues of inaccurate time stamps, but now that should be mitigated.

### DIFF
--- a/src/xmpp/message.c
+++ b/src/xmpp/message.c
@@ -1143,22 +1143,10 @@ _handle_groupchat(xmpp_stanza_t* const stanza)
         message->timestamp = oldest_delay;
     }
 
-    //1390 Bug fix explained
-    /*
-    Previously the code from 1136 to 1145 looked like this
-    bool is_muc_history = FALSE;
-    if (message->timestamp != NULL) {
-        is_muc_history = TRUE;
-        g_date_time_unref(message->timestamp);
-        message->timestamp = NULL;
-    }
+    //1938 Bug fix explained
+    //Instead of discarding the timestamp associated with the message without checking it
+    //First we check if we have an older delay, at which point we replace the time stamp with that.
 
-    // we want to display the oldest delay
-    message->timestamp = stanza_get_oldest_delay(stanza);
-
-    However, this is discarding a possibly correct timestamp instead of checking
-    it against the oldest delay. This could be a server quirk in that it doesn't work.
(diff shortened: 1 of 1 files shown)


jubalh (changes requested):
Thanks for this PR.
Before I look at the code itself. Please make sure you follow.
So first thing you have to do is find the contributing.md file. Read it.
Then you can check other PRs and see what they look like.

You describe the PR as comments in the code. You will find out that nobody else is doing this :) Use the body of the git commit message.

Once you are done you can click on request review.

jubalh:
Ping

Issue #1987, opened 2024-06-20
Disable notifications when present


avidseeker:
## Expected Behavior
When actively participating in a chat, don't send notification about new messages from that chat.

## Current Behavior
All notifications from currently active chat are shown.

## Possible Solution
Profanity has a presence detection that automatically sets "AFK" when no interaction is happening. When a chat window is open and user presence is detected, don't send notifications about the opened chat. Presence should not be detected when profanity is in a detached tmux session.

jubalh:
There is `/notify chat current on|off` and `/notify room current on|off`.

PR #2021, opened 2025-03-05
feat/1987-DisableSelWinNotifs

Bionicgitman:
Implementing #1987 's request.


jubalh (approved):
Looks good!

jubalh:
Thanks for this PR!
Any reason its just a draft yet?

Bionicgitman:
I have yet to include the AFK portion, I will submit it soon

sjaeckel (commented):
I just had a look at the existing implementation and found that we already have `PREF_NOTIFY_ROOM_CURRENT` resp. `PREF_NOTIFY_CHAT_CURRENT`. Did you try enabling those before making these changes?

The main difference is then the check for the idle time, which should happen inside `prefs_do_room_notify()` resp. `prefs_do_chat_notify()`  and the time should maybe be configurable? (CC @jubalh) The `ChatState` is only for 1on1 chats, so we can't use that.

Privat33r-dev (changes requested):
In general, as it was mentioned earlier, disabling notification for the current chat room is already enabled in a form of a preference, condition for which is being checked inside of the `prefs_do_chat_notify`, called to calculate `notify` variable.

https://github.com/profanity-im/profanity/blob/5385cc28beca6513ef2ad66a405b296f1507b73c/src/config/preferences.c#L334 

I suggest to instead make changes in the `prefs_do_chat_notify` and `prefs_do_room_notify`, using configurable preference for idling, as it was suggested earlier by @sjaeckel.

Bionicgitman:
Started working on adding idle time as an option. Tell me if there are any problems with it or if it isn't complete.

jubalh:
Well, what's strange for sure is that there are a couple of merge commits. Maybe get rid of them so the tree is clean. And in the future just rebase.

jubalh:
Is it possible to clean  up the commit history?

jubalh:
The commit history is even more of a mess now :)

jubalh:
> I just had a look at the existing implementation and found that we already have PREF_NOTIFY_ROOM_CURRENT resp. PREF_NOTIFY_CHAT_CURRENT. Did you try enabling those before making these changes?

Closing this for this reason and due to polluted history.

PR #2022, opened 2025-03-06
Memory fixes

jubalh:
WIP

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -4917,8 +4917,6 @@ cmd_sendfile(ProfWin* window, const char* const command, gchar** args)
         goto out;
     }
 
-    FILE* fh = fdopen(fd, "rb");
-
     gboolean omemo_enabled = FALSE;
     gboolean sendfile_enabled = TRUE;
 
@@ -4951,6 +4949,8 @@ cmd_sendfile(ProfWin* window, const char* const command, gchar** args)
         win_println(window, THEME_ERROR, "-", "Sending encrypted files via http_upload is not possible yet.");
         goto out;
     }
+
+    FILE* fh = fdopen(fd, "rb");
 
     if (omemo_enabled) {
 #ifdef HAVE_OMEMO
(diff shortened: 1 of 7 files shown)


src/command/cmd_funcs.c
```diff
@@ -4952,6 +4950,8 @@ cmd_sendfile(ProfWin* window, const char* const command, gchar** args)
         goto out;
     }
 
+    FILE* fh = fdopen(fd, "rb");
```
sjaeckel, line 4953:
```suggestion
    auto_FILE FILE* fh = fdopen(fd, "rb");
```

and remove the potential `fclose()` calls

sjaeckel (changes requested):
Basically LGTM

src/command/cmd_funcs.c
jubalh: This one is passed to HTTPUpload though. And we already free more of that stuff later on. So I think its best not to mix it to free some stuff and not others.

sjaeckel: yeah, right I missed that.

src/common.c
```diff
@@ -130,6 +130,36 @@ auto_free_char(char** str)
     free(*str);
 }
 
+/**
+ * Closes the file descriptor.
+ *
+ * @param fd file descriptor handle. If NULL, no action is taken.
+ */
+void
+auto_close_gfd(gint* fd)
```
sjaeckel, line 139: Just out of curiosity: why did you decide to have our own version instead of using `g_autofd`?

sjaeckel (changes requested):

src/common.c
jubalh: Never knew this existed. I only now found https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3007 when googling. devhelp doesnt seem to contain it?

jubalh: Should I use that or we want the log_error()

sjaeckel: Phew, good question ... I've never seen a failed `[f]close()`, so I'm not sure if there's any value in it. But maybe that's also only the case because nobody usually checks their return value?
Let's just keep using ours.

PR #2034, opened 2025-04-07
Documentation

SM9276:
I wrote some helpful comments for alot of functions that are found in the /src and also in /src/xmpp


sjaeckel (commented):
I'm absolutely no fan of comments where the only purpose is "being a prose form of the functionality implemented" without any further insights or details.

Also this will most likely never be updated and stand forever in that version, since my experience with FLOSS projects shows that nobody cares about comments/documentation besides the maintainers and I don't want to take care of them.

IMO this is wasted time (for writing) and space (in your mind when looking through/reading the code). _And_ taking the second point into consideration: a potential source for confusion in the future, _if_ those comments aren't updated.

Additionally to that, this is no library where someone is supposed to re-use any API. It makes sense that our python and C API's have documentation, but all the APIs in this PR are internal.

I appreciate that you want to improve this project, but from me this gets a :-1: and I wouldn't want this merged.

Issue #2059, opened 2025-08-06
Configure error on Cygwin.


Yard000:
Hello,
Spent a bit of time trying to build it on Cygwin. Installed a lot of libs and what not and in the end I cant solve this error:

checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking for glib-2.0 >= 2.62.0... yes
checking for gio-2.0... yes
checking for library containing fmod... none required
checking for libcurl >= 7.62.0... yes
checking for sqlite3 >= 3.22.0... yes
./configure: line 15241: syntax error near unexpected token `,'
./configure: line 15241: `ACX_PTHREAD(, [AC_MSG_ERROR([pthread is required])])'

Pthread is there:

$ ls -al /usr/include/pthread.h
-rw-r--r-- 1 Admin Administrators 11233 Jul 15 01:02 /usr/include/pthread.h

Any ideas?

jubalh:
BTW I don't own a Windows machine. And to the best of my knowledge none of the currently active developers do.
I assume that more things will not work on Windows. It would be cool to get a new contributor who takes care about that.

Regarding "just making it work" I would suggest to use [WSL](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux).

PR #2062, opened 2025-08-30
build: replace ACX_PTHREAD with AX_PTHREAD and vendor ax_pthread.m4 (fix Cygwin configure)

ritesh006:
# build: replace ACX_PTHREAD with AX_PTHREAD and vendor ax_pthread.m4 (fix Cygwin configure)

On Cygwin, running `./configure` failed with:

    ./configure: line XXXX: syntax error near unexpected token `,'
    ./configure: line XXXX: `ACX_PTHREAD([], [AC_MSG_ERROR([pthread is required])])'

**Root cause:** the legacy macro `ACX_PTHREAD` is not provided by modern
Autoconf distributions. The maintained macro is `AX_PTHREAD` (from
autoconf-archive).

**Fix:** - Replace `ACX_PTHREAD` with `AX_PTHREAD` in `configure.ac`. -
Ensure `AC_CONFIG_MACRO_DIR([m4])` is set. - Vendor `m4/ax_pthread.m4`
so builds from a git checkout succeed.

**Result:** - `autoreconf -fi && ./configure && make` now succeeds on
Cygwin x86_64. - Profanity builds and runs (`Version 0.15.1dev...`)
without further changes. - Linux builds continue to work as `AX_PTHREAD`
is widely available.

This resolves issue #2059.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -95,7 +95,7 @@ PKG_CHECK_MODULES([curl], [libcurl >= 7.62.0], [],
 PKG_CHECK_MODULES([SQLITE], [sqlite3 >= 3.22.0], [],
     [AC_MSG_ERROR([sqlite3 3.22.0 or higher is required])])
 
-ACX_PTHREAD([], [AC_MSG_ERROR([pthread is required])])
+AX_PTHREAD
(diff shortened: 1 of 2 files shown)


jubalh:
> for this any issue is created?

Nope, but you can just create a pull request.

PR #2067, opened 2025-09-13
Add types to function prototypes

mcalierno:
Running make (macOS, Clang 16.0.0) results in errors complaining about passing arguments without function prototypes.

Specifically:
- bookmark_ignore_on_connect()
- cons_show_qrcode()

Adding parameters to prototypes results in successful build.

### DIFF
--- a/src/tools/bookmark_ignore.h
+++ b/src/tools/bookmark_ignore.h
@@ -36,7 +36,7 @@
 #ifndef BOOKMARK_IGNORE_H
 #define BOOKMARK_IGNORE_H
 
-void bookmark_ignore_on_connect();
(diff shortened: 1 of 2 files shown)


jubalh (changes requested):
Thanks! That is correct there is a mismatch between the function defintion in the c and h file.
Please remove the " (currently causing build to fail)." from the commit message. Make sure you read the CONTRIBUTING.md file :)

jubalh (approved):
Small nitpick: you are not just adding the types you are also adding the parameter name :) So its like to align the function signature in the c and header file. Just for you to know.

jubalh:
I hope we can see more contributions to Profanity from you!

Issue #1957, opened 2024-02-18
Profanity starts talking OTR if you paste a message containing a body with an OTR start message


mdosch:
When you paste an XMPP stanza consisting of a body with an OTR start message profanity does start talking OTR although it didn't receive an OTR start message but a plain message which has an body with an OTR start in the payload, e.g.:

```
<message to='someone@somewhere.com' from='someone@somewhere.else.com/far-away' xml:lang='en' type='chat'><body>?OTRv2? *this is a test* has requested an Off-the-Record private conversation. However, you do not have a plugin to support that. See https://otr.cypherpunks.ca/ for more information.</body></message>
```

![D8TSr5JfS7mpjVnMVoPJuw](https://github.com/profanity-im/profanity/assets/5103003/a78b7998-2b4a-4cf1-8b16-fec719536255)

## Expected Behavior

Don't start talking OTR.

## Current Behavior

Profanity starts talking OTR.

## Steps to Reproduce (for bugs)

1. Open a chat to someone running profanity with OTR support.
2. Send the message pasted above.
3. See the recipient replying with an OTR start message

## Environment
* Debian stable (bookworm)
* Profanity and libstrophe from bookworm backports

```
apt policy profanity
profanity:
  Installiert:           0.14.0-1~bpo12+1
  Installationskandidat: 0.14.0-1~bpo12+1
  Versionstabelle:
 *** 0.14.0-1~bpo12+1 500
        100 http://localhost:3142/deb.debian.org/debian bookworm-backports/main amd64 Packages
        100 /var/lib/dpkg/status
     0.13.1-2 500
        500 http://localhost:3142/deb.debian.org/debian bookworm/main amd64 Packages
```

```
 apt policy libstrophe0 
libstrophe0:
  Installiert:           0.13.0-1~bpo12+1
  Installationskandidat: 0.13.0-1~bpo12+1
  Versionstabelle:
 *** 0.13.0-1~bpo12+1 500
        100 http://localhost:3142/deb.debian.org/debian bookworm-backports/main amd64 Packages
        100 /var/lib/dpkg/status
     0.12.2-1 500
        500 http://localhost:3142/deb.debian.org/debian bookworm/main amd64 Packages
```

```
profanity --version
Profanity, version 0.14.0
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2023 Michael Vetter <jubalh@iodoru.org>.
[...]

PR #2072, opened 2025-10-18
Fix OTR whitespace tag detection to prevent false positives

aryansri05:
## Description
   Fixes OTR whitespace tag false positive detection.
   
   ## Problem
   The current implementation uses `strstr()` to detect OTR whitespace tags, which matches the pattern anywhere in the message. This causes normal user messages containing these patterns to incorrectly trigger OTR session initialization.
   
   ## Solution
   Changed to use `strncmp()` to verify that V1/V2 tags immediately follow the base tag, ensuring only legitimate OTR whitespace tags trigger session establishment.
   
   ## Testing
   Built and tested on macOS with libotr 4.1.1.


jubalh:
Is this related to https://github.com/profanity-im/profanity/issues/1957 ? If yes, could you reference the issue in the commit message?

sjaeckel:
I took the liberty to re-format your changes, so it passes CI. I hope that's OK for you :)

PR #2075, opened 2025-10-25
Update compiler flags for homebrew packages on Apple silicon

mcalierno:
Add /opt/homebrew/{include,lib} flags for apple silicon Macs.

Homebrew moved from /usr/local to /opt/homebrew for Apple silicon Macs, but /opt/homebrew/{include,lib} are not included as default search paths by Clang/GCC.

As a result, without adding these paths, running 'make' fails with errors complaining that <gcrypt.h>, <signal/key_helper.h> etc cannot be found.
<br/>
**UPDATE:**

Use PKG_CHECK_MODULES to check for libgcrypt: previously `./configure --enable-omemo` resulted in

> configure: gcrypt >= 1.7.0 not found, OMEMO support not enabled
> configure: error: OMEMO support requires a library which is missed

despite libgcrypt being installed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -336,9 +336,9 @@ AM_CONDITIONAL([BUILD_OMEMO], [false])
 if test "x$enable_omemo" != xno; then
     OMEMO_LIBS=""
     PKG_CHECK_MODULES([libsignal], [libsignal-protocol-c >= 2.3.2],
-        [OMEMO_LIBS="-lsignal-protocol-c"
-         AC_CHECK_LIB([gcrypt], [gcry_md_extract],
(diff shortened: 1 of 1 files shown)


sjaeckel (approved):

sjaeckel (approved):
LGTM!

Thanks for the PR. @jubalh is sick unfortunately and I don't know when he will be back, so this could take some time to be merged.

Not sure if you have the time & patience, but maybe you want to add CI for Mac?

If you go for it I'd recommend https://github.com/nektos/act for developing that, which should also work on Mac as of their docs.

PR #2082, opened 2025-12-10
profanity: exit gracefully on SIGTERM and SIGHUP

balejk:

* step 1

Launch profanity and connect to server. Verify it can be seen as connected by other clients.

* step 2

Send SIGTERM to it or close the terminal it's running in. Verify in the other clients that it immediately comes offline, rather than only after a while when the connection times out.

### DIFF
--- a/src/profanity.c
+++ b/src/profanity.c
@@ -196,6 +196,13 @@ _connect_default(const char* const account)
 }
 
 static void
+sigterm_handler(int sig)
+{
+    log_info("Received signal %d, exiting", sig);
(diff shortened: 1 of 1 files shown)


jubalh:
Sorry for the long time to wait and thank you for your PR.

I guess this is useful to cleanup properly, update our presence (do we even do that?) and close the connection properly.

I would appreciate if you could expand the commit message a bit. For example adding to the commit body why you do this, the purpose.

While at it you can also remove the `profanity:` on the first commit line.

PR #2090, opened 2026-02-23
fix: define `prefs_changes_print` outside of `prefs_changes`

botantony:
This causes build failures on macOS runners in Homebrew, see: https://github.com/Homebrew/homebrew-core/pull/268970 https://github.com/Homebrew/homebrew-core/actions/runs/22305120219/job/64538952484?pr=268970

I didn't run `make format` because it modified other unrelated files (guess they were not formatted before) but I assume this PR follow the style guides

### DIFF
--- a/src/config/preferences.c
+++ b/src/config/preferences.c
@@ -287,6 +287,18 @@ prefs_load(const char* config_file)
     _prefs_load();
 }
 
+static void
+prefs_changes_print(const char* key, const char* newval, const char* oldval, gboolean* banner_shown)
+{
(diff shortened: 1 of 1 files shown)


sjaeckel:
> `clang does not support nested functions; this is a complex feature which is infrequently used, so it is unlikely to be implemented`

This is a useful feature which is infrequently used, since clang doesn't have support for it.

If it were my project I'd say: FU clang! Don't turn the reality around and justify not implementing a useful feature because it doesn't have usage, if you're the reason it doesn't get used.

Since it isn't my project: do whatever you want.

Issue #2098, opened 2026-02-28
Autocomplete path for sendfile causes SIGABRT


mdosch:
When doing `/sendfile ~/.config/<tab>` profanity stops with SIGABRT:

```
Program received signal SIGABRT, Aborted.
__pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0)
    at ./nptl/pthread_kill.c:44
⚠️ warning: 44	./nptl/pthread_kill.c: Datei oder Verzeichnis nicht gefunden
```

```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0)
    at ./nptl/pthread_kill.c:44
        tid = <optimized out>
        ret = 0
        pd = <optimized out>
        old_mask = {__val = {0}}
        ret = <optimized out>
#1  0x00007ffff736b47f in __pthread_kill_internal (threadid=<optimized out>, signo=6) at ./nptl/pthread_kill.c:89
No locals.
#2  0x00007ffff7314942 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
        ret = <optimized out>
#3  0x00007ffff72fc4ac in __GI_abort () at ./stdlib/abort.c:77
        act = {__sigaction_handler = {sa_handler = 0x8090a0b0c0d0e0f, sa_sigaction = 0x8090a0b0c0d0e0f}, sa_mask = {
            __val = {283686952306183, 4051950174, 0, 0, 0, 0, 0, 0, 0, 0, 140737340988388, 140737342323488, 
              22906500375, 593, 16029259789475600896, 0}}, sa_flags = -146897325, sa_restorer = 0x7fffffffdfb0}
#4  0x00007ffff72fd4e5 in __libc_message_impl (fmt=fmt@entry=0x7ffff748234b "%s\n")
    at ../sysdeps/posix/libc_fatal.c:138
        ap = {{gp_offset = 16, fp_offset = 0, overflow_arg_area = 0x7fffffffe0a0, reg_save_area = 0x7fffffffe030}}
        fd = 2
        iov = {{iov_base = 0x7ffff7485df8, iov_len = 40}, {iov_base = 0x7ffff748234d, iov_len = 1}, {
            iov_base = 0x62, iov_len = 140737345546977}, {iov_base = 0x55555561fe3c, iov_len = 140737341001319}, {
[...]

PR #2100, opened 2026-02-28
Fix file autocompletion bugs and restore cycling

jubalh:
Fix https://github.com/profanity-im/profanity/issues/2098


jubalh:
@mdosch

PR #2104, opened 2026-03-06
Improve OMEMO UX

jubalh:
Hopefully this enhanced the user experience for OMEMO users.


jubalh:
When googling `attribute(nonnull...)` I found the GLIB `G_GNUC_NONNULL` and used that.

sjaeckel:
Looks better, thanks!

jubalh:
Lets use attribute(nonnull and not G_GNUC_NONNULL (for now). The latter needs more setup and we use more __attribute in other places.

PR #2105, opened 2026-03-08
fix: Sanitize account names

### LINKED ISSUE #2103: Profanity core dumps when trying to connect to an IPv6 address
Version: 0.16.0
OS: Linux Mint x86/64

Profanity core dumps when connecting to an IPv6 address..
Image

Steps to reproduce:
- Start profanity
- /connect username@[ipv6:address] server ipv6:address
- [ENTER PW]
- .. core dumps ..

When it crashes, there are no .core files in the current working directory either.
Might be a bug I dunno :<

jubalh:
GKeyFile parses ini-like config files. So the characters '[', ']', and '=' cannot be used.
We will replace them with '_' now.

This bug was found when a user wanted to connect to an IPv6 address (user@[ipv6:address]).

One problem could be that we get duplicate account names with they would use one of those 3 characters. But I consider this very unlikely.

Fix https://github.com/profanity-im/profanity/issues/2103

### DIFF
--- a/src/config/accounts.c
+++ b/src/config/accounts.c
@@ -59,12 +59,33 @@ static GKeyFile* accounts;
 static Autocomplete all_ac;
 static Autocomplete enabled_ac;
 
+static gchar*
+_sanitize_account_name(const char* const name)
+{
+    if (!name) {
+        return NULL;
+    }
+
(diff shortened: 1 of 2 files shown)


jubalh:
Last PR referenced this by mistake. Sorry.

Issue #2045, opened 2025-06-12
101th history line (Buffer?) always get deleted


mirror230469:
Hi, after upgrading the profanity from v0.14.0 to v0.15.0, I cannot scroll up above the 100th history
line by pressing PAGEUP. The 101th history line keep getting deleted by profanity when new message comes in.
This is the error message in `profantiy.log` everytime it delete the 101th history line.

```
prof: WRN: Ncurses Overflow! From: muc@somexmppaddress/someusername, pos: 99, ID: some-uuid, message: > The actual message from someusername
```
The `someusername` and the `actual message` is from the new incoming message, not the deleted message. but it always shows `pos: 99`

## Environment
NixOS 25.05
Profanity from nix package manager

```
$ profanity --version
Profanity, version 0.15.0
Copyright (C) 2012 - 2019 James Booth <boothj5web@gmail.com>.
Copyright (C) 2019 - 2025 Michael Vetter <jubalh@iodoru.org>.
License GPLv3+: GNU GPL version 3 or later <https://www.gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Build information:
XMPP library: libstrophe
Desktop notification support: Enabled
OTR support: Enabled (libotr 4.1.1)
PGP support: Enabled (libgpgme 1.24.2-unknown)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Enabled (3.12.10)
GTK icons/clipboard: Enabled
GDK Pixbuf: Enabled
```

dependencies:
```bash
$ # perl used to chop leading nix store hash path
$ nix path-info --json $(which profanity)  | jq -r 'to_entries[]|.value.references[]' | perl -F -lane 'print @F[44..$#F]'
libgcrypt-1.10.3-lib
qrencode-4.1.1
libnotify-0.8.6
gpgme-1.24.2
cairo-1.18.2
gtk+3-3.24.49
profanity-0.15.0
libsignal-protocol-c-2.3.3
harfbuzz-10.2.0
openssl-3.4.1
python3-3.12.10
readline-8.2p13
libgpg-error-1.51
glib-2.84.1
gdk-pixbuf-2.42.12
glibc-2.40-66
libstrophe-0.14.0
expat-2.7.1
at-spi2-core-2.56.2
libX11-1.8.12
libassuan-3.0.2
libotr-4.1.1
zlib-1.3.1
curl-8.13.0
ncurses-6.5
[...]

PR #2122, opened 2026-03-19
feat(ui): implement dynamic pad resizing

jubalh:
The ncurses pad used for rendering window history had a fixed height of 100 lines, which caused "Ncurses Overflow!" errors and broke scrolling once the content exceeded this limit.

This commit replaces the fixed limit with a dynamic resizing strategy. We use a minimum height of 100 lines.
If a pad reaches its threshold of 3000 lines, a full redraw is triggered to reclaim space from discarded messages and reset the pad height. Pads are shrunk back to the minimum height during every redraw to minimize memory usage for inactive windows.

Message history should always be smoothly scrollable now.

d7e46d64fe6bc9926eb404dad6a4803e1fc2e8b2 set it from 1000 to 10000. f27fa98717631abaa647d4224235d4fafb00a7f9 set it from 1000 to 100. Probably also relevant is 4fe2c423b147417ae181b595b8ba9df72c0c5ec1.

Close: https://github.com/profanity-im/profanity/pull/2074
Fixes: https://github.com/profanity-im/profanity/issues/2045

I tested this by sending a lot of messages in a 1:1 chat. With this change I could scroll up fine. And `Ncurses Overflow` wasn't in the logs anymore.

### DIFF
--- a/src/ui/window.c
+++ b/src/ui/window.c
@@ -39,7 +39,31 @@
 #include "xmpp/xmpp.h"
 #include "xmpp/roster_list.h"
 
-static const int PAD_SIZE = 100;
+static const int PAD_MIN_HEIGHT = 100;
+static const int PAD_THRESHOLD = 3000;
+static gboolean _in_redraw = FALSE;
+
+static void
+_win_ensure_pad_capacity(ProfWin* window, WINDOW* win, int lines_needed)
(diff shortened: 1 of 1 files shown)


jubalh:
Reminder: test in MUC

PR #2135, opened 2026-03-27
ci: add macOS Homebrew CI

GunniBusch:
## Summary
- add a macOS Homebrew CI job alongside the Linux matrix
- install dependencies via Homebrew using the official setup action
- run the existing build matrix script on macOS for ~~Autotools~~ and Meson

## Notes
- Autotools disables OMEMO on macOS because Homebrew disabled libsignal-protocol-c in August 2025
- Meson keeps OMEMO coverage via libomemo-c

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: Linux
+name: CI
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


jubalh:
> Also regarding libsignal-protocol-c, homebrew doesn't support that, but should I still test for it? personally I would not do it because the majority of users will use hombrew so they cant install it anyway.

Agreed, no need to test for it.

> Also I currently selected macOS-latest runner which is the newest macOS version and arm. Should I also add more like an earlier version or intel x86?

I'm fine with the current way :)

PR #2150, opened 2026-04-11
fix: resolve issues with async editor

### LINKED ISSUE #2148: Neovim unusable as editor
When using the editor (neovim in my case) it is totally unusble. It doesn't do what
I want and behves erratic. Hard to explain, you have to try it yourself.

## Expected Behavior

Just use nvim normally.

## Current Behavior

Weird behavior while writing.

## Possible Solution

Seems reverting 36ec2b0 fixes ist. So maybe a bug in the new behavior.

1. Use neovim as editor
2. Press `Alt+C` in a chat and try to write using nvim.

## Environment

* Debian testing/unstable
* Sway 1.11-3
* Neovim 0.11.6-1
* glib version: libglib2.0-0t64 2.88.0-1
* libstrophe version: 0.14.0-5+b1

jubalh:
If we build Profanity without GTK we dont iterate the context so we will never know when the editor exited. Calling `g_main_context_iteration()` or finally switching to GMainLoop fixes this.
So far calling `tray_update()` does this for us in GTK builds. So this went unnoticed.

The editor will inherit ignored signal handlers (SIGINT, SIGTSTP, SIGPIPE) from us. Neovim uses libuv which seems to check SIGINT on startup to determine whether the environment is interactive or not. We now reset these signals to SIG_DFL in the child process before execvp.

Profanity uses readline which still competed for terminal input with the editor. We only took care about other Profanity UI processes in our previous commit. This led to misordered characters in the editor.

In my tests with vim everything worked fine and these bugs were discovered when a user used neovim.

Fixes: https://github.com/profanity-im/profanity/issues/2148
Ref: 36ec2b0ae1fa7afcdfe6f166323955b8bd121b0f

### DIFF
--- a/src/profanity.c
+++ b/src/profanity.c
@@ -139,6 +139,8 @@ prof_run(gchar* log_level, gchar* account_name, gchar* config_file, gchar* log_f
         session_process_events();
         iq_autoping_check();
         ui_update();
+        while (g_main_context_iteration(NULL, FALSE))
+            ;
 #ifdef HAVE_GTK
         tray_update();
 #endif
--- a/src/tools/editor.c
+++ b/src/tools/editor.c
@@ -14,6 +14,7 @@
 #include <errno.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#include <signal.h>
 
 #include "config/files.h"
 #include "config/preferences.h"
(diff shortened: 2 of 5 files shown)


jubalh:
I tested this with on openSUSE Tumbleweed with:
gnome3 + gnome-terminal + foot + vim/nvim
sway + alacritty + foot + vim/nvim
And it worked in all cases.

PR #2161, opened 2026-04-27
Fix message receipts

jubalh:
When we wanted to adhere better to the XEPs we actually did a mistake.

In commit 421bb2e (so the mistake is there since 0.17.0), a check was added to only request receipts if
support was confirmed via entity capabilities. This was too restrictive as it defaulted to 'not supported' when capabilities were not yet cached (new sessions or bare JIDs).

Following XEP-0184 recommendations, we now only omit the receipt request if we have positive knowledge that the contact does not support it. If capabilities are unknown, we follow the users preference.

Ref: 421bb2e2616af0f20500536d9842174b0943d0b2


jubalh:
@sezuan please test.

sezuan:
Works! Thank you.

Issue #2163, opened 2026-05-06
SQLite error in _migrate_to_v3() 


mdosch:
I realized I sometimes get old messages from a year ago replayed instead of more recent messages when opening a chat.
When looking in the error logs I found a lot of these:

> 2026-05-05T20:09:51.847229+02: 00255105: prof: ERR: SQLite error in _add_to_db() (prepare): ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
> 2026-05-05T20:10:01.425463+02: 00255105: prof: ERR: SQLite error in _add_to_db() (prepare): ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
> 2026-05-05T20:11:12.853284+02: 00255105: prof: ERR: SQLite error in _add_to_db() (prepare): ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
> 2026-05-05T20:11:34.653412+02: 00255105: prof: ERR: [OMEMO][STORE] SG_ERR_INVALID_KEY_ID
> 2026-05-05T20:12:21.256679+02: 00255105: prof: ERR: SQLite error in _add_to_db() (prepare): ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
> 2026-05-05T20:12:55.195326+02: 00255105: prof: ERR: SQLite error in _add_to_db() (prepare): ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
> 2026-05-05T20:13:38.516691+02: 00255105: prof: ERR: SQLite error in _add_to_db() (prepare): ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
> 2026-05-05T20:15:09.686811+02: 00255105: prof: ERR: SQLite error in _add_to_db() (prepare): ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint

When looking for the first occurrence of db issues I found this:
> 2026-04-11T17:01:34.263737+02: 00066757: conn: DBG: Q_ADD: 0x561a4078b280
> 2026-04-11T17:01:34.276203+02: 00066757: prof: DBG: [Keyfile]: free /home/martin/.local/share/profanity/accounts
> 2026-04-11T17:01:34.279811+02: 00066757: prof: DBG: [Keyfile]: free /home/martin/.local/share/profanity/accounts
[...]

PR #2165, opened 2026-05-15
fix(database): resolve migration failure and improve init safety

jubalh:
Ensure the DbVersion table contains exactly one row during migrations by using DELETE and INSERT instead of UPDATE. This fixes issues where multiple version rows caused unique constraint failures and rolled back schema migrations.

Properly close the database on initialization failure to prevent subsequent operations on a broken state.

Initialize new databases with the latest schema version to avoid redundant migration attempts.

Fixes https://github.com/profanity-im/profanity/issues/2105

### DIFF
--- a/src/database.c
+++ b/src/database.c
@@ -159,10 +159,12 @@ log_database_init(ProfAccount* account)
     }
 
     if (db_version == -1) {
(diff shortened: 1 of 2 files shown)


mdosch:
This fixed the DB issue for me.

jubalh:
I made a mistake in the issue number. This PR fixes https://github.com/profanity-im/profanity/issues/2163.

PR #2166, opened 2026-05-16
Stop OTR from leaking network availability

techmetx11:
In Profanity, OTR can be abused to leak the user's network availability, without a subscription request and with just their bare JID. This may have been exploited in the wild by malicious users and spammers.

This PR will stop `libotr` from injecting messages if the recipient doesn't have a subscription

### DIFF
--- a/src/otr/otr.c
+++ b/src/otr/otr.c
@@ -96,6 +96,16 @@ static void
 cb_inject_message(void* opdata, const char* accountname,
                   const char* protocol, const char* recipient, const char* message)
 {
+    PContact contact = roster_get_contact(recipient);
+
+    if (contact == NULL) {
+        return;
+    }
+
+    if (p_contact_subscribed(contact) == FALSE) {
+        return;
+    }
+
     free(message_send_chat_otr(recipient, message, FALSE, NULL));
 }
 


jubalh:
There seems to be missing context here. Could you expand a bit what on what is happening and how?

Also I think there are people that want to chat using OTR and don't grant subscription to the roster to the other party. So this PR might need redoing. Probably a setting will be needed for `/otr` to decide about the behaviour.

But let's hear more about the original concern first.

techmetx11:
> There seems to be missing context here. Could you expand a bit what on what is happening and how?
> 
> Also I think there are people that want to chat using OTR and don't grant subscription to the roster to the other party. So this PR might need redoing. Probably a setting will be needed for `/otr` to decide about the behaviour.
> 
> But let's hear more about the original concern first.

Spammers have been abusing OTR presumably to figure out if users are online, because OTR will inject a message upon receiving a handshake message (`?OTRv23?`) with no regard for if the entity is authorized to know about the user's network availability.

What also makes this worse is that if the attacker doesn't use `<no-store>`, `<no-copy>` or `<private>`, Profanity will continue to respond to the same OTR handshake between client restarts because it gets the same handshake message from the server's history. Essentially allowing the attacker to figure out when a Profanity user has come online just by receiving OTR messages from them (and without the user being aware of it)

monocles.chat fixed this issue somehow. It doesn't respond to OTR handshakes from strangers

jubalh:
I'm curious what other clients do. But I think not many even still support OTR at this point, right?

jubalh:
Ok I think I understood the problem now well enough.
I'll look into it when I find some free time. Thanks for reporting the issue!

jubalh:
@techmetx11 please take a look at https://github.com/profanity-im/profanity/pull/2168.
I think this is how we should do it. What do you think?

techmetx11:
> @techmetx11 please take a look at #2168. I think this is how we should do it. What do you think?

I think that's better than what I did here. LGTM!

PR #2167, opened 2026-05-19
fix(ui): guard terminal writes while suspended

### LINKED ISSUE #2162: When editor is active profanity continues to update statusbar seconds.
Using latest and greatest master from today:

When I invoke an editor (vim in my case) the time in the status bar is still updated and confuses the editor. Maybe a side effect of the async editor.

jubalh:
We forgot to guard `beep()` and `flush()`.

Ref: 36ec2b0ae1fa7afcdfe6f166323955b8bd121b0f
Ref: dd76f9087fe93db5ced608ee3c744f299b933cbe

Fixes: https://github.com/profanity-im/profanity/issues/2162


mdosch:
Worked well for me on Debian Unstable with foot. :)

PR #2168, opened 2026-05-19
fix(otr): prevent OTR handshake from leaking presence to strangers

jubalh:
`cb_is_logged_in()` incorrectly returned `PRESENCE_ONLINE` for contacts not in the roster or without a presence subscription.

According to libotr API documentation, returning 1 informs the library that it is safe to send automated background traffic (heartbeats or handshake responses).

Because of this, when spammers sent an initial OTR query (`?OTR?v23?`), we would automatically inject a handshake response, confirming the users network availability.

This violates XMPP Presence Privacy best practices defined in RFC 6121 Section 4.1, which states that presence information must not be revealed to entities without explicit authorization.

Fixes: https://github.com/profanity-im/profanity/pull/2166

### DIFF
--- a/src/otr/otr.c
+++ b/src/otr/otr.c
@@ -68,28 +68,18 @@ static int
 cb_is_logged_in(void* opdata, const char* accountname, const char* protocol, const char* recipient)
 {
     jabber_conn_status_t conn_status = connection_get_status();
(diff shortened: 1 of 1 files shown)


jubalh:
See https://github.com/profanity-im/profanity/pull/2166#issuecomment-4498043604

PR #2170, opened 2026-05-25
feat(ci): Create Fedora RPMs using Packit

bkmgit:
Add a pipeline to enable builds using https://packit.dev/ on generation of Fedora RPMs on each commit,

To get it to work, someone with commit access will need to agree to make a Fedora account at:
https://accounts.fedoraproject.org/

and agree to Fedora Code of Conduct, then follow the steps at:

http://packit.dev/docs/guide#1-set-up-packit-integration

* If CI passes, testing RPMs for Fedora will be created.
* Install built RPMs on a machine running Fedora.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,11 @@
+specfile_path: packaging/fedora/profanity.spec
+
+notifications:
+  pull_request:
+    successful_build: true
+
+jobs:
+- job: copr_build
+  trigger: pull_request
+  targets:
+  - fedora-all
--- a/packaging/fedora/profanity.spec
+++ b/packaging/fedora/profanity.spec
@@ -1,0 +1,179 @@
+Name:           profanity
+Version:        0.18.1
+Release:        %{autorelease}
+Summary:        A console based XMPP client
+
+License:        GPL-3.0-or-later WITH cryptsetup-OpenSSL-exception
+URL:            https://profanity-im.github.io/
+Source0:        %{url}/tarballs/%{name}-%{version}.tar.xz
+
+BuildRequires:  gcc
+BuildRequires:  meson
+# Base:
+BuildRequires:  libstrophe-devel
+BuildRequires:  ncurses-devel
+BuildRequires:  glib2-devel
+BuildRequires:  libcurl-devel
+BuildRequires:  readline-devel
+BuildRequires:  sqlite-devel
+BuildRequires:  python-unversioned-command
+# Optional dependencies for support:
+# Desktop notification support
+BuildRequires:  libnotify-devel
+# OTR support
+BuildRequires:  libotr-devel
+# PGP support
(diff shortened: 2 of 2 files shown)


jubalh:
What is the purpose of these rpms? It's for testing whether rpm building works?

bkmgit:
A user had indicated wish to use very latest version.  Since distro policies prefer stability in packaged versions in releases, thought it would be useful to make these available here. 

OBS seems to have something similar:
https://openbuildservice.org/help/manuals/obs-user-guide/cha-obs-scm-ci-workflow-integration.html

jubalh:
OBS has a feature where you install a webhook, and upon each commit (to master or where you define it) OBS will build a new rpm. This rpm is stored/retrievable from OBS.
So basically OBS will do everything and on the GitHub site you have just the trigger for the webhook.

Is it the same with this solution here?

With OBS we wouldn't have to add a spec file to the upstream project. This would be stored in the OBS package. It then automatically pulls the sources from upstream when a webhook arrives and creates a tarball from them. This has the advantage that the same spec file like for distribution packaging can be used without upstream having to keep them in sync.

bkmgit:
> OBS has a feature where you install a webhook, and upon each commit (to master or where you define it) OBS will build a new rpm. This rpm is stored/retrievable from OBS. So basically OBS will do everything and on the GitHub site you have just the trigger for the webhook.
> 
> Is it the same with this solution here?

Very similar.

> 
> With OBS we wouldn't have to add a spec file to the upstream project. This would be stored in the OBS package. It then automatically pulls the sources from upstream when a webhook arrives and creates a tarball from them. This has the advantage that the same spec file like for distribution packaging can be used without upstream having to keep them in sync.

The built RPMs would be on Fedora infrastructure.  There would be a message on each pull request if tests ran successfully.  If you wish to have them on OBS, that is fine.  The Fedora and Suse spec files have some minor differences, can keep them separate, or add some #ifdef statements.

jubalh:
> There would be a message on each pull request if tests ran successfully.

Is there another project where I can see how it looks/works?

> If you wish to have them on OBS, that is fine. 

Personally I don't care, I would just like to omit to have to keep the spec file up to date.

bkmgit:
Some examples:
- https://github.com/bkmgit/python-colormath2/pull/15
- https://github.com/enactic/openarm_can/pull/135

It does seem possible to automate spec file updates using https://github.com/packit/specfile

Am willing to explore OBS as well.

jubalh:
I have to confess I'm not a big fan of it.
My main reason is that I don't want to set up more accounts.

I would also prefer is distro infrastructure could just build git packages and provide them for interested users by themselves.

jubalh:
I'll close this since I currently don't have the capacities to open another account and learn how to use packit etc.

PR #2178, opened 2026-06-15
fix(ox): filter keys listed by `/ox keys` to those matching xmpp prefix

### LINKED ISSUE #2177: Ox: Listing keys lists all PGP keys in the system
`/ox keys` lists all PGP keys known to the system. So it shows a very huge amount of keys of which most can not be used for Ox. According to [XEP-0373: OpenPGP for XMPP](https://xmpp.org/extensions/xep-0373.html#openpgp-verification) the user ID must be in the form "xmpp:juliet@example.org" which is not the case for the majority of PGP keys in my system.

## Expected Behavior
Show only Ox keys.

## Current Behavior
Shows all PGP keys known to the system.

## Possible Solution
Filter for IDs with leading `xmpp:`

1. `/ox keys`
2. See an incredible long list of keys which are mostly not from XMPP contacts and cry.

## Context

As I think Ox will gain traction after Conversations supports it, I'd like to suggest some improvements to profanitys Ox support.

## Environment
* Debian sid/testing

```
profanity --version
Profanity, version 0.18.2dev.fix/pubsub.bbd9442d
Copyright (C) 2019 - 2026 Michael Vetter <jubalh@iodoru.org>.

Build information:
XMPP library: libstrophe
OTR support: Disabled
PGP support: Enabled (libgpgme 2.1.0)
OMEMO support: Enabled
C plugins: Enabled
Python plugins: Disabled
GTK icons/clipboard: Enabled
GDK Pixbuf: Disabled
```

jubalh:
The `/ox keys` command previously listed all PGP keys known to the system, displaying a massive number of keys of which the vast majority cannot be used for OX.

According to XEP-0373: OpenPGP for XMPP, the User ID of the key must correspond to the JID of the user as an 'xmpp' URI ("xmpp:juliet@example.org"). General PGP keys in the user's keyring that lack this prefix are invalid for OX.

* Introduce internal helper `_ox_gpg_get_keys()` to retrieve keys containing an 'xmpp:' UID prefix.
* Create new `ox_gpg_list_keys()` that runs a second-pass query specifically looking for corresponding private keys in the secret keyring.
* Update `/ox keys` to call `ox_gpg_list_keys()` instead of the generic `p_gpg_list_keys()`.

Fixes: https://github.com/profanity-im/profanity/issues/2177

### DIFF
--- a/src/command/cmd_funcs.c
+++ b/src/command/cmd_funcs.c
@@ -7508,7 +7508,7 @@ cmd_ox(ProfWin* window, const char* const command, gchar** args)
     // Should we move this to a common command
     // e.g. '/openpgp keys'?.
     else if (g_strcmp0(args[0], "keys") == 0) {
-        GHashTable* keys = p_gpg_list_keys();
+        GHashTable* keys = ox_gpg_list_keys();
         if (!keys || g_hash_table_size(keys) == 0) {
             cons_show("No keys found");
             return TRUE;
--- a/src/pgp/ox.c
+++ b/src/pgp/ox.c
@@ -28,18 +28,10 @@
 
 static gpgme_key_t _ox_key_lookup(const char* const barejid, gboolean secret_only);
 static gboolean _ox_key_is_usable(gpgme_key_t key, const char* const barejid, gboolean secret);
+static GHashTable* _ox_gpg_get_keys(gboolean check_secret);
 
-/*!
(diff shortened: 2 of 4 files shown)


jubalh:
@mdosch please test :)

mdosch:
Works well, thanks!

PR #2181, opened 2026-06-25
Implement MUC MAM

jubalh:
I recommend to *not* run this on your regular system in case we didn't get things right yet regarding the database.
Either backup your sqlite db before or work on a separate linux user account or use a container or something like that.

### DIFF
--- a/profanity.doap
+++ b/profanity.doap
@@ -360,7 +360,7 @@
         <xmpp:status>partial</xmpp:status>
         <xmpp:version>1.1.0</xmpp:version>
         <xmpp:since>0.14.0</xmpp:since>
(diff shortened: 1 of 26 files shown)


jubalh:
Seems to work well. I still have some ideas and improvements to do later. Might merge it already though.

jubalh:
Note: I didn't test yet whether message loading from DB works still as expected if MAM is disabled.

jubalh:
That functional test fail is expected. I'll need to adapt the tests according to my changes. Probably should adapt stabber to advertise features.

jubalh:
https://github.com/profanity-im/stabber/commit/95a186e09caf759d99b7053cb0feee907dcdbfc5

PR #2183, opened 2026-07-04
fix(mam): Prevent duplicated and out-of-order history messages

jubalh:
This resolves an issue where MAM history messages were being appended to the window display buffer in as they arrived, causing them to appear out of order and duplicated when the RSM batch completed and prepended the same sorted messages.

Bypass immediate display and file logging for MAM messages and allow them to only be saved to the database.
Render them in a single chronological block once the query completes.

So in server_events.c we now only decrypt and save MAM messages to db. Once we receive `<fin>` we trigger `_mam_rsm_id_handler()` and `_mam_buffer_commit_handler()` which call `mucwin_db_history() and `chatwin_db_history()` which will query the DB and print them to the screen via `win_print_old_history()`.

### DIFF
--- a/src/event/server_events.c
+++ b/src/event/server_events.c
@@ -358,9 +358,13 @@ sv_ev_room_message(ProfMessage* message)
     }
 
     if (message->is_mam) {
-        // For MAM, print to window but disable all notifications, unread count increment etc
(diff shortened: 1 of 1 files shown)


jubalh:
Confirmed by bug reporter maniac in our MUC.

PR #2185, opened 2026-07-04
More mam fixes

jubalh:
`maniac` mentioned `when I scroll like, 2 months _backwards in time_ in the chat history, I can only scroll _forwards in time_ a month.
So we found a bug in rsm pagination.
Also add a note how to report bugs if you don't have an GH account.
And last but not least a fix for the wrong display name.


jubalh:
Tested by maniac and vamp898

PR #2195, opened 2026-07-20
fix(omemo): Correctly display our own sent OMEMO messages in MUCs and 1:1 chats

jubalh:
We log our own messages when we send them.
Thus we don't have the stable archive ID that the server assigns to it.

When we get MAM history (or standard MUC history), we look for the ID,
think the message is new and thus display a decryption error because
we can't decrypt the message due to double-ratchet.

Unify database update and retrieval functions using the prof_msg_type_t enum
to stay DRY, and support both MUC and 1:1 chats.

Update stable ID for self-reflected messages in our database.
So later we can identify them correctly and load them from the DB.

### DIFF
--- a/src/database.c
+++ b/src/database.c
@@ -896,3 +896,90 @@ cleanup:
 
     return FALSE;
 }
+
+gboolean
+log_database_update_archive_id(const prof_msg_type_t type, const char* const room_or_contact_jid, const char* const stanza_id, const char* const archive_id)
+{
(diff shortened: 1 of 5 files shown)


jubalh:
Confirmed by vamp898 in the MUC to fix the issue he reported.

jubalh:
> Confirmed by vamp898 in the MUC to fix the issue he reported.

Later he found another issue. I updated the commit, lets see.

jubalh:
Confirmed to fix the issue.

PR #2196, opened 2026-07-21
fix(mam): do not discard self sent messages received via MAM if sent …

jubalh:
…offline

Profanity is offline and the user is sending messages from another client,

When Profanity comes online and requests MAM history, these self sent messages are received.
Before sv_ev_incoming_message classified all MAM messages from our own bare JID as a self reflection, attempted to update their archive ID in the database, and then immediately discarded them by returning early. Since the database update failed because the message was sent while offline, the message was permanently lost and never logged or displayed.


jubalh:
Tested by vamp898

PR #2201, opened 2026-07-30
Update LICENSE.txt

Maseru898:
"New" Maintainer

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,5 +1,6 @@
 Profanity
 Copyright (C) 2012 - 2019 James Booth <boothj5@gmail.com>
+Copyright (C) 2019 - 2026 Michael Vetter <jubalh@iodoru.org>
 
 Profanity is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by


jubalh (changes requested):
Thanks! See contributing.md to fix the commits.

Maseru898:
Is this docs or chore?

jubalh:
chore

jubalh:
Fix the CI so I can merge.

jubalh:
@Maseru898 you are still not following the rules ;)

jubalh:
"Update LICENSE.txt" doesn't make sense in the first place. Such a generic title doesn't tell us anything. Why don't you use `chore: Add current maintainer to LICENSE.txt` or something similarly descriptive?

jubalh:
Is there any problem with this? I'm surprised simply following CONTRIBUTING.md and making an adjustment in  commit messages takes so much time.

Maseru898:
You're right, its a combination of two issues. Not being able to push 99% of the time due to network restrictions and also simply not having time to begin with.

It was just a bad idea opening this without properly reading through documentation and everything, i am causing more issues  than i am helping, sorry for that.

